PR #9, opened 2015-01-21
Move to loader.js@3.0.0 and use ember-cli dependencies

chadhietala:


dgeb:
Thanks again @chadhietala!

I should have just included this in v0.1.0, but instead I had to update to v0.2.0 with the new dependencies.

PR #10, opened 2015-01-22
Update ES6 module transpiler and proper sourcemap concatenation

chadhietala:
We're trying to update loader js in https://github.com/ember-cli/ember-cli/pull/3054 and we noticed that ember-test-helpers were producing an incorrect AMD output.  This PR replaces the old ES6 transpiler for the one that is used by Ember CLI and produces the correct output.  I also added the sourcemap concatenator which knows how to properly concat sourcemaps.

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -1,14 +1,14 @@
-var concat     = require('broccoli-concat');
 var pickFiles  = require('broccoli-static-compiler');
 var mergeTrees = require('broccoli-merge-trees');
-var compileES6 = require('broccoli-es6-concatenator');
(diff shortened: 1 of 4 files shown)


stefanpenner:
awesome, seems like this may require a large version bump.

dgeb:
Thanks @chadhietala!

@stefanpenner @rwjblue do you want me to bump this from 0.0.7 to 0.1.0 or 1.0.0?

rwjblue:
@dgeb - 0.1.0

dgeb:
ok, v0.1.0 has been published :shipit:

PR #12, opened 2015-01-31
Ensure that `setup` callback is invoked with the correct context.

rwjblue:
Previously, the `setup` callback was invoked via `invokeSteps` (and
therefore was bound to the `TestModule` instance itself).  Now it is
invoked with the same context as the tests themselves.

* Ensure `setup` callback is invoked with the test context.
* Ensure `teardown` callback is invoked with the test context.
* Removed `beforeTeardown`.
* Added `afterTeardown` which is invoked with the test-module context.
* Simplify callback ordering tests.

/cc @dgeb

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -36,6 +36,7 @@ export default Klass.extend({
 
   initSetupSteps: function() {
     this.setupSteps = [];
+    this.contextualizedSetupSteps = [];
 
     if (this.callbacks.beforeSetup) {
       this.setupSteps.push( this.callbacks.beforeSetup );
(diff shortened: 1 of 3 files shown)


lib/ember-test-helpers/test-module.js
```diff
@@ -45,11 +45,6 @@ export default Klass.extend({
     this.setupSteps.push(this.setupContainer);
     this.setupSteps.push(this.setupContext);
     this.setupSteps.push(this.setupTestElements);
-
-    if (this.callbacks.setup) {
-      this.setupSteps.push( this.callbacks.setup );
-      delete this.callbacks.setup;
```
dgeb: We want to delete `setup` from `this.callbacks` before `contextualizeCallbacks` is called. Otherwise, `setup` will get added to `cache['setup']`.

Perhaps we need another array, something like `postSetupSteps`, that will be invoked with the proper `this.context`. And then we should allow `invokeSteps` to take a `context` argument, which will be either `this` or `this.context`. 

I can do this as a separate PR if you like or we can chat on IRC.

dgeb: Maybe to be clear we should introduce both `contextualizedSetupSteps` and `contextualizedTeardownSteps`, both of which will be invoked with the module's `context`.

rwjblue: Sounds good, I'll update in a few....

rwjblue:
@dgeb - Updated per conversation in IRC and comments here.

dgeb:
Many thanks @rwjblue - this is a big improvement!

rwjblue:
Tagged as 0.3.0.

PR #17, opened 2015-02-20
make sure bower_components folder is where broccoli expects

CodeOfficer:
Bower components were installing into `app/bower_components` due to a mysterious `.bowerrc` file higher up in my file system. `npm start' would fail because bower-components were not being found. This makes sure bower components will be installed where broccoli expects them.


rwjblue:
:+1:

PR #18, opened 2015-02-21
destroy the subject in teardown

CodeOfficer:
I'm not sure what there is to test here so I didn't add any tests yet. 

Currently the container is destroyed which takes care of destroying any dependencies brought in by `needs` but the subject is never properly town down. This makes sure that happens.

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -62,6 +62,7 @@ export default Klass.extend({
       delete this.callbacks.teardown;
     }
 
+    this.teardownSteps.push(this.teardownSubject);
     this.teardownSteps.push(this.teardownContainer);
     this.teardownSteps.push(this.teardownContext);
     this.teardownSteps.push(this.teardownTestElements);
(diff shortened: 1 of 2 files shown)


rwjblue:
I'd like to see the following test near the end of https://github.com/switchfly/ember-test-helpers/blob/master/tests/test-module-test.js:

```javascript
moduleFor('component:x-foo', 'component:x-foo -- created subjects are cleaned up', {
  beforeSetup: function() {
    setupRegistry();
  },

  afterTeardown: function() {
    var subject = this.cache.subject;
    
    ok(subject.isDestroyed);
  }
});

test("subject's created in a test are destroyed", function() {
  this.subject();
});

moduleFor('component:x-foo', 'component:x-foo -- uncreated subjects do not error', {
  beforeSetup: function() {
    setupRegistry();
  }
});

test("subject's created in a test are destroyed", function() {
  expect(0);
});
```

lib/ember-test-helpers/test-module.js
```diff
@@ -122,6 +123,14 @@ export default Klass.extend({
     }
   },
 
+  teardownSubject: function() {
+    var subject = this.cache.subject;
+
+    if (subject) {
+      Ember.run(subject, 'destroy');
```
rwjblue: It is possible that `subject` does not have a `destroy` method so you will probably want to do something like:

```javascript
Ember.run(function() {
  Ember.tryInvoke(subject, 'destroy');
});
```

CodeOfficer:
^_^

rwjblue:
:heart:

rwjblue:
I'm :+1:, but I'll give @dgeb a chance to chime in...

dgeb:
Looks great - thanks @CodeOfficer!

rwjblue:
released in v0.3.4 of ember-test-helpers and v0.2.10 of ember-qunit.

PR #26, opened 2015-03-16
Render for tagless component should not call `view.$()`

bantic:
In Ember 1.11, calling `this.$()` for a tagless component (`tagName === ''`), ember will [throw an assertion complaining](https://github.com/emberjs/ember.js/blob/06e41ad7ccd28558dd4e651aa070bc06f7757821/packages/ember-views/lib/views/view.js#L943-L946).

`moduleForComponent` always calls `this.$()` in its `render` callback, so in Ember 1.11 you can't effectively test the rendering of tagless components.

This adds an isolated commit with a failing test (and ember upgraded to 1.11.0-beta.5) and a second commit fixing that test.

I'm not sure if this is the best solution. It makes `this.render` simply return `undefined` for tagless components, which may be more surprising for a user than the assertion message that ember will raise.

Perhaps the `render` callback's call signature can change to include a `shouldGetElement` boolean instead?
So for a tagless component you would call `this.render(false)` in your test. And in that case `render` could raise its own assertion when called without `false` for a tagless component.

Alternatively, could use a heuristic to just find the first child of the ContainerView in the `#ember-testing` div, so that `this.render` always returns an element.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -25,7 +25,7 @@
   ],
   "dependencies": {
     "klassy": "cerebris/klassy.js#0.1.0",
-    "ember": "^1.10.0"
+    "ember": "1.11.0-beta.5"
   },
   "devDependencies": {
     "qunit": "^1.15.0",
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -44,7 +44,9 @@ export default TestModule.extend({
         });
       });
 
-      return view.$();
+      if (view.tagName !== '') {
+        return view.$();
+      }
     };
 
     this.callbacks.append = function() {
--- a/tests/test-module-for-component-test.js
+++ b/tests/test-module-for-component-test.js
@@ -39,11 +39,16 @@ var BoringColor = Ember.Component.extend({
   }
 });
 
+var TaglessThing = Ember.Component.extend({
+  tagName: ''
+});
+
 function setupRegistry() {
   setResolverRegistry({
     'component:x-foo': Ember.Component.extend(),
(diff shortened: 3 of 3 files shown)


rwjblue:
I think that I would prefer to simply remove the `view.$()` call all together.  It would be a breaking API change, so we would have to bump ember-test-helpers to 0.4.0, but I think that it makes the most sense.  If you need the element you can always call `this.$()` (which will need to be updated to call `this.render(); this.subject.$()`)...

rwjblue:
@dgeb - Do you have any objections to changing `this.render()` not return the element, and bumping the version accordingly?

mixonic:
@rwjblue my thoughts are to decouple this from `this.$()`.

```
var element = Ember.get(view, 'element');
if (element) {
  return jQuery(element);
}
return null;
```

bantic:
@rwjblue I like that suggestion. I like @mixonic's suggestion for decoupling the `this.$()` but if a breaking change is on the table that makes the most sense to me, rather than making `this.render()` return a sometimes-null-sometimes-element.

If that sounds good to @dgeb I can rework this PR to move it to that.

dgeb:
I am not heavily invested in this issue, so if you all have reached consensus then I am :+1: as well.

mixonic:
`append` would still return the element if available?

mixonic:
Oh `append` has been removed. So it would, but would still raise a deprecation, and `render` would not. And `this.$()` would raise if you call it without an element I presume.

bantic:
I made a PR that changes `render` to not call `view.$()` anymore, and had to make a few other small changes to make it work properly. #27

PR #27, opened 2015-03-17
[breaking] Remove `view.$()` call in `render()` of module for component

bantic:
from #26 

This PR makes two changes:

  * use a `cachedCalls` object internally in TestModule to track whether a given callback has been called. Previously, the code only did a check on the result (in `this.cache[key]`), which means that a callback that returns a falsy value will (incorrectly, it seems) be called multiple times. This seemed likely to result in inconsistent behavior. (This change is necessary for the following change)
  * change `this.callbacks.render` to *not* return anything. It simply renders the component. This is a _breaking change_. Previously, `render` would return the rendered view's `this.$()` method, but this is problematic because there are cases where a view's `$()` method should not be called. Specifically, in Ember 1.11.0.beta-5, calling `this.$()` on a tagless component (where `tagName = ''`) will result in a failed assertion: ["You cannot access this.$() on a component with `tagName: \'\'` specified."](https://github.com/emberjs/ember.js/blob/06e41ad7ccd28558dd4e651aa070bc06f7757821/packages/ember-views/lib/views/view.js#L943-L946), making it difficult-to-impossible to test such a component in a unit test. 

Test users can still call `this.$()` to get the jquery-wrapped element (in cases where that makes sense — if the component is tagless they will still get the failed assert mentioned above)

cc @rwjblue @mixonic

### DIFF
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -45,8 +45,8 @@ export default TestModule.extend({
     };
 
     this.callbacks.append = function() {
-      Ember.deprecate('this.append() is deprecated. Please use this.render() instead.');
-      return this.callbacks.render();
+      Ember.deprecate('this.append() is deprecated. Please use this.render() or this.$() instead.');
+      return context.$();
     };
 
     context.$ = function() {
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -31,11 +31,10 @@ export default TestModule.extend({
 
     this.callbacks.render = function() {
       var containerView = Ember.ContainerView.create({container: container});
-      var view = Ember.run(function(){
+      Ember.run(function(){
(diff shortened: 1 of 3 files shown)


rwjblue:
:+1: - LGTM

mixonic:
@bantic can we make `append` continue to return he jquery object? By trying for and wrapping `view.element`?

We certain have projects that rely on `append` returning DOM, and the move to `render` seems like a good time to change them. But I would be sad if we changed `append` without backwards compat.

bantic:
@mixonic coming up...

bantic:
Updated `append` test to ensure it:
  * makes a deprecation
  * returns the jquery-wrapped this.$() element

mixonic:
+1

bantic:
@dgeb I think this is good to go, if you are happy with it

dgeb:
Good improvements - thanks @bantic!

dgeb:
Just tagged v0.4.0.

PR #28, opened 2015-03-24
Ensure `moduleForModel` uses the correct store for `subject`.

rwjblue:
Previously, the `container` was shared across each module (NOT each test) inside `this.subject()` and `this.store()`.

This ensures that those callbacks grab the `container` for each test run (instead of using the closure scope variable).

Credit to @bmac for tracking down the cause, and providing a reliable test reproduction.

### DIFF
--- a/lib/ember-test-helpers/test-module-for-model.js
+++ b/lib/ember-test-helpers/test-module-for-model.js
@@ -28,11 +28,15 @@ export default TestModule.extend({
     }
 
     callbacks.store = function(){
+      var container = this.container;
+
       return container.lookup('store:main');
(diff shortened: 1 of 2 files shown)


dgeb:
Thanks @rwjblue and @bmac <3

I'll update ember-mocha soon.

subhashb:
@rwjblue and @dgeb,
Should this fix be ported to `ember-cli-qunit` as well? I faced the same issue with `"ember-cli-qunit": "0.3.9"` and worked around it by creating the model using `store.createRecord` instead of `this.subject()`

PR #32, opened 2015-03-28
Asynchronous module hooks

ef4:
This makes it possible for module hooks to be asynchronous.

My immediate use case for this is that the only way to do an asynchronous action in a mocha `beforeEach` hook is to return a promise from said hook.

I'm submitting corresponding PRs to ember-mocha and ember-qunit.


rwjblue:
:+1:

PR #33, opened 2015-04-02
Register some built in Ember Data objects if ED exists on the page.

bmac:
This will make it easier to test serializers without having to know
about the transform's Ember Data registers under the hood.

Not sure the best way to test this without adding Ember Data as a dependency.

### DIFF
--- a/lib/ember-test-helpers/isolated-container.js
+++ b/lib/ember-test-helpers/isolated-container.js
@@ -66,6 +66,18 @@ export default function isolatedContainer(fullNames) {
   container.register('view:select', Ember.Select);
   container.register('route:basic', Ember.Route, { instantiate: false });
 
+  var globalContext = typeof global === 'object' && global || self;
+  if (globalContext.DS) {
+    var DS = globalContext.DS;
+    registry.register('transform:boolean', DS.BooleanTransform);
+    registry.register('transform:date', DS.DateTransform);
+    registry.register('transform:number', DS.NumberTransform);
+    registry.register('transform:string', DS.StringTransform);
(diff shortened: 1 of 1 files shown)


lib/ember-test-helpers/isolated-container.js
```diff
@@ -66,6 +66,18 @@ export default function isolatedContainer(fullNames) {
   container.register('view:select', Ember.Select);
   container.register('route:basic', Ember.Route, { instantiate: false });
 
+  var global = typeof global === 'object' && global || window;
```
rwjblue: Because `var global` is hoisted, I don't think you can make this `typeof global` check it should work if you choose a different local variable name though.  Also, `self` should be used instead of `window` (as it works in WebWorkers and normal browser scenarios, but `window` does not work in WebWorkers).

I'm thinking of something like:

```javascript
var globalContext = typeof global === 'object' && global || self;
```

bmac: Will do.

bmac: Updated thanks for the feedback.

rwjblue:
LGTM

@dgeb - Any objections?

dgeb:
I could live with this for now, but it just highlights how backwards our isolatedContainer implementation is now. 

We should straighten this out before ember 2.0. I've promised an Ember RFC on containers / registries and will take this into account.

In the mean time, I will merge.

rwjblue:
Thanks @dgeb!  Also, I am really getting tired of the whack-a-mole here, and definitely am looking forward to your RFC.

PR #42, opened 2015-05-03
Test coverage for `render` during component integration test setup

ef4:
Assuming [this PR](https://github.com/rwjblue/ember-qunit/pull/167) lands in ember-qunit, we can support  the use of `this.render` during component integration test setup hooks, as discussed in the comments to https://github.com/switchfly/ember-test-helpers/pull/38.

This PR updates the embedded copy of ember-qunit's test function to match that change, and adds test coverage.


ef4:
This got rolled into #60.

PR #44, opened 2015-05-06
Add container items required for Glimmer.

rwjblue:
This is getting pretty annoying. We need a *much* better way to deal with these default/fallback container items, but we also need tests to work in Glimmer-land.  So, grrr, but YOLO...

### DIFF
--- a/lib/ember-test-helpers/isolated-container.js
+++ b/lib/ember-test-helpers/isolated-container.js
@@ -67,6 +67,16 @@ export function isolatedRegistry(fullNames) {
   container.register('view:select', Ember.Select);
   container.register('route:basic', Ember.Route, { instantiate: false });
 
+  // added in Glimmer
+  container.register('component:-link-to', Ember.LinkView);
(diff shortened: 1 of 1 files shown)


rwjblue:
https://github.com/emberjs/ember.js/pull/11041 was added to expose `_LegacyEachView`.

dgeb:
> This is getting pretty annoying. We need a much better way to deal with these default/fallback container items, but we also need tests to work in Glimmer-land.

No doubt. I am hopeful that we can find a way out of the woods while implementing  https://github.com/emberjs/rfcs/pull/46. However, if there isn't sufficient support for that, we will need a more narrowly focused alternative to solve registry initialization in Ember (and not here).

dgeb:
@rwjblue btw, you probably realize this, but this is still failing on canary.

rwjblue:
Yep, it required the upstream PR to be merged and published.  Should be good now (took a while for Travis to run on the Ember merge commit).

dgeb:
:+1:

Issue #46, opened 2015-05-08
this.subject() throws error in moduleForModel if { integration: true }


mitchlloyd:
I have a test where I was passing `{ integration: true }` in order to use a model with a hasMany relationship. After upgrading from 0.3.1 to 0.3.2, running `this.subject()` throws an error:

```
TypeError: Cannot read property 'createRecord' of undefined
```

I've found that when setting the integration flag, `this.container.lookup('store:main')` returns `undefined`. Without the flag, `this.container.lookup('store:main')` returns a store object.

rwjblue:
This is because `modelForModel` with `integration: true` is not running the initializers (or the setup code from `isolatedContainer`).

I think we should be adding [all of these registrations](https://github.com/switchfly/ember-test-helpers/blob/0a514b2367c19cf6f67728629617b63c55128dfc/lib/ember-test-helpers/isolated-container.js#L57-L94) from [inside the `_setupIntegrationContainer](https://github.com/switchfly/ember-test-helpers/blob/0a514b2367c19cf6f67728629617b63c55128dfc/lib/ember-test-helpers/test-module.js#L222-L238) in addition to calling `Ember.Application.buildRegistry`.

rwjblue:
We should extract those registrations into a stand alone function, then call it from both `isolatedContainer` and `_setupIntegrationContainer`.

mitchlloyd:
Thanks for the quick reply! For now I've got a workaround using `DS._setupContainer(this.container)`.

wagenet:
@rwjblue you also said this was the same as the "missing component:-text-area" stuff, right?

rwjblue:
Yes

nikz:
Added a failing test for this in #51

mitchlloyd:
Does it seem reasonable that Ember and Ember Data could maintain this list of registrations?

For instance, in Ember, this minimal set of container items could be extracted from the [application `buildRegistry` method](https://github.com/emberjs/ember.js/blob/257c2bc8dcc4c6f418abf6b1142c688d9dcee0c7/packages/ember-application/lib/system/application.js#L993) and be exposed as a public API.

I know that is beyond the scope of this bug, but I just wanted to see if this was a decent idea.

rwjblue:
[...]

Issue #41, opened 2015-05-02
TestModuleForComponent, integration: true, and {{link-to}} incompatible


jamesarosen:
I have a very simple component:
```hbs
{{#if (some-helper-that-relies-on-the-session)}}
  {{#link-to 'index'}}Home{{/link-to}}
{{/if}}
```
and a very simple test:
```js
componentTest('home-link', 'HomeLinkComponent', {
  integration: true,

  beforeEach: function() {
    App = startApp();
    component = this.subject();
    this.render();
  },

  afterEach: function() {
    Ember.run(App, 'destroy');
    App = null;
  }
});

test('anything, really', function(assert) {
  assert.ok(true);
});
```

With those three factors together, I get `Cannot read property 'recognizer' of undefined` and `Something you did caused a view to re-render after it rendered but before it was inserted into the DOM.`

If I remove the link, the test passes. If I turn off `integration: true`, the test passes.

In this case, I need `integration: true` because the helper requires the basic app infrastructure.

jamesarosen:
This seems related to https://github.com/eviltrout/ember-performance/issues/34

jamesarosen:
I've opened https://github.com/ember-cli/ember-cli/issues/3995 as a possible solution since it doesn't seem to me that `TestModuleForComponent` should really be responsible for starting routing.

jamesarosen:
It's possible this will be -- or could be -- resolved by #38.

technomage:
Still seeing this after the merge of #38.

mitchlloyd:
Just a note that this issue with `{ integration: true }` and `{{link-to}}` remains after the merge of #49.

rwjblue:
@mitchlloyd - Yep, I tested just a bit ago.  It is getting closer though.... :smile:

jeremywrowe:
I am seeing the same thing with `{ integration: true }` `{{input}}`

here is the error:
```
Uncaught Error: Assertion Failed: HTMLBars error: Could not find component named "-text-field" (no component or template with that name was found)
```

Hope it helps. I am very excited to see this style of testing for components moving forward!

MattMSumner:
I ran into this today. Adding `import Router from '../../router'`fixed it. Is this expected?

MattMSumner:
Is there a solution for this yet? I'm loving `{ integration: true }` component tests but also using `link-to`.

rwjblue:
Should be working with beta of ember and 0.4.0 of ember-qunit

[...]

PR #49, opened 2015-05-14
Remove isolatedContainer and use Ember.Application.buildRegistry.

rwjblue:
The isolatedContainer was way to finicky and caused breakage any time Ember
or Ember Data decided to change what the default contents should be.

This replaces the `isolatedContainer` with the container built by
`Ember.Application.buildRegistry` (or `.buildContainer`) which has been
available since 1.0.0.  `.buildRegistry` does result in a container that is
fully fleshed out but it also has a resolver attached, so in the
non-integration case we replace that resolver with an empty function.


rwjblue:
This should fix #46 and possibly #41.

rwjblue:
FYI - this test suite checks Ember 1.10, 1.11, release (1.12), beta (future 1.13), and canary (future 2.0.0).

@dgeb - I'd like your feedback/review on this one...

rwjblue:
@dgeb - Ping?

dgeb:
@rwjblue sorry for the delay - will review tonight or tomorrow morning!

dgeb:
@rwjblue what do you think of exporting separate methods `buildRegistry` and `buildContainer`, where `buildContainer` takes a `registry` as an argument?

rwjblue:
@dgeb - Seems OK to me, what would the purpose/result be?  Just to replace the `isolatedContainer` export?

dgeb:
@rwjblue I was thinking that it might be useful to be able to build multiple containers from the same registry. But anyway, this is all internal at this point, and my preference is to [reform all of this](https://github.com/emberjs/rfcs/pull/46) so that tests don't even deal with registries and containers, so I'm :+1: if you just want to merge now and refactor later ...

rwjblue:
@dgeb - Thanks for reviewing!

PR #60, opened 2015-06-14
Compatibility with Ember 1.13+

ef4:
This cleans up deprecations for use with Ember 1.13+, plus one small
private API break (the changed `Resolver.create()` call).

The removed tests were all testing container resolution of things that
are deprecated. There is still coverage for container resolution of
Components, which should be sufficient.

The test suite still emits a several deprecations, but they are all
coming from Ember Data and will need to get fixed there.

I worked on this PR with @stefanpenner.

This PR also supersedes #42, which was ready to go and relevant.

### DIFF
--- a/tests/test-module-for-component-test.js
+++ b/tests/test-module-for-component-test.js
@@ -13,7 +13,7 @@ var PrettyColor = Ember.Component.extend({
   classNames: ['pretty-color'],
   attributeBindings: ['style'],
   style: function(){
-    return 'color: ' + this.get('name') + ';';
+    return new Ember.Handlebars.SafeString('color: ' + this.get('name') + ';');
   }.property('name')
 });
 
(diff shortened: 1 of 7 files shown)


stefanpenner:
LGTM

rwjblue:
Why are all the tests failing?

ef4:
Ha, because I leaked ES6, and it happened to work natively in Chrome.

ef4:
Fixed, Travis is green.

tests/test-module-for-component-test.js
```diff
@@ -114,65 +114,6 @@ test('can lookup components in its layout', function() {
...
-  expect(1);
-  var component = this.subject({
-    colors: ['red', 'green', 'blue'],
-    layout: Ember.Handlebars.compile("{{view 'select'}}")
-  });
-  this.render();
-  equal(component._state, 'inDOM');
-});
-
-test('can lookup default Ember.Views in its layout', function() {
```
rwjblue: We should replace this one with a {{component}} helper version (not blocking this PR though).

ef4: Sounds good, I will work on that, since I also noticed another incompatibility with the new deprecations.

PR #61, opened 2015-06-15
Don't replace Ember's internal view:toplevel

ef4:
This fixes another deprecation in Ember 1.13. ember-test-helpers was
replacing `view:toplevel` with `Ember.View`, which is deprecated. We
should instead just leave `view:toplevel` alone, because the internal
version is not deprecated.


ef4:
@rwjblue this is the remaining open PR I mentioned yesterday, it's the last one I needed to get everything working cleanly with 1.13.

PR #62, opened 2015-06-15
Add a test for {{component}} helper

ef4:
This replaces the `{{view}}` helper test coverage that I removed in #60 with a test of the `{{component}}` helper instead.

This makes the test suite incompatible with Ember 1.10, which seems reasonable given that we've now released 1.13, so the list of supported releases remains the same length.


rwjblue:
:+1:

PR #66, opened 2015-06-24
Additional dom testing events.

blimmer:
I was originally trying to prove that something was broken, but it looks like it's related to ember-cli-mocha, which I'm in the process of writing an issue for. I figured I would still offer these additional tests I wrote to get more coverage.

If these tests are not useful, feel free to close this.


rwjblue:
Thanks!  I love more tests!

PR #68, opened 2015-07-01
allow getProperties and setProperties for component integration tests

CodeOfficer:
Is this worth testing? O.o

### DIFF
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -129,8 +129,19 @@ export default TestModule.extend({
       });
     };
 
+    context.setProperties = function(hash) {
+      Ember.run(function() {
+        Ember.setProperties(context, hash);
+      });
+    };
+
     context.get = function(key) {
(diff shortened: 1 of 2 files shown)


rwjblue:
> Is this worth testing?

Only if you don't want it removed as "unused" in a future refactor :stuck_out_tongue_closed_eyes:

rwjblue:
Also, :+1: on this, maybe we should add `getProperties` too?

pangratz:
I was about to submit such a PR. I think this is a useful addition.

> maybe we should add getProperties too?

:+1:

CodeOfficer:
that work for ya?

rwjblue:
Test failures were fixed in https://github.com/switchfly/ember-test-helpers/pull/79, restarting Travis build.

rwjblue:
Thank you!

rwjblue:
Released in ember-qunit@0.4.2.

pangratz:
Thanks @CodeOfficer

PR #71, opened 2015-07-02
Ember.keys is deprecated in favor of Object.keys

jpadilla:
https://github.com/emberjs/ember.js/pull/11459 marks `Ember.keys` for deprecation in favor of Object.keys.

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -192,7 +192,7 @@ export default Klass.extend({
     this.cache = this.cache || {};
     this.cachedCalls = this.cachedCalls || {};
 
-    var keys = Object.keys(callbacks);
+    var keys = (Object.keys || Ember.keys)(callbacks);
 
     for (var i = 0, l = keys.length; i < l; i++) {
       (function(key) {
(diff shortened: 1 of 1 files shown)


lib/ember-test-helpers/test-module.js
```diff
@@ -192,7 +192,7 @@ export default Klass.extend({
     this.cache = this.cache || {};
     this.cachedCalls = this.cachedCalls || {};
 
-    var keys = Ember.keys(callbacks);
+    var keys = Object.keys(callbacks);
```
rwjblue: Can you make this fall back to Ember.keys if Object.keys is not present? If we don't do that, we break older browser compat for Ember 1.13...

jpadilla: @rwjblue would something like this do?

```javascript
var keys = ('keys' in Object ? Object.keys : Ember.keys)(callbacks);
```

rwjblue: `(Object.keys || Ember.keys)(callbacks)` should do the trick, I think.

jpadilla:
@rwjblue done

PR #72, opened 2015-07-02
Call setupRouter()

nathanhammond:
The router instance is never set up, though it is expected to be inside of the Ember Router at runtime. Related to [emberjs/ember.js#11610](https://github.com/emberjs/ember.js/issues/11610).


nathanhammond:
Different solution for 11610.

PR #73, opened 2015-07-02
Guard against exceptions in the setup steps.

winding-lines:
When there are exceptions in the setup steps (for example needs component missing)
the implementation is not properly capturing those exceptions.
This fix uses the slighlty longer syntax for creating promises which
wraps the call in a function and allows RSVP to properly capture
the exception. When using the shorter RSVP.resolve() form the exception
happens outside (before) the RSVP invocation and cannot be captured.

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -107,7 +107,8 @@ export default Klass.extend({
     function nextStep() {
       var step = steps.shift();
       if (step) {
-        return Ember.RSVP.resolve(step.call(context)).then(nextStep);
+        // guard against exceptions, for example missing components referenced from needs.
+        return new Ember.RSVP.Promise(function(ok) {ok(step.call(context));}).then(nextStep);
       } else {
         return Ember.RSVP.resolve();
       }
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -108,7 +108,9 @@ export default Klass.extend({
       var step = steps.shift();
       if (step) {
         // guard against exceptions, for example missing components referenced from needs.
-        return new Ember.RSVP.Promise(function(ok) {ok(step.call(context));}).then(nextStep);
+        return new Ember.RSVP.Promise(function(ok) {
+          ok(step.call(context));
+        }).then(nextStep);
       } else {
         return Ember.RSVP.resolve();
       }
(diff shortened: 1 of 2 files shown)


lib/ember-test-helpers/test-module.js
```diff
@@ -107,7 +107,8 @@ export default Klass.extend({
     function nextStep() {
       var step = steps.shift();
       if (step) {
-        return Ember.RSVP.resolve(step.call(context)).then(nextStep);
+        // guard against exceptions, for example missing components referenced from needs.
+        return new Ember.RSVP.Promise(function(ok) {ok(step.call(context));}).then(nextStep);
```
chadhietala: @winding-lines Can you format this to

```
Ember.RSVP.Promise(function(ok) {
  ok(step.call(context));
}).then(nextStep)
```

chadhietala: Also wouldn't this just work.

```
  return Ember.RSVP.resolve(step.call(context)).then(nextStep, function(err) {
   throw err;
  })
```

winding-lines: @chadhietala your proposal does not work. You can rewrite both the original code and your proposal as the following:

```javascript
var current = step.call(context); // throws exception, promise code never reached
return Ember.RSVP.resolve(current).then(..)
```

The call to `step.call()` will happen before the promise, even though it is written on the same line. With the proposed change the promise is active before the throwing call and can setup a try-catch to catch the exception. 

Note that `promise.resolve()` creates a new function anyways so there is no change in performance or memory allocation.

rwjblue: Yes, this is roughly what I was thinking also (though with .catch instead of two args to .then).

winding-lines: @rwjblue a `.catch()

rwjblue: We should use RSVP.resolve().then(function() { return step.call(); })

winding-lines: @rwjblue yes that would also work, however you are 'wasting' a resolve().

rwjblue: Ya,
makes sense

rwjblue:
This looks good, can you add a test for this?

winding-lines:
@rwjblue I've added the tests. Without the improved exception handling the whole test loop breaks when the new test is hit. What we are seeing in production is that the CI builds 'hang'. broken

With the new code the processing continues, in normal operation the test themselves would fail instead of the setup. There is also no 'hanging'. working

winding-lines:
Let me know if I should collapse all the commits in one.

rwjblue:
Thanks for your work on this!

winding-lines:
Thanks for merging!

PR #74, opened 2015-07-06
Update test-module-for-model.js

quaertym:
This is to avoid deprecation warning:
```
You tried to look up 'store:main', but this has been deprecated in favor of 'service:store'.
```


rwjblue:
:+1:

PR #78, opened 2015-07-17
Throw when integration and needs declared

bcardarella:
`needs` will switch off integration testing, this is not obvious. If
both are declared we should throw a descriptive Error.

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -18,6 +18,10 @@ export default Klass.extend({
     this.name = description || subjectName;
     this.callbacks = callbacks || {};
 
+    if (this.callbacks.integration && this.callbacks.needs) {
+      throw new Error("cannot declare 'inegration: true' and 'needs' in the same module");
+    }
+
     if (this.callbacks.integration) {
       this.isIntegration = callbacks.integration;
       delete callbacks.integration;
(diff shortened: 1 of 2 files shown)


bcardarella:
The current test suite is failing and I didn't have time to look into the error that already exist before writing this patch. Once those are resolved I can revisit this patch if there are build issues.

rwjblue:
Sorry for the delay here, I think that we have everything fixed up test wise on master.  Would you mind rebasing to get Travis to re-run with the updates?

bcardarella:
whoops

lib/ember-test-helpers/test-module.js
```diff
@@ -18,6 +18,10 @@ export default Klass.extend({
     this.name = description || subjectName;
     this.callbacks = callbacks || {};
 
+    if (this.callbacks.integration && this.callback.needs) {
```
rwjblue: Should be `this.callbacks.needs` (it is `this.callback` ATM).

tests/test-module-test.js
```diff
@@ -159,3 +159,20 @@ test("needs is not needed (pun intended) when integration is true", function() {
   var otherComponent = this.container.lookup('component:not-the-subject');
   ok(otherComponent, 'another component can be resolved when integration is true');
 });
+
+test("throws an error when declaring integration: true and needs in the same module", function() {
+  expect(1);
```
rwjblue: (sorry) - It looks like 2 assertions are made in the `beforeSetup` just above here causing this to error (because we actually make 3 assertions).

rwjblue:
Thank you!

Issue #80, opened 2015-07-20
Does not work with Ember 2.0.0 beta 3


martndemus:
- It uses `Ember.View`, which is (soft-)removed since beta.3
- Tests need `ember-data#canary` because `ember-data < 2` uses `ArrayPolyfills`

I'm cooking up a PR to fix this. Help is definitely welcome.

martndemus:
ContainerView uses `{{view}}`, see emberjs/ember.js#11818

rwjblue:
https://github.com/switchfly/ember-test-helpers/pull/79 fixes some of this. But I'm still figuring out how to remove container view usage.

martndemus:
This might be naive but this is what fixes it for me:

```diff
this.callbacks.render = function() {
-      var containerView = Ember.ContainerView.create({container: container});
+      var subject;
+
       Ember.run(function(){
-        var subject = context.subject();
-        containerView.pushObject(subject);
-        containerView.appendTo('#ember-testing');
+        subject = context.subject();
+        subject.appendTo('#ember-testing');
       });
 
       _this.teardownSteps.unshift(function() {
         Ember.run(function() {
-          Ember.tryInvoke(containerView, 'destroy');
+          Ember.tryInvoke(subject, 'destroy');
         });
       });
     };
```

martndemus:
#79, #81, #82, and #83 together should fix all problems.

rwjblue:
>  I can't figure out how to use an older version of ED with ember-try in the 1.11 scenario.

We can't.  ember-try only allows configuration of `dependencies`, not `devDependencies`.  This could probably be fixed in ember-try...

martndemus:
Lets give that a try ;)

rwjblue:
Fixed by https://github.com/switchfly/ember-test-helpers/pull/79 https://github.com/switchfly/ember-test-helpers/pull/80 and https://github.com/switchfly/ember-test-helpers/pull/81.

PR #82, opened 2015-07-20
Stop using ContainerView

martndemus:
See #80


rwjblue:
Pulled into https://github.com/switchfly/ember-test-helpers/pull/79.

PR #83, opened 2015-07-20
Future proof for Ember 2

martndemus:
See #80


martndemus:
Closing, better fix already in #79.

PR #85, opened 2015-07-21
Only override container methods if present.

rwjblue:
The intent of this code initially was to suppress deprecation messages that could not be fixed by the test suite.

In Ember 2.0.0 these deprecated methods are being removed, this change ensures that we only expose them when the container already has them.

See https://github.com/emberjs/ember.js/pull/11853 for details.


dgeb:
Thanks @rwjblue :+1:

PR #87, opened 2015-07-22
ensure we correctly cleanup the context

stefanpenner:
- [x] pending confirmation in my setup

### DIFF
--- a/lib/ember-test-helpers/test-context.js
+++ b/lib/ember-test-helpers/test-context.js
@@ -7,3 +7,7 @@ export function setContext(context) {
 export function getContext() {
   return __test_context__;
 }
+
+export function unsetContext() {
+  __test_context__ = undefined;
+}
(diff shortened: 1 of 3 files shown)


lib/ember-test-helpers/test-context.js
```diff
@@ -7,3 +7,7 @@ export function setContext(context) {
 export function getContext() {
   return __test_context__;
 }
+
+export function unsetContext() {
```
rwjblue: Why do we need another method here? Could you call 'setContext(undefined)'?

stefanpenner: wanted it to be very obvious that It was specifically to unset, not just an overloaded the existing set.

rwjblue: Gotcha, :+1:

chadhietala:
@stefanpenner :+1: Thanks for dedicating your Mondays to fixing leaks.

stefanpenner:
Maybe I should spend Tuesday's learning how to spell:p

stefanpenner:
@rwjblue i think this is good to go.

rwjblue:
Released in:

* ember-test-helpers@0.5.5
* ember-qunit@0.4.5

PR #90, opened 2015-07-28
Consistently use the new view registry

ef4:
Internally, the `EventDispatcher` looks for the global view registry on the container and then falls back to using `View.views`. Since we were creating it directly instead of via the container, it had no container reference, and so it was doing all view tracking using the fallback. 

This is problematic, because the global fallback path is now only accessible from inside Ember (by design, because we're trying to eliminate it.)

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -172,7 +172,7 @@ export default Klass.extend({
     this.context = undefined;
     unsetContext();
 
-    if (context.dispatcher) {
+    if (context.dispatcher && !context.dispatcher.isDestroyed) {
       Ember.run(function() {
         context.dispatcher.destroy();
       });
@@ -229,6 +229,7 @@ export default Klass.extend({
 
     this.container = items.container;
     this.registry = items.registry;
+
 
     var thingToRegisterWith = this.registry || this.container;
     var router = resolver.resolve('router:main');
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -59,7 +59,7 @@ export default TestModule.extend({
       thingToRegisterWith.injection(this.subjectName, 'layout', layoutName);
     }
 
-    context.dispatcher = Ember.EventDispatcher.create();
+    context.dispatcher = this.container.lookup('event_dispatcher:main') || Ember.EventDispatcher.create();
     context.dispatcher.setup({}, '#ember-testing');
 
     this.callbacks.render = function() {
@@ -96,9 +96,16 @@ export default TestModule.extend({
 
     this.actionHooks = {};
 
-    context.dispatcher = Ember.EventDispatcher.create();
+    context.dispatcher = this.container.lookup('event_dispatcher:main') || Ember.EventDispatcher.create();
     context.dispatcher.setup({}, '#ember-testing');
     context.actions = module.actionHooks;
+
+    // This thing is registered as "view:" instead of "component:"
+    // because only views get the necessary _viewRegistry
+    // injection. Which is arguably an Ember bug, but doesn't impact
+    // normal app usage since apps always use a "view:" at the top
+    // level.
+    (this.registry || this.container).register('view:test-holder', Ember.Component.extend());
 
     context.render = function(template) {
       if (!template) {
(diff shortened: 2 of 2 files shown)


lib/ember-test-helpers/test-module.js
```diff
@@ -230,6 +230,13 @@ export default Klass.extend({
     this.container = items.container;
     this.registry = items.registry;
 
+
+    var viewRegistry = this.container.lookup('-view-registry:main');
+    if (viewRegistry) {
+      var View = Ember.__loader.require('ember-views/views/view').default;
+      View.views = viewRegistry;
```
rwjblue: Looking up the `EventDispatcher` from the container (as you changed above), should be sufficient to ensure that it internally has access to `-view-registry:main`, we should not need to set the internal `Ember.View.views`.

ef4: No, this was necessary. Ember is internally using the View.views list still, and it only copies that list out of the container during app instance initialization.

Our integration tests don't go down that initialization path.

rwjblue: In all cases Ember should be using `-view-registry:main` then falling back to its internal `View.views`.  The `-view-registry:main` is setup during [Ember.Application.buildRegistry](https://github.com/emberjs/ember.js/blob/1d7b833664292bc6644adc41a7dd9a842ea078c5/packages/ember-application/lib/system/application.js#L889) (injection is setup [directly below](https://github.com/emberjs/ember.js/blob/1d7b833664292bc6644adc41a7dd9a842ea078c5/packages/ember-application/lib/system/application.js#L891)) which we are calling from [here](https://github.com/switchfly/ember-test-helpers/blob/3315c6de26b80d0520e05862df46954e82f2e886/lib/ember-test-helpers/build-registry.js#L44).

Where in Ember are we still using `View.views` without checking `-view-registry:main` first?

ef4: Ah, I think I see the problem. ember-test-helpers is doing a `Component.create`, and although it passes in a container, [that is insufficient.](https://github.com/emberjs/ember.js/blob/df9fba3f19874d9f64d7d176be4803bbf8223278/packages/ember-views/lib/views/view.js#L1323)

To play nicely with dependency injection, ember-test-helpers should be doing lookupFactory instead. Is there a generic component factory available? Something equivalent to the old `view:default`?

rwjblue: No, I don't think something like that exists, but the following code should work for us [here](https://github.com/switchfly/ember-test-helpers/blob/704e91f826f1a5ef1dcb50f541f3d5e188676492/lib/ember-test-helpers/test-module-for-component.js#L113):

```javascript
var thingToRegisterWith = this.registry || this.container;
thingToRegisterWith.register('component:-template-under-test', Ember.Component.extend({
  layout: template
});
module.component = this.container.lookup('component:-template-under-test');
```

ef4: This gets weirder. Even if I use lookupFactory, `_viewRegistry` is only injected onto things registered as `view:foo`, not things registered as `component:foo`. That seems like a potential Ember bug, although it doesn't effect a normal app since they use `view:toplevel` (not `component:toplevel`, and then each view inherits a viewRegistry from its parent.

rwjblue: Yeah, I agree that this should be fixed upstream in Ember (we should inject `-view-registry:main` into both `view` and `component`), but the solution with the smallest amount of change might just be:

```javascript
module.component = Ember.Component.create({
        layout: template,
        container: module.container,
        _viewRegistry: module.container.lookup('-view-registry:main')
      });
``` 

We should still fix that injection upstream though....

rwjblue:
I'd like a test that we can use to help prevent regressions here.  Seems like this bug prevents testing of event dispatcher handled events, right?  I can try to PR a test against your branch if you'd like...

ef4:
We actually have test coverage for event dispatching, and the bug wasn't breaking the tests because it was canceling itself out. All the views were being placed on View.views, and that's where the event dispatcher was expecting to find them.

I only noticed a problem due to ember-mobiletouch failing to find `this.container` in the event dispatcher.

So I don't see a good test for this condition that doesn't peek into internals and test things that don't really matter to users. This issue was really about upgrading the test helpers to use newer internals.

PR #93, opened 2015-07-31
[BUGFIX] Pass in model name to `store.adapterFor`

seanpdoyle:
Silences the following deprecation:

DEPRECATION: Passing classes to store methods has been removed.
Please pass a dasherized string instead of undefined
[deprecation id: ds.store.passing-classes-deprecated]


seanpdoyle:
ping @rwjblue

PR #96, opened 2015-08-19
Allow overriding services/factories in the registry.

rwjblue:
By default in Ember things that are automatically resolved take precedence over things that are manually registered. This is somewhat counterintuitive to many people, as they would like to be able to override a service or other factory in a given set of tests (for example to use a mock analytics service).

The ability to do this has existed since the SSR/Fastboot work that split the registry from the container in 1.11, but I only recently realized how I could leverage that work here.

This change adds new functionality on Ember 1.11+, but allows older versions to continue to work as they do today (aka non-breaking).

---

Example:

Assuming you have a `app/services/analytics.js` and want to use a mock service for a set of component integration tests, you might do something like this:

```javascript
import MockAnalyticsService from '../mock-services/analytics';
import { moduleForComponent, test } from 'ember-qunit';

moduleForComponent('something-cool', {
  integration: true,
  beforeEach() {
    this.registry.register('service:analytics', MockAnalyticsService);
  }
});
```

In the above scenario, the mock service would be used instead of the one at `app/services/analytics.js`.

### DIFF
--- a/lib/ember-test-helpers/build-registry.js
+++ b/lib/ember-test-helpers/build-registry.js
@@ -27,7 +27,7 @@ function exposeRegistryMethodsWithoutDeprecations(container) {
 }
 
 export default function(resolver) {
-  var registry, container;
+  var fallbackRegistry, registry, container;
   var namespace = Ember.Object.create({
     Resolver: { create: function() { return resolver; } }
   });
@@ -41,10 +41,12 @@ export default function(resolver) {
   }
 
   if (Ember.Application.buildRegistry) {
-    registry = Ember.Application.buildRegistry(namespace);
-    registry.register('component-lookup:main', Ember.ComponentLookup);
+    fallbackRegistry = Ember.Application.buildRegistry(namespace);
+    fallbackRegistry.register('component-lookup:main', Ember.ComponentLookup);
 
-    registry = registry;
+    registry = new Ember.Registry({
+      fallback: fallbackRegistry
+    });
     container = registry.container();
     exposeRegistryMethodsWithoutDeprecations(container);
   } else {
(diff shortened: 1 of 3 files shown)


rwjblue:
I stole this from recent work @dgeb did with container / registry reform in Ember itself, and confirmed that the fallback registry works properly in Ember 1.11+.

@dgeb - Would you mind reviewing?

dgeb:
Great work @rwjblue!

rondale-sc:
Does this only apply to component integration tests?  Or can this be more widely applied to acceptance tests?

rondale-sc:
I was unable to reference `this.registry` from an acceptance test is why I ask.  btw.

rwjblue:
Acceptance tests generally do not use ember-qunit, so this doesn't do anything for them. 

More work to come for acceptance tests...

rondale-sc:
Right on!  Side note, this is awesome!

tim-evans:
Could this be extended to stub out modules as well? (`this.requirejs.entries`)

rwjblue:
@tim-evans - Generally, I would rather not do that. What is your use-case?

tim-evans:
Mocking out functions that are the only export from a file.

rwjblue:
If the functions are looked up via the container/resolver, it will already work...

tim-evans:
Hmm... I'll test that! Seems like a nice way to do that. (They're currently using `import`)

mixonic:
This requires documentation.

rwjblue:
@mixonic - TBH, I'm not sure we can document it.  AFAIK, registry is private.

rwjblue:
Also, this project contains no documentation (though we could document in the ember-qunit README which is a bit more fleshed out).

PR #98, opened 2015-09-09
Fix typo in thrown error

HeroicEric:
Fixes a typo in the error message when declaring both `integration: true` and `needs` in the same module


rwjblue:
Nope, not a bug.  We are just ahead of the rest of the english language speakers.

HeroicEric:
:smile:

HeroicEric:
Build is failing on ember-canary but I don't think it's at all related to this change

rwjblue:
Confirm, not related. I did rerun just to see if it is a transient failure.  Would you mind opening an issue so we can track it down after merging this?

Issue #99, opened 2015-09-09
[CANARY] Test suite fails against ember-canary 2.2.0


HeroicEric:
The test suite is failing on ember-canary.

A build with the failures can be seen at https://travis-ci.org/switchfly/ember-test-helpers/builds/79391044

Relevant section of the build failing is:

```
not ok 21 PhantomJS 1.9 - component:Component Integration Tests: it can handle actions
    ---
        actual: >
            null
        message: >
            Died on test #1     at http://localhost:7357/testem/qunit.js:251
                at test (http://localhost:7357/assets/ember-test-helpers-tests.amd.js:1764)
                at http://localhost:7357/assets/ember-test-helpers-tests.amd.js:1199
                at http://localhost:7357/assets/loader.js:118
                at tryFinally (http://localhost:7357/assets/loader.js:21)
                at http://localhost:7357/assets/loader.js:124
                at http://localhost:7357/assets/test-support.js:41: Assertion Failed: The action 'didFoo' did not exist on [object Object]
        Log: >
    ...
ok 22 PhantomJS 1.9 - component:Component Integration Tests: it accepts precompiled templates
ok 23 PhantomJS 1.9 - component:Component Integration Tests: it supports DOM events
ok 24 PhantomJS 1.9 - component:Component Integration Tests: it supports updating an input
not ok 25 PhantomJS 1.9 - component:Component Integration Tests: it supports dom triggered focus events
    ---
        actual: >
            
        expected: >
            init
        Log: >
    ...
```

PR #100, opened 2015-09-09
Do not register a router service for Ember <1.13

mixonic:
Versions of Ember prior to 1.13 require a full router booted by the app, or no router at all. This ensures they receive no router at all instead of the 1.13-safe one. Result is that the link-to helper can be used in Ember 1.11 and 1.12 component tests (unit and integration).

See also these PRs that attempted to fix the issue in Ember:

* https://github.com/emberjs/ember.js/pull/11829
* https://github.com/emberjs/ember.js/issues/11825

IMO, it is more appropriate that it be fixed here, since ember-test-helpers is basically dealing with fragile/intimate API surface when it sets up a router in the container.

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -3,6 +3,7 @@ import { getContext, setContext, unsetContext } from './test-context';
 import { Klass } from 'klassy';
 import { getResolver } from './test-resolver';
 import buildRegistry from './build-registry';
+import hasEmberVersion from './has-ember-version';
 
 export default Klass.extend({
   init: function(subjectName, description, callbacks) {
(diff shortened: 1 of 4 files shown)


lib/ember-test-helpers/test-module.js
```diff
@@ -4,6 +4,13 @@ import { Klass } from 'klassy';
 import { getResolver } from './test-resolver';
 import buildRegistry from './build-registry';
 
+function hasEmberVersion(major, minor) {
```
rwjblue: Can you move this to its own file?  We have a scenario in tests (for closure actions) where we are using a regexp to determine if a given test should run, and this is a much cleaner solution.

Something like `/lib/ember-test-helpers/matches-ember-version.js`?

rwjblue:
Can you confirm that the test added here failed prior to the changes?

rwjblue:
BTW - The failure in Travis is for canary builds and is being tracked in https://github.com/switchfly/ember-test-helpers/issues/99.

mixonic:
Confirmed that the only failure here is the one in #99 

screenshot

rwjblue:
Thank you!

PR #104, opened 2015-09-27
prepend tear down steps like tearDownForComponent in the begining of tearDownSteps array

tsing80:
Currently,   [component.destroy()](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module-for-component.js#L211)  is only run after  #ember-testing DOM element is [emptied](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module.js#L188). When the component which use third party library  like jquery-ui  or bootstrap javascript component do cleanup using  willDestroyElement() hook, it cannot cleanup properly  because the element is already removed from DOM. In additional , teardownStep should be run in reverse order as setupSteps in general. So this PR also fix the [_resetViewRegistry()](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module-for-component.js#L55) hook in TestModuleForComponent.

### DIFF
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -47,12 +47,12 @@ export default TestModule.extend({
         throw new Error("component integration tests do not support `subject()`.");
       };
       this.setupSteps.push(this.setupComponentIntegrationTest);
-      this.teardownSteps.push(this.teardownComponent);
+      this.teardownSteps.unshift(this.teardownComponent);
     }
 
     if (Ember.View && Ember.View.views) {
       this.setupSteps.push(this._aliasViewRegistry);
-      this.teardownSteps.push(this._resetViewRegistry);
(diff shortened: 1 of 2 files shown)


tsing80:
@rwjblue ping

rwjblue:
Thanks for the reminder! This looks good, but I'd love to have a test that ensures the ordering of teardown is what we expect.  A good place to start is by adding a test similar to [this one](https://github.com/switchfly/ember-test-helpers/blob/master/tests/test-module-test.js#L27-L65) that shows the ordering of all the hooks.  You would need to add a component that pushes a `component-destroyed` entry to the `callbackOrder` when it's destroy was called, then `deepEqual` the final array in `afterTeardown` to ensure it happens properly.

tsing80:
@rwjblue I have added a test based on the one in component unit test

rwjblue:
awesome, thank you!

PR #105, opened 2015-09-30
Add public API for injection & registration

ef4:
In an app, we encourages users to use injection over container.lookup,
and we encourage `application.register` over `registry.register`. But in our integration tests, there is no good alternative to direct `this.container.lookup` and `this.registry.register`.

So I added `register` and `inject` capability directly to the integration test context. `register` is fairly trivial:

```js
    test('registering something', function(assert) {
      this.register('component:x-foo', Ember.Component.extend({
        classNames: ['i-am-x-foo']
      }));
      this.render(hbs`{{x-foo}}`);
      assert.equal(this.$('.i-am-x-foo').length, 1, 'found x-foo');
    });
```

`inject` lets you inject anything into the test context:

```js
    test('injecting a service', function(assert) {
      this.inject.service('notifications');
      this.render(hbs`{{my-notification-viewer}}`);

      // The notifications service has been injected into our test context,
      // so we can access it here and tell it to do something.
      this.get('notifications').receivedNotification("Hello world");

      assert.equal(this.$('.notification').text(), 'Hello world');
    })
```

You can also give the injected thing an alternate name using `as`:

```js
    test('injecting a service with aliased name', function(assert) {
      this.inject.service('notifications', { as: 'messages' });
      this.render(hbs`{{my-notification-viewer}}`);

      // The notifications service has been injected into our test context
      // as this.get('messages').
      this.get('messages').receivedNotification("Hello world");

      assert.equal(this.$('.notification').text(), 'Hello world');
    })
```

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -143,10 +143,22 @@ export default Klass.extend({
       container:  this.container,
       registry: this.registry,
       factory:    factory,
-      dispatcher: null
+      dispatcher: null,
+      register: function() {
+        var target = this.registry || this.container;
+        return target.register.apply(target, arguments);
+      },
+      inject: {}
     });
 
-    this.context = getContext();
+    var context = this.context = getContext();
+
+    Object.keys(Ember.inject).forEach(function(typeName) {
+      context.inject[typeName] = function(name, opts) {
+        var alias = (opts && opts.as) || name;
+        Ember.set(context, alias, context.container.lookup(typeName + ':' + name));
+      };
+    });
   },
 
   setupTestElements: function() {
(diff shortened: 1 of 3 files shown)


lib/ember-test-helpers/test-module-for-component.js
```diff
@@ -188,6 +188,18 @@ export default TestModule.extend({
       }
       hook.apply(module, Array.prototype.slice.call(arguments, 1));
     };
+    context.register = function() {
```
rwjblue: IMHO, these belong in the main `test-module` file, as they are not really related to components...

ef4: Good point. Will do.

pixelhandler:
This looks nice! I've been reaching into the container and using register for a while in tests but it will be great to have a blessed path to doing so :+1:

ef4:
I moved the methods into the base test-context so they're available in all test types, and updated some tests to show that they work.

PR #106, opened 2015-10-02
Guard for missing Ember.inject.

rwjblue:
This is not present on Ember < 1.10.

Also, update ember-try config for Ember 1.10 support.


ef4:
LGTM.

PR #112, opened 2015-10-19
Add `wait` helper.

rwjblue:
This allows component integration tests that need to deal with async interactions to have an easy way to model their tests.

Example test (assuming the component does some async in an action triggered by the button click):

```javascript
import { moduleForComponent, test } from 'ember-test-helpers';

test('it works when async exists in an event/action', function(assert) {
  this.render('{{my-foo}}');

  this.$('button').click();

  return wait()
    .then(() => {
      // assertions for after async behavior
    });
});
```

The guts of this `wait` helper are pulled from the `wait` acceptance test helper in Ember itself. Eventually we may be able to make Ember itself provide an implementation that we can share here and in the acceptance testing world.

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -25,7 +25,7 @@ var lib = pickFiles('lib', {
 
 var tests = pickFiles('tests', {
   srcDir: '/',
-  files: ['test-support/*.js', '*.js'],
+  files: ['**/*.js'],
   destDir: '/tests'
 });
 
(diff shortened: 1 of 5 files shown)


ef4:
:clap:

I think the stock integration test `wait` also waits for any pending ajax, using ajaxstart events on window. That is something we can add here.

rwjblue:
Added AJAX waiting support.

knownasilya:
Looks like an advanced testing tutorial/video might be nice, since there have been so many testing changes recently.

adong:
I just want to point out that `wait` helper has been removed in https://github.com/emberjs/ember-test-helpers/pull/818. 😢

rwjblue:
Long live `settled`! :smile:

PR #115, opened 2015-10-28
Add support for `integration: 'legacy'`

hjdivad:
This semantics of the `integration` flag have changed in a
backwards-incompatible way since v0.4.3.  This corresponds to an ember upgrade
(using ember-cli) between v1.12 and v1.13.

In order to ease this upgrade path for ember-cli users, `integration: 'legacy'`
will set up a component test with v0.4.3 semantics.  Of course, newer tests
would still be encouraged to be written with `integration: true`.

### DIFF
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -1,4 +1,4 @@
-import TestModule from './test-module';
+import { default as TestModule, UNIT_TEST, LEGACY_INTEGRATION_TEST, INTEGRATION_TEST } from './test-module';
 import Ember from 'ember';
 import { getResolver } from './test-resolver';
 
@@ -14,11 +14,7 @@ export default TestModule.extend({
 
     this.componentName = componentName;
 
-    if (callbacks.needs || callbacks.unit || callbacks.integration === false) {
-      this.isUnitTest = true;
-    } else if (callbacks.integration) {
-      this.isUnitTest = false;
-    } else {
+    if (! (callbacks.needs || callbacks.unit) && this.isUnitTest()) {
       Ember.deprecate(
         "the component:" + componentName + " test module is implicitly running in unit test mode, " +
         "which will change to integration test mode by default in an upcoming version of " +
@@ -27,11 +23,6 @@ export default TestModule.extend({
         false,
         { id: 'ember-test-helpers.test-module-for-component.test-type', until: '0.6.0' }
       );
-      this.isUnitTest = true;
-    }
-
-    if (!this.isUnitTest) {
-      callbacks.integration = true;
     }
 
     if (description) {
@@ -40,14 +31,23 @@ export default TestModule.extend({
       this._super.call(this, 'component:' + componentName, callbacks);
     }
 
-    if (this.isUnitTest) {
-      this.setupSteps.push(this.setupComponentUnitTest);
-    } else {
-      this.callbacks.subject = function() {
-        throw new Error("component integration tests do not support `subject()`.");
-      };
-      this.setupSteps.push(this.setupComponentIntegrationTest);
-      this.teardownSteps.unshift(this.teardownComponent);
+    switch (this.testType) {
+      case UNIT_TEST:
+        this.setupSteps.push(this.setupComponentUnitTest);
+        break;
+      case LEGACY_INTEGRATION_TEST:
+        this.setupSteps.push(this.setupLegacyComponentTest);
+        this.setupSteps.push(this.setupComponentLegacyIntegrationTest);
+        break;
+      case INTEGRATION_TEST:
+        this.callbacks.subject = function() {
+          throw new Error("component integration tests do not support `subject()`.");
+        };
+        this.setupSteps.push(this.setupComponentIntegrationTest);
+        this.teardownSteps.unshift(this.teardownComponent);
+        break;
+      default:
+        throw new Error("Unexted test type `" + this.testType + "`");
     }
 
     if (Ember.View && Ember.View.views) {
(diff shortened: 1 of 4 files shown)


hjdivad:
@rwjblue here's a PR for that `integration: 'legacy'` flag we talked about.  I'm still not 100% on the internals of `ember-test-helpers` so please let me know if there's stuff in this PR that looks wonky that I should examine more closely.

Thanks.

hjdivad:
Now the tests even pass!

As an aside the test failures indicate that this feature will not work with ember 2.0, which seems fine to me.

thoughts?

ef4:
What was the semantic change for the integration flag?

hjdivad:
@ef4 There are two things that impacted tests I had written.

1. `render` now renders into an outer context rather than rendering the component directly.  Among other things, this means that `render` requires a template.  This isn't really that big of a deal as it's easy to make this change to tests.

2. `subject` is now disallowed.  This means primarily that there is no sensible way of getting a reference to the component. That makes a lot of sense with the current approach, in which interaction with the component in a component test is done primarily via the DOM.  But previously, where `integration: true` more or less just meant "you don't have to write `needs` entries for everything" it would often make sense to set the flag when writing component unit tests in which you *would* make assertions about component properties.

ef4:
But those semantics are literally the original definition of `integration:true`. There was no earlier `integration:true` mode with different semantics.

Prior to the introduction of that flag, all tests were unit tests, and they used `render` and `subject` as you describe.

hjdivad:
@ef4 whatever the intended semantics, on version 0.4.3 with ember 1.12, you could use `subject` and `render` as I describe with `integration: true` as a means of writing unit tests without needing to specify `needs`.

ie on version 0.4.3 you could write tests like

```js
moduleForComponent('my-component', {
  integration: true, // no need to specify `needs` now
});

test("my-component has this flag set after you click a thing", function () {
  const component = this.subject();

  this.render();

  Ember.run(function () {
    component.$('button').click();
  });

  equal(component.get('flagB'), true, 'flagB is now set');
});
```

rwjblue:
@ef4 - `integration: true` for `moduleForComponent` definitely changed meaning when the current (`this.render` with a template string) style was added.  Prior to those changes, `integration: true` simply gave you a container with no resolver restrictions (so you didn't have to specify `needs`).

rwjblue:
Specifically, the chain of events (as far as I can track down) is:

* https://github.com/switchfly/ember-test-helpers/pull/21 added `TestModuleForIntegration`, which was not exposed by `ember-qunit` (because we were bikeshedding about the exact final user-land API).
* https://github.com/switchfly/ember-test-helpers/pull/24 moves `integration: true` to the top level `TestModule`
* ember-test-helpers@0.3.6 and was released in [ember-qunit@0.2.12](https://github.com/rwjblue/ember-qunit/tree/v0.2.12)
* https://github.com/switchfly/ember-test-helpers/pull/38 removed `TestModuleForIntegration` and moved that logic into `TestModuleForComponent`
* ember-test-helpers@0.4.4 was released in [ember-qunit@0.3.2](https://github.com/rwjblue/ember-qunit/blob/v0.3.2)

So from 0.3.6 through 0.4.3 you could use `integration: true` as a way to avoid specifying `needs`.  

---

*tldr;* I really mucked things up :disappointed:

lib/ember-test-helpers/test-module-for-component.js
```diff
@@ -190,6 +190,70 @@ export default TestModule.extend({
...
+      registry.register('component-lookup:main', Ember.ComponentLookup);
+      this.registry = registry;
+      this.container = registry.container();
+    } else {
+      this.container = Ember.Application.buildContainer(namespace);
+      this.container.register('component-lookup:main', Ember.ComponentLookup);
+    }
+  },
+
+  setupLegacyComponentTest: function () {
```
rwjblue: How does this differ from https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module-for-component.js#L72-L120?  It seems that these are essentially duplicated...

rwjblue: https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module.js#L129-L135 determines if the container that is being used is isolated or has the full resolver (note how we clobber the underlying `resolver` [here](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module.js#L284-L288) after manually resolving all of the `needs` specified items).

rwjblue: Sorry for beating around the bush so much here.  I think that legacy mode should do the following:

* Change the conditional https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module.js#L129-L135 to just use `_setupIntegratedContainer` for legacy mode.
* Use `setupComponentUnitTest` when in legacy mode.

The combination of those two changes, essentially sets us back to the behavior of ember-test-helpers 0.3.6 - 0.4.3.

rwjblue:
@hjdivad -- Let me know if you have any questions about my last few comments here...

rwjblue:
Closing for now, happy to reopen if you want to pick this back up.

hjdivad:
@rwjblue hey sorry it took so long to find time to try this out again.  I've redone the branch with the changes you mentioned and our test suite passes.  The only other change I had to make was this one: https://github.com/hjdivad/ember-test-helpers/blob/integration-legacy/lib/ember-test-helpers/test-module-for-component.js#L202-L204  to prevent `subject` from getting indirectly stomped.

Should be good to reopen or I can open a new PR if you'd prefer.

rwjblue:
Weird, apparently I can only reopen if there hasn't been a force push to the branch the closed PR was submitted from (which seems kinda dumb/weird).

rwjblue:
Submitted https://github.com/switchfly/ember-test-helpers/pull/121 to replace

hjdivad:
@rwjblue that is a strange restriction.

rwjblue:
@hjdivad hmm, why?  `intetration: true` has not changed meaning in any test format other than `moduleForComponent`, why introduce additional API for all module types as a fix for poor SemVer management in module types where no mistake was made?

rwjblue:
I am perfectly willing to be wrong here (and can revert #122), but I want to make sure I understand why.

rwjblue:
D'OH!! The comment above was about the GH restriction of reopening PR's. :facepalm:

PR #118, opened 2015-11-09
Work with `Ember.getOwner`.

rwjblue:
* Allows usage of ES6 syntax.
* Remove usage of broccoli-static-compiler.
* Use ember-getowner-polyfill's `fake-owner` to build a mock owner (since there is no way create an owner directly).
* Add tests confirming that `Ember.getOwner` can be used on instances created through the container.
* Add ability to call `Ember.getOwner(this)` on that test context directly (potentially paving the way to deprecate `this.register`, `this.lookup`, and friends).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ sudo: false
 language: node_js
 
 install:
-  - npm install -g bower broccoli-cli
-  - npm install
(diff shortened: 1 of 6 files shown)


tim-evans:
:tada: Thanks for this!!

PR #119, opened 2015-11-10
Use newly exposed mixins to create owner.

rwjblue:
This is much less brittle than relying on the polyfill for forward/future support.


dgeb:
Nicely done :clap:

PR #121, opened 2015-11-23
Add support for `integration: 'legacy'`

rwjblue:
This semantics of the `integration` flag have changed in a backwards-incompatible way since v0.4.3.  This corresponds to an ember upgrade (using ember-cli) between v1.12 and v1.13.

In order to ease this upgrade path for ember-cli users, `integration: 'legacy'` will set up a component test with v0.4.3 semantics.  Of course, newer tests would still be encouraged to be written with `integration: true`.

---

Replacement PR for https://github.com/switchfly/ember-test-helpers/pull/115.

/cc @hjdivad

### DIFF
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -30,17 +30,17 @@ export default TestModule.extend({
       this.isUnitTest = true;
     }
 
-    if (!this.isUnitTest) {
-      callbacks.integration = true;
-    }
-
     if (description) {
       this._super.call(this, 'component:' + componentName, description, callbacks);
     } else {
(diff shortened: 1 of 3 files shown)


rwjblue:
This looks good to me, we'll need to add a deprecation when using `isLegacy` at some point, to make it acceptable to remove in the next "major" bump (0.5.0 of ember-qunit).

rwjblue:
I'm going to merge this, and submit a follow up PR to tweak `isLegacy` to only be allowable from `moduleForComponent` style tests (since that is the only place where I handled things incorrectly with `integration: true` changing meaning).

rwjblue:
Follow up PR: https://github.com/switchfly/ember-test-helpers/pull/122

PR #127, opened 2015-12-05
Fix registry resolver for isolated container @2.3+

### LINKED ISSUE #126: Tests fail for latest beta/canary
The tests are failing for the latest beta and canary versions of Ember. As far as I can tell, it's related to a change in the registry in which `registry.resolver` became `registry.resolver.resolve`.

I will work on creating a PR since we need the latest canary for performance enhancements that have recently landed.

nickiaconis:
Fixes #126.

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -296,6 +296,9 @@ export default Klass.extend({
 
     if (this.registry) {
       this.registry.fallback.resolver = function() {};
+      if (hasEmberVersion(2,3)) {
+        this.registry.fallback.resolver.resolve = function() {};
+      }
     } else {
       this.container.resolver = function() {};
     }
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -267,9 +267,14 @@ export default Klass.extend({
     }
   },
 
-  _setupContainer: function() {
+  _setupContainer: function(isolated) {
     var resolver = getResolver();
-    var items = buildRegistry(resolver);
+
+    var items = buildRegistry(!isolated ? resolver : Object.create(resolver, {
+      resolve: {
+        value: function() {}
+      }
+    }));
 
     this.container = items.container;
     this.registry = items.registry;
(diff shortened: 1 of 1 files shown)


rwjblue:
This does seem like a good fix, but it seems bad that a change in Ember broke like this. I'll try to dig into the change in Ember and see who is at fault (Ember or ember-test-helpers).

nickiaconis:
Ember deprecated `registry.resolver` being a method (in favor of `registry.resolver.resolve` being a method), and properly handles cases when the registry is created with a `resolver`method, but only in the constructor. Assignment to the property after the fact is not accounted for. I'm not sure how it would be.

rwjblue:
Awesome, thank you for doing the research on that!

dgeb:
> Assignment to the property after the fact is not accounted for. I'm not sure how it would be.

Yeah, this is the case we decided not to allow and deprecate. Not that we want a deprecation to always happen here - we need a better solution.

lib/ember-test-helpers/test-module.js
```diff
@@ -296,6 +296,9 @@ export default Klass.extend({
 
     if (this.registry) {
       this.registry.fallback.resolver = function() {};
+      if (hasEmberVersion(2,3)) {
+        this.registry.fallback.resolver.resolve = function() {};
+      }
```
dgeb: Might be _slightly_ better as:

```
      if (hasEmberVersion(2,3)) {
        this.registry.fallback.resolver = {
          resolve: function() {};
        };
      } else {
        this.registry.fallback.resolver = function() {};
      }
```

Will consider whether this can be handled even more elegantly.

rwjblue:
@dgeb - I was thinking that we might be able to restructure our usage here, to simply create the registry with the right resolver (instead of clobbering it after instantiation). I believe that would work properly for all versions (and we wouldn't need a conditional), it also makes ember-test-helpers follow the "right" pattern...

dgeb:
> I was thinking that we might be able to restructure our usage here, to simply create the registry with the right resolver (instead of clobbering it after instantiation)

Sounds like the right approach to me @rwjblue :+1:

nickiaconis:
@rwjblue - I've updated the PR to alter the resolver's resolve method before the registry is created. Is this the way we should be doing it? If this looks good to you, I can squash my commits.

nickiaconis:
It seems the npm install failed on Travis. How can I run it again?

rwjblue:
Thank you!

rwjblue:
Released in ember-qunit@0.4.17 and ember-mocha@0.8.8.

PR #129, opened 2015-12-09
Add moduleForAcceptance

MattEddy:
# Problem
Currently the behavior necessary for starting and tearing down Ember apps in tests is located in [blueprints](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/tests/helpers/module-for-acceptance.js). Blueprints expose the user to complexity and create extra boilerplate.

# Solution
Create a `TestModuleForApplication` helper. This:
 1. avoids the use of blueprints
 2. gives the ember team a natural way to pass testing improvements down to users
 3. prevents the user from accessing the `testApp` directly in an acceptance test
 4. brings hooks for writing tests to a single location. This improves documentation and creates a stronger separation of concerns between blueprints and tests

# Design
First we pulled out a  `AbstractTestModule`. This is a place to put shared behavior between `TestModule` and `TestModuleForApplication`. It also helps crystalize what needs to be implemented in order to be a 'TestModule'. Next, we implemented a `TestModuleForApplication`. The name 'TestModuleForApplication' was chosen to extend the naming pattern already defined. An acceptance test needs an application, just as a component needs an integration.  

`TestModuleForApplication` does several things. It sets up and tears down a testApp. It provides four temporal hooks, `beforeSetup`/`afterTeardown`, which allow a user of the module to do things before and after the test app is created, and `setup`/`teardown` which both execute while the app is running.

`TestModuleForApplication` provides options for the user of the module to customize the test app. The `Application` option which allows you to define what the test app will be, and a `config` where you can specify things like the `rootElement`.

### DIFF
--- a/lib/ember-test-helpers.js
+++ b/lib/ember-test-helpers.js
@@ -1,7 +1,8 @@
-import Ember                  from 'ember';
-import TestModule             from 'ember-test-helpers/test-module';
+import Ember from 'ember';
+import TestModule from 'ember-test-helpers/test-module';
+import TestModuleForAcceptance from 'ember-test-helpers/test-module-for-acceptance';
 import TestModuleForComponent from 'ember-test-helpers/test-module-for-component';
-import TestModuleForModel     from 'ember-test-helpers/test-module-for-model';
+import TestModuleForModel from 'ember-test-helpers/test-module-for-model';
 import { getContext, setContext } from 'ember-test-helpers/test-context';
 import { setResolver } from 'ember-test-helpers/test-resolver';
 
@@ -9,6 +10,7 @@ Ember.testing = true;
 
 export {
   TestModule,
+  TestModuleForAcceptance,
   TestModuleForComponent,
   TestModuleForModel,
   getContext,
--- a/lib/ember-test-helpers/abstract-test-module.js
+++ b/lib/ember-test-helpers/abstract-test-module.js
@@ -1,0 +1,135 @@
+import { Klass } from 'klassy';
+import { _setupAJAXHooks, _teardownAJAXHooks } from './wait';
+import { setContext, unsetContext } from './test-context';
+
+import Ember from 'ember';
+
+export default Klass.extend({
+  init(description, options) {
+    this.description = description;
+    this.callbacks = options || {};
+
+    this.initSetupSteps();
+    this.initTeardownSteps();
+  },
+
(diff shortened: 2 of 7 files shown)


mmun:
Hmm, not sure what's up with the test error. Builds finally locally.

MattEddy:
Added explanation to PR message

mmun:
@dgeb, @rwjblue: Do you have any feedback for this PR?

dgeb:
I'm :+1: on moving this functionality to a test module in ember-test-modules and out of the blueprint.

The changes look pretty solid. I'm not crazy about the name `AbstractTestModule`, but understand that you need to keep `TestModule` the same and just step back one level. Anyway, this is just an implementation detail that won't be exposed, so it's fine.

I think we need to offer some lifecycle hooks for application instances. Perhaps we could offer an application instance initializer hook which would give access to the instance and allow for custom registrations / lookups just like a regular instance initializer. We want those customizations done on the instance, not the `Application`, so that we can tear it down and throw it away after every test run.

mmun:
(I'm also not a fan of AbstractTestModule :P)

I'm not exactly sure what you mean by lifecycle hooks. I'd prefer, if possible, to not to add any extra hooks to the test module's `callbacks` that the let you configuration the application. It feels weird to me to have configuration that applies to an entire module of tests but with no way to override per test.

Would implementing something like `this.boot()` be sufficient to get this mergeable?

dgeb:
@mmun My main concern here is that there's access to the `Application` instance but not the `ApplicationInstance` instance (oof). 

In other words, `getContext().application instanceof Ember.Application` not `getContext().application instanceof Ember.ApplicationInstance`. 

Our test suites could run faster if we didn't create instances of both `Application` and `ApplicationInstance` for each test.

I realize this is a change from the current blueprinted behavior, and could be handled in a separate PR / discussion if you and @rwjblue think that would be better and we just want to move this forward. This PR on its own is a good step forward.

mmun:
Would it be beneficial to share the Application across _all_ acceptance test modules?

dgeb:
> Would it be beneficial to share the Application across all acceptance test modules?

Yes, I think so. There should be no unique state stored on the Application, so it should be feasible.

rwjblue:
Thanks @MattEddy and @mmun!  

We will continue to iterate...

Issue #133, opened 2015-12-11
Callback context is inconsistent


trentmwillis:
Currently when callbacks for a test module are [contextualized](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module.js#L263), they get bound to the test module rather than the test context. This causes inconsistent behavior, since `setup`/`teardown` get [invoked with the test context](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module.js#L129).

Practically, this means that if you do something such as:

```javascript
moduleFor('component:foo-bar', {
  integration: true,

  setup() {
    this.set('title', 'first title'); // works as expected
  },

  changeTitle() {
    this.set('title', 'second title'); // doesn't work
  }
});

test('it renders', function(assert) {
  this.changeTitle();
});
```

`changeTitle` won't work properly, instead you have to do:

```javascript
changeTitle() {
  this.context.set('title', 'second title');
}
```

This is inconsistent and not intuitive. Can we change it such that all callbacks are invoked with the test context? This might limit access to some test module properties, but I feel most of those shouldn't be getting accessed anyway.

rwjblue:
Generally, I agree with you that the context should be the same for all of the callbacks.  Unfortunately, this is a breaking change and I am not sure how exactly we would go about deprecating the current behavior.

@dgeb - Thoughts?

mmun:
I also agree. Having consistent context on all the callbacks is important for preserving the QUnit mental model.

trentmwillis:
@rwjblue true, definitely breaking. Could we do something like proxy attributes from the test module onto the context? Would probably get ugly though.

rwjblue:
@trentmwillis - Yes, @mmun and I were chatting this through in slack and I came up with the following snippet, that should work in a backwards compatible way:

```js
buildDeprecatedContext(module, context) {
  let keysForDeprecation = Object.keys(module);
  
  if (keysForDeprecation.length === 0) {
    return; // do not call Object.create if we don't need to
  }

  let wrappedContext = Object.create(context);

  for (let index = 0, length = keysForDeprecation.length; index++; index < length) {
    let key = keys[index];

    Object.defineProperty(wrappedContext, key, {
      get: function() {
        Ember.deprecate('ffffuuuuuu, fix it!');
        return module[key];
      }
    });
  });
},
```

And call that from `contextualizeCallbacks` (or something like that).  

We would need tests for:

[...]

PR #136, opened 2015-12-15
Change callback context to deprecate test module properties

trentmwillis:
Addresses #133.

One note: in the case where a user-defined context property collides with a property on the test module, nothing special is done since the user-defined value isn't added to the context until after binding the callbacks.

### DIFF
--- a/lib/ember-test-helpers/test-module.js
+++ b/lib/ember-test-helpers/test-module.js
@@ -246,13 +246,17 @@ export default Klass.extend({
     this.cachedCalls = this.cachedCalls || {};
 
     var keys = (Object.keys || Ember.keys)(callbacks);
+    var keysLength = keys.length;
 
-    for (var i = 0, l = keys.length; i < l; i++) {
-      this._contextualizeCallback(context, keys[i]);
+    if (keysLength) {
+      var deprecatedContext = this._buildDeprecatedContext(this, context);
+      for (var i = 0; i < keysLength; i++) {
+        this._contextualizeCallback(context, keys[i], deprecatedContext);
+      }
     }
   },
 
-  _contextualizeCallback: function(context, key) {
+  _contextualizeCallback: function(context, key, callbackContext) {
     var _this     = this;
     var callbacks = this.callbacks;
     var factory   = context.factory;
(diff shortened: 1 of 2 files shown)


lib/ember-test-helpers/test-module.js
```diff
@@ -269,6 +271,36 @@ export default Klass.extend({
     };
   },
 
+  /*
+    Builds a version of the passed in context that contains deprecation warnings
+    for accessing properties that exist on the module.
+  */
+  _buildDeprecatedContext: function(module, context) {
+    var deprecatedContext = Object.create(context);
```
rwjblue: I would prefer to avoid the `Object.create(context)` call when there are no `keysForDeprecation`, can you tweak this to be behind a guard?

Something like this slight reordering:

```
var keysForDeprecation = Object.keys(module);

if (keysForDeprecation.length === 0) {
  return this;
}

var deprecatedContext = Object.create(context);
// .... other stuff
```

rwjblue: Wait, sorry, I misread.  This is only being called when needed (because `_contextualizeCallback` is called only when needed), but we can still cache the deprecated context (instead of building a new one for every function that is in `options`).

lib/ember-test-helpers/test-module.js
```diff
@@ -269,6 +271,36 @@ export default Klass.extend({
...
+  },
+
+  /*
+    Defines a key on an object to act as a proxy for deprecating the original.
+  */
+  _proxyDeprecation: function(obj, proxy, key) {
+    if (typeof proxy[key] === 'undefined') {
+      Object.defineProperty(proxy, key, {
+        get: function() {
+          Ember.deprecate('Accessing the test module property "' + key + '" from a callback is deprecated.', false, { id: 'ember-test-helpers.test-module.callback-context', until: '0.6.0' });
```
rwjblue: The `until` value here is tricky.  I agree that 0.6.0 is the correct version number of this project, but none of this projects consumers actually see this version.  For example, ember-mocha is already on 0.8.  

It is likely that nothing can be done here, but I wanted to call it out and see if y'all had any ideas.

/cc @mixonic @dgeb

rwjblue:
This looks good, only one thing to tweak (caching the `deprecatedContext` across multiple callbacks)...

tests/test-module-test.js
```diff
@@ -94,13 +94,43 @@ moduleFor('component:x-foo', 'component:x-foo -- callback context', {
...
+test("can access test context properties from a callback's 'this' and not raise a deprecation", function() {
+  var deprecations = [];
+  Ember.deprecate = function(message) {
+    deprecations.push(message);
+  };
+
+  this.foo = 'bar';
+
+  equal(this.getFoo(), 'bar');
+  ok(!deprecations.length);
```
rwjblue: I am unsure why, but this assertion is failing on ember-1.13.  I am rerunning in Travis in case it is a fluke.

lib/ember-test-helpers/test-module.js
mixonic: Oof. Libraries outside of Ember core using `deprecate` is going to be painful for sure. There is no `project` property, which there probably should be if we really consider them a public API to be used when deprecating other libs.

Tricky b/c `until` is required.

rwjblue: I think the name spacing that we have done (and is done here) helps a bit. It at least says ember-test-helpers so you can know what to look for. Perhaps we should add a project option here, and add it to the API for the debug handlers feature so that various tools have a better chance at this...

rwjblue: Either way, I think this PR is doing the best we can ATM.

mixonic: Agreed.

lib/ember-test-helpers/test-module.js
trentmwillis: Good point, will adjust this to cache the `deprecatedContext`

tests/test-module-test.js
trentmwillis: It's possible that 1.13 is throwing a deprecation unrelated to the changes here. Will investigate this before updating again.

trentmwillis:
@rwjblue updated to cache the deprecated context across callbacks and fixed the failure with 1.13. Turns out `Ember.deprecate` was getting called, but the `test` value was `true`, so it wasn't really throwing a deprecation.

rwjblue:
Thanks @trentmwillis!

PR #138, opened 2016-01-13
Make build-registry work correctly when ember-data addon is used

pangratz:
Credit for fix code goes to @rwjblue, I am merely the humble PR creator.

This addresses emberjs/data#4071.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -98,6 +98,18 @@ module.exports = {
       resolutions: {
         "ember": "canary"
       }
+    },
+    {
+      name: 'ember-data-2.3',
+      dependencies: {
+        "ember": "components/ember#release",
+      },
+      devDependencies: {
+        "ember-data": "~2.3"
+      },
+      resolutions: {
+        "ember": "release"
+      }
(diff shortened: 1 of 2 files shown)


config/ember-try.js
```diff
@@ -98,6 +98,18 @@ module.exports = {
...
+      name: 'ember-data-2.3',
+      dependencies: {
+        "ember": "components/ember#release",
+      },
+      devDependencies: {
+        "ember-data": "~2.3"
+      },
+      resolutions: {
+        "ember": "release"
+      }
```
pangratz: Hmm, I just realized that ember-try only works with bower dependencies for now. So I am not sure this config is really helpful to test the code path below :pensive:

rwjblue: Agreed, but it doesn't hurt...

rwjblue:
Thank you! I'll release shortly...

seawatts:
@rwjblue just curious, do you have an ETA? Also will this have to kick off a release in ember-quint and ember-cli-qunit as well?

PR #145, opened 2016-01-27
Update ember-try

pangratz:
By using the latest version of ember-try, we can change dependencies of
npm packages too. This is useful for the ember-data-2.3 scenario,
because ember-data 2.3 is an addon, which needs to be installed via npm
instead of bower.

---

**TODO**

- [x] wait for kategengler/ember-try#45 to be merged and released in new version
- [x] update `ember-data-2.3` scenario so it uses `ember-data` npm package

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -101,14 +101,28 @@ module.exports = {
     },
     {
       name: 'ember-data-2.3',
-      dependencies: {
-        "ember": "components/ember#release",
+      bower: {
+        dependencies: {
+          "ember": "components/ember#release",
+        },
+        devDependencies: {
+          // Note that the bower dependency for ember-data is set here to ~2.3,
+          // altough since 2.3 ember-data is a proper addon which is installed
+          // via npm; we are setting the version for the bower dependency here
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome, thanks for working on this!

config/ember-try.js
```diff
@@ -101,14 +101,28 @@ module.exports = {
...
+      bower: {
+        dependencies: {
+          "ember": "components/ember#release",
+        },
+        devDependencies: {
+          // Note that the bower dependency for ember-data is set here to ~2.3,
+          // altough since 2.3 ember-data is a proper addon which is installed
+          // via npm; we are setting the version for the bower dependency here
+          // so we make sure that ember-data 2.3 is used. This can be removed
+          // once there is support for removing dependencies in an ember-try
```
rwjblue: Let's remove ember-data from the default `bower.json` and add it to each scenario.  That way we do not need ember-try to support removing...

rwjblue:
Closing in favor of https://github.com/switchfly/ember-test-helpers/pull/148.

PR #150, opened 2016-04-12
Prefer Ember.assign over Ember.merge

martndemus:
See: http://emberjs.com/blog/2016/04/11/ember-2-5-released.html#toc_code-ember-assign-code

PR #153, opened 2016-04-29
Use new path to loader.js

danjamin:
In regards to #150, this should fix the path issue.  However, there seems to be subsequent ember-data related issues.  @rwjblue I bower installed (without saving) ember-data and that addressed those issues only to be left with some `assert.async` related issues.  I figured I'd PR this change and see what happens.  Granted, the tests still won't pass.


rwjblue:
Thanks!  This is a great start!

PR #155, opened 2016-05-25
Ensure prior render is cleared when calling `this.render`.

### LINKED ISSUE #111: How should mutliple render calls in one test behave?
I just generated a component integration test for one of my components which has a hardcoded elementId (F*ck me, right?) and the test exploded into my face with
```Assertion Failed: Attempted to register a view with an id already in use: myComponentName```
because of the two render calls (block and non-block form).

@rwjblue on Slack:
> Should we allow multiple `this.render` calls in one test, and if we do should we teardown the prior one before doing another?
> the answers do not seem obvious to me
> I guess I would lean towards cleaning up if a `this.render` was already run....
> if you care and have time, i'd love an issue in ember-test-helpers for this
> it seems somewhat edge-casey but seems valuable to get it right

rwjblue:
Previously, calling `this.render` twice within the same test (without calling `this.clearRender()` in between) would cause any components created in the first render to never be destroyed (and therefore they cannot cleanup any listeners they created).

This change essentially ensures that we call `clearRender` before each `this.render` invocation so that if a component had been previously instantiated it is cleared and reset.

Fixes #111.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -89,6 +89,7 @@ module.exports = {
     },
     {
       name: 'ember-canary',
+      allowedToFail: true,
       dependencies: {
         "ember": "components/ember#canary"
       },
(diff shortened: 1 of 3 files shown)


rwjblue:
Failure is due to issues on canary. I created https://github.com/emberjs/ember.js/issues/13557 to track that upstream

PR #156, opened 2016-05-30
RFC: add waitForTimersAndIntervals

kellyselden:
After reading https://github.com/emberjs/ember.js/issues/3008, I think I understand it as `Ember.run.later` blocks and `setTimeout` doesn't block `wait()`. My scenario is a library I'm using is using `setTimeout` ([here](https://github.com/jhurliman/node-rate-limiter)), yet I still want my `wait()` call to block.
I whipped this code up quick to test it out and it works as expected. The implementation is probably terrible. I just wanted to start a conversation.

### DIFF
--- a/lib/ember-test-helpers/wait.js
+++ b/lib/ember-test-helpers/wait.js
@@ -55,9 +55,56 @@ export default function wait(_options) {
   var waitForTimers = options.hasOwnProperty('waitForTimers') ? options.waitForTimers : true;
   var waitForAJAX = options.hasOwnProperty('waitForAJAX') ? options.waitForAJAX : true;
   var waitForWaiters = options.hasOwnProperty('waitForWaiters') ? options.waitForWaiters : true;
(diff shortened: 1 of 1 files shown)


kellyselden:
BTW, I'm also tackling this problem from a different angle here https://github.com/machty/ember-concurrency/issues/55

Turbo87:
I'm a little worried about overwriting the native functions like this. @kellyselden is this still a relevant to you?

kellyselden:
@Turbo87 It's still being successfully used [here](https://github.com/kellyselden/package-hint-historic-resolver/blob/feb8d15ae573cbdb3718a3b1585dfdec05972e05/tests/integration/components/dependency-component-test.js#L10). I'm not sure if there has been a better way discovered of how to do this.

rwjblue:
I'm pretty uncomfortable with monkey patching `setTimeout` & co. I think an addon can easily do this and leverage the new `ember-test-waiters` to provide the waiting experience.

Going to close this in favor of that path forward....

PR #157, opened 2016-06-04
Remove usage of `context` in `moduleForComponent`.

rwjblue:
Setting `context` or `controller` on an `Ember.Component` was absolutely not supposed to work (ever), but this library worked around that to serve its own needs when `Ember.View` was deprecated and removed.

This PR makes `moduleForComponent` work properly against canary (soon to be 2.7.0), by avoiding this archaic concept and simply tracking the things `set` on the test context manually.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -64,40 +64,126 @@ module.exports = {
       }
     },
     {
-      name: 'ember-release',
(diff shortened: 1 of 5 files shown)


rwjblue:
Closing in favor of #158

PR #158, opened 2016-06-05
Refactor `moduleForComponent` rendering to use outlet system.

rwjblue:
Setting context or controller on an Ember.Component was absolutely not supposed to work (ever), but this library worked around that to serve its own needs when Ember.View was deprecated and removed.

This PR makes `moduleForComponent` work properly against canary (soon to be 2.7.0), by avoiding this archaic concept and using normal outlet system.

### DIFF
--- a/lib/ember-test-helpers/-legacy-overrides.js
+++ b/lib/ember-test-helpers/-legacy-overrides.js
@@ -1,0 +1,90 @@
+import Ember from 'ember';
+
+import hasEmberVersion from './has-ember-version';
+
+export function preoutletRefactorSetupIntegrationForComponent() {
+  var module = this;
+  var context = this.context;
+
+  this.actionHooks = {};
+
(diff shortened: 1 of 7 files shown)


lib/ember-test-helpers/-legacy-overrides.js
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    if (hasEmberVersion(2,0)) {
+      return ret;
+    }
+  };
+
+  context.setProperties = function(hash) {
+    var ret = Ember.run(function() {
+      return Ember.setProperties(context, hash);
```
krisselden: this method has a return value?

rwjblue: Only in Ember 2.0+, it is intended to emulate what `Ember.setProperties` does: https://github.com/emberjs/ember.js/blob/dc4b60be5849a6dad2cc464822294a435e08e31c/packages/ember-metal/lib/set_properties.js#L37

PR #164, opened 2016-06-26
Improved error message

seansellek:
The error currently thrown when an integration test receives an unexpected action does not provide any guidance as to how to resolve the error. Googling also yields no useful results. This error might commonly occur when 1) an action wasn't meant for the context and should have been sent to the component, or 2) an action was supposed to bubble up but was not stubbed.

Expanded the error message to include guidance for these two scenarios.


rwjblue:
Can you wrap the line at around 80 characters, and give an example of what the output looks like?

rwjblue:
@seansellek - Ping?

Turbo87:
closing due to inactivity

PR #166, opened 2016-07-28
Updating Package Versions within current range

elwayman02:
All new package versions are within their currently specified ranges, just updated to the latest release in that range.


elwayman02:
@rwjblue up to you if you want to merge...master build will also fail anyway, we need to fix issues with beta/canary Ember.

PR #172, opened 2016-08-02
Enable usage when jQuery is not present.

rwjblue:


rwjblue:
Exposing the element for now as `this._element`. Keeping it private for now allows us to change it later if needed...

Issue #173, opened 2016-08-11
Ability to set Resolver per-module


trentmwillis:
In the current world, it is easy to set a Resolver to be used for an entire test-suite by using the `setResolver` method. This is fine in the vast majority of use cases, but with the advent of Engines, and particularly in-repo-engines, we need some additional flexibility [as shown in this issue](https://github.com/dgeb/ember-engines/issues/123).

One approach that I suggest is to allow setting a `resolver` option that is passed into the test module like so:

```js
import EngineResolver from 'some-engine/resolver';

moduleForComponent('awesome-component', {
  resolver: EngineResolver.create()
});
```

Then, when constructing the container/registry for that test module, we'll use the passed in Resolver instead of the globally defined one that was set with `setResolver`.

This provides backwards compatibility while giving enough flexibility that the ember-engines addon could provide an additional wrapped to allow getting a proper resolver by engine-name. I imagine something like:

```js
moduleForComponent('awesome-component', {
  resolver: getEngineResolver('some-engine')
});
```

rwjblue:
This seems good to me.

dgeb:
Works for me as well 👍

Turbo87:
resolved by #174

rwjblue:
Thank you @turbo87

PR #174, opened 2016-08-11
Introduce resolver option to test module

trentmwillis:
Addresses #173. After implementing, I realize this could potentially break some one that has already been using the `resolver` property in their passed in environment, but I think that risk is acceptable.

### DIFF
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -1,6 +1,5 @@
 import TestModule from './test-module';
 import Ember from 'ember';
-import { getResolver } from './test-resolver';
 import hasEmberVersion from './has-ember-version';
 import { preGlimmerSetupIntegrationForComponent } from './-legacy-overrides';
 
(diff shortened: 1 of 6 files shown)


rwjblue:
Seems like this should be changed in `test-module.js` (`moduleFor` also supports `integration: true` so it needs/uses a resolver) around [here](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module.js#L240).  Once that is changed, I'm not sure that the `test-module-for-integration.js` or `test-module-for-component` need to be changed as much...

trentmwillis:
@rwjblue `test-module.js` is changed. `test-module-for-component` inherits from `test-module`, so that is fine. Then `test-module-for-integration` inherits directly from `abstract-test-module`, so it also needs the changes.

Also, looks like Canary and Alpha are failing, for ambiguous reasons.

rwjblue:
Yeah, I keep fighting with canary + alpha on CI.  I'll look at that separately.

trentmwillis:
Seems like we could just make Canary an allowed failure. Also, could we drop some of the versions we test against and split the tests up into different processes?

rwjblue:
Looks like canary was passing in https://github.com/switchfly/ember-test-helpers/pull/175, maybe try rebasing?

trentmwillis:
@rwjblue rebased and all is green now

rwjblue:
Released in v0.5.33.

PR #175, opened 2016-08-16
Allow test frameworks to set context to use.

rwjblue:
Supports efforts in https://github.com/rwjblue/ember-qunit/pull/237.

### DIFF
--- a/lib/ember-test-helpers/abstract-test-module.js
+++ b/lib/ember-test-helpers/abstract-test-module.js
@@ -1,13 +1,17 @@
 import { Klass } from 'klassy';
 import { _setupAJAXHooks, _teardownAJAXHooks } from './wait';
-import { setContext, unsetContext } from './test-context';
+import { getContext, setContext, unsetContext } from './test-context';
 
 import Ember from 'ember';
 
-const assign = Ember.assign || Ember.merge;
+// calling this `merge` here because we cannot
+// actually assume it is like `Object.assign`
+// with > 2 args
+const merge = Ember.assign || Ember.merge;
 
 export default Klass.extend({
   init(name, options) {
+    this.context = undefined;
     this.name = name;
     this.callbacks = options || {};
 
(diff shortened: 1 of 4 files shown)


lib/ember-test-helpers/abstract-test-module.js
```diff
@@ -100,12 +104,26 @@ export default Klass.extend({
...
+
+    merge(context, {
       dispatcher: null,
       inject: {}
-    }, options);
+    });
+    merge(context, options);
+
+    setContext(context);
+    this.context = context;
```
Turbo87: this line seems unnecessary given the line before

rwjblue: The line before doesn't set `this.context`

lib/ember-test-helpers/abstract-test-module.js
```diff
@@ -100,12 +104,26 @@ export default Klass.extend({
...
+
+  setContext(context) {
+    this.context = context;
+  },
+
+  getContext() {
+    if (this.context) { return this.context; }
 
-    setContext(config);
+    return this.context = getContext() || {};
```
Turbo87: can we make this:

```js
if (!this.context) {
  this.context = getContext() || {};
}
return this.context;
```

lib/ember-test-helpers/abstract-test-module.js
Turbo87: ah crap, you're right... missed that the `this` was missing... this is not Java... 😉

lib/ember-test-helpers/abstract-test-module.js
rwjblue: Sure, seems functionally the same.  Out of curiosity what makes you prefer that pattern over what I did originally?

Turbo87: it doesn't have the assignment in the return statement. you could make it two statements, but I guess the other proposal is shorter

Issue #178, opened 2016-09-16
Reset test elements instead of emptying them


trentmwillis:
In our test suite we have a need to set some content in `#ember-testing` before our tests are run. This works fine for the first test, but after any test that uses `moduleFor`, that content is removed as part of [`teardownTestElements`](https://github.com/switchfly/ember-test-helpers/blob/128589b2f9ab185b2b6db65810e8406449defdf9/lib/ember-test-helpers/abstract-test-module.js#L138).

Instead of completely emptying the `#ember-testing` element, I think it would be better to cache the initial state of the element in `setupTestElements` and restore it in `teardownTestElements`.

cc @sangm

rwjblue:
Hmm.  I'm not sure of the right path forward here.  `#ember-testing` is intended to be functionally similar to `#qunit-fixture`, where the actual element is completely controlled by the testing harness/framework.

I suspect using `QUnit.testStart` to setup your desired `ember-testing` element contents would be a good work around for now (maybe that is good enough? I dunno).

rwjblue:
Welp: https://github.com/jquery/qunit/blob/e2b38da0be2ab4dc6fe548910fe5f71c55321f95/runner/fixture.js#L13-L25

rwjblue:
I think this is good then, following along like the link above should work well.

sangm:
So, the issue at hand is that in testing, rootElement is set to `ember-testing`, and putting divs inside anything else won't let the events hash trigger...

rwjblue:
Closed by #180

PR #180, opened 2016-09-19
Reset ember-testing div to initial state on teardown

trentmwillis:
Addresses #178

### DIFF
--- a/lib/ember-test-helpers/abstract-test-module.js
+++ b/lib/ember-test-helpers/abstract-test-module.js
@@ -95,11 +95,15 @@ export default Klass.extend({
   },
 
   setupTestElements() {
-    if (!document.querySelector('#ember-testing')) {
+    let testEl = document.querySelector('#ember-testing');
+    if (!testEl) {
       let element = document.createElement('div');
       element.setAttribute('id', 'ember-testing');
 
       document.body.appendChild(element);
+      this.fixtureResetValue = '';
+    } else {
+      this.fixtureResetValue = testEl.innerHTML;
     }
   },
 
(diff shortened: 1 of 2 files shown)


lib/ember-test-helpers/abstract-test-module.js
```diff
@@ -95,11 +95,15 @@ export default Klass.extend({
...
-    if (!document.querySelector('#ember-testing')) {
+    let testEl = document.querySelector('#ember-testing');
+    if (!testEl) {
       let element = document.createElement('div');
       element.setAttribute('id', 'ember-testing');
 
       document.body.appendChild(element);
+      this.fixture = '';
+    } else {
+      this.fixture = testEl.innerHTML;
```
rwjblue: Can you change this to `this.fixtureResetValue`, and initialize it in the constructor?

rwjblue:
@trentmwillis - ping?

lib/ember-test-helpers/abstract-test-module.js
trentmwillis: I can change the name. For initialization, I originally tried it in the constructor, but [`test-module` overrides `init` and doesn't call `_super`](https://github.com/switchfly/ember-test-helpers/blob/77f9a53da9d8c19a85b3122788caadbcc59274c2/lib/ember-test-helpers/test-module.js#L8-L44) so it never gets set in that scenario. I could add it in both locations, but I figured this was cleaner. Does cause the shape to morph, but figured since this was just a testing environment that it's not a huge deal.

trentmwillis:
Sorry @rwjblue got pulled into a bunch of Engines work, will try to wrap this up today.

trentmwillis:
@rwjblue changed the property name and added another test (was a pain to figure out how to test it).

sangm:
@rwjblue any updates on this?

PR #181, opened 2016-11-02
Overriding toString() to return the test subject name

kiwiupover:
A fix for: https://github.com/emberjs/ember-qunit/issues/247

In an integration test where a component is instantiated the test is the outside context.  Ember when makes an assertion at that point the test is the object that ember is using in the assertion object hench `test.toString()` returns `[object Object]` not a helpful error.

After this PR for this test: 
```js
moduleFor('component:x-foo', 'component:x-foo -- callback context', { ...

this.toString() >> 'test for:  component:x-foo'
```

```js
moduleForAcceptance('TestModuleForAcceptance | Basic acceptance tests', { ...
 
this.toString() >> 'test for: TestModuleForAcceptance | Basic acceptance tests'
```

```js
moduleForIntegration('Component Integration Tests', { ...
 
this.toString() >> 'test for: Component Integration Tests'
```

```js 
moduleForComponent('pretty-color', 'component:pretty-color -- this.render in setup', { ...

this.toString() >> 'test for: component:pretty-color'
```

```js 
moduleForModel('kiwis', 'model test', { ...

this.toString() >> 'test for: model:kiwis'
```

### DIFF
--- a/lib/ember-test-helpers/abstract-test-module.js
+++ b/lib/ember-test-helpers/abstract-test-module.js
@@ -132,15 +132,15 @@ export default Klass.extend({
   },
 
   setToString() {
-    this.context.toString = function() {
+    this.context.toString = () =>  {
       if(this.subjectName) {
         return `test for: ${this.subjectName}`;
       }
-      
+
       if(this.name) {
         return `test for: ${this.name}`;
       }
-    }.bind(this);
+    };
   },
 
   setupAJAXListeners() {
--- a/lib/ember-test-helpers/abstract-test-module.js
+++ b/lib/ember-test-helpers/abstract-test-module.js
@@ -134,11 +134,11 @@ export default Klass.extend({
   setToString() {
     this.context.toString = () =>  {
       if(this.subjectName) {
-        return `test for: ${this.subjectName}`;
+        return `test context for: ${this.subjectName}`;
       }
 
       if(this.name) {
-        return `test for: ${this.name}`;
+        return `test context for: ${this.name}`;
       }
     };
   },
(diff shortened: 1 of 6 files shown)


lib/ember-test-helpers/abstract-test-module.js
```diff
@@ -130,6 +131,18 @@ export default Klass.extend({
...
+  setToString() {
+    this.context.toString = function() {
+      if(this.subjectName) {
+        return `test for: ${this.subjectName}`;
+      }
+      
+      if(this.name) {
+        return `test for: ${this.name}`;
+      }
+    }.bind(this);
```
rwjblue: I think we can avoid the `.bind(this)` by using an arrow function

kiwiupover: yes yes

kiwiupover:
The tests are looking sad in travis I will investigate.

kiwiupover:
@rwjblue this is ready again. 

The only questions I have should we use `test for:` in the `moduleForAcceptance` and the `moduleForIntegration` test in those cases we get `test for: TestModuleForAcceptance | Basic acceptance tests` and `test for: Component Integration Tests` thoughts?

kiwiupover:
@trentmwillis in regards to https://github.com/emberjs/ember-qunit/issues/247

rwjblue:
> The only questions I have should we use test for: in the moduleForAcceptance and the moduleForIntegration test in those cases we get test for: TestModuleForAcceptance | Basic acceptance tests and test for: Component Integration Tests thoughts?

Ya, seems ok to me.  We might also tweak it to be `test context for:`, what do you think?

kiwiupover:
sounds good to me @rwjblue

kiwiupover:
@rwjblue done!

kiwiupover:
Ping @rwjblue I think this is is ready.

rwjblue:
Thanks @kiwiupover!

PR #183, opened 2016-11-22
Replace "klassy" with ES6 classes

Turbo87:
/cc @trentmwillis @rwjblue @dgeb @stefanpenner

### DIFF
--- a/lib/ember-test-helpers/test-module-for-acceptance.js
+++ b/lib/ember-test-helpers/test-module-for-acceptance.js
@@ -2,18 +2,18 @@ import AbstractTestModule from './abstract-test-module';
 import Ember from 'ember';
 import { getContext } from './test-context';
 
-export default AbstractTestModule.extend({
+export default class extends AbstractTestModule {
   setupContext() {
-    this._super({ application: this.createApplication() });
-  },
+    super({ application: this.createApplication() });
+  }
 
   teardownContext() {
     Ember.run(() => {
       getContext().application.destroy();
     });
 
-    this._super();
-  },
+    super();
+  }
 
   createApplication() {
     let { Application, config } = this.callbacks;
@@ -27,4 +27,4 @@ export default AbstractTestModule.extend({
 
     return application;
   }
-});
+}
--- a/lib/ember-test-helpers/test-module-for-acceptance.js
+++ b/lib/ember-test-helpers/test-module-for-acceptance.js
@@ -4,7 +4,7 @@ import { getContext } from './test-context';
 
 export default class extends AbstractTestModule {
   setupContext() {
-    super({ application: this.createApplication() });
+    super.setupContext({ application: this.createApplication() });
   }
 
   teardownContext() {
@@ -12,7 +12,7 @@ export default class extends AbstractTestModule {
       getContext().application.destroy();
     });
 
-    super();
+    super.teardownContext();
   }
 
   createApplication() {
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -44,11 +44,7 @@ export default TestModule.extend({
       this.isUnitTest = true;
     }
 
-    if (description) {
-      this._super('component:' + componentName, description, callbacks);
-    } else {
-      this._super('component:' + componentName, callbacks);
-    }
+    this._super('component:' + componentName, description, callbacks);
 
     if (!this.isUnitTest && !this.isLegacy) {
       callbacks.integration = true;
--- a/lib/ember-test-helpers/test-module-for-component.js
+++ b/lib/ember-test-helpers/test-module-for-component.js
@@ -24,11 +24,13 @@ export default TestModule.extend({
       callbacks = {};
     }
 
+    let integrationOption = callbacks.integration;
+
     this.componentName = componentName;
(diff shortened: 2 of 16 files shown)


rwjblue:
Will take some time to do a more thorough review, but I definitely approve the effort (and from prior conversations with @dgeb I believe he does as well).

Turbo87:
💚  ... thanks Travis :)

dgeb:
Yeah, Klassy is a lib I will gladly retire. Thanks @Turbo87.

lib/ember-test-helpers/test-module-for-acceptance.js
```diff
@@ -2,18 +2,18 @@ import AbstractTestModule from './abstract-test-module';
 import Ember from 'ember';
 import { getContext } from './test-context';
 
-export default AbstractTestModule.extend({
+export default class extends AbstractTestModule {
   setupContext() {
-    this._super({ application: this.createApplication() });
-  },
+    super({ application: this.createApplication() });
```
rwjblue: This should be `super.setupContext` (bare `super` is only valid in `constructor`).

lib/ember-test-helpers/test-module-for-acceptance.js
```diff
@@ -2,18 +2,18 @@ import AbstractTestModule from './abstract-test-module';
...
+  }
 
   teardownContext() {
     Ember.run(() => {
       getContext().application.destroy();
     });
 
-    this._super();
-  },
+    super();
```
rwjblue: Should be `super.teardownContext`

lib/ember-test-helpers/test-module-for-component.js
```diff
@@ -133,18 +136,18 @@ export default TestModule.extend({
...
-      return setupComponentIntegrationTest;
+      return setupComponentIntegrationTest.apply(this, arguments);
     }
-  })(),
+  }
 
-  setupContext: function() {
-    this._super.call(this);
+  setupContext() {
+    super();
```
rwjblue: bare `super` is only valid in `constructor`

lib/ember-test-helpers/test-module-for-integration.js
```diff
@@ -93,7 +95,7 @@ export default AbstractTestModule.extend({
       return container.lookupFactory(subjectName);
     };
 
-    this._super({
+    super({
```
rwjblue: bare `super` is only valid in `constructor`

lib/ember-test-helpers/test-module.js
```diff
@@ -94,29 +90,29 @@ export default AbstractTestModule.extend({
...
+  setupContext() {
     var subjectName = this.subjectName;
     var container = this.container;
 
     var factory = function() {
       return container.lookupFactory(subjectName);
     };
 
-    this._super({
+    super({
```
rwjblue: bare `super` is only valid in `constructor`

Turbo87: ouch, forgot about that... thanks!

Turbo87:
@rwjblue done!

Turbo87:
@rwjblue I do agree that those need updates to remove `klassy`, but I don't see why they would error? The way I see it we would just include an unnecessary dependency if we didn't update them, but everything should still be running the same way as before.

Turbo87:
also I tend to say that `klassy` is an implementation detail and as such private API so no need for a major version bump, but we obviously will have to test whether the new `ember-test-helpers` version would run smoothly with older testing libs like `ember-cli-qunit`.

rwjblue:
> I don't see why they would error?

When this PR lands the `klassy` dep will no longer be present, those calls to `resolve.sync('klassy', { baseDir: emberTestHelpersPath })` will begin to error out.

rwjblue:
> I tend to say that klassy is an implementation detail and as such private API

I agree in general. 

> no need for a major version bump

Disagree, for the reason mentioned above. I believe that existing versions of ember-qunit / ember-cli-qunit will error after a `rm -rf node_modules && npm install`.

Turbo87:
> I believe that existing versions of ember-qunit / ember-cli-qunit will error after a rm -rf node_modules && npm install.

I just tried it on a fresh app with `npm link ember-test-helpers` and both testing libs and didn't see any errors

rwjblue:
I don't this npm link is good enough. You would need to ensure there is no version of `klassy` present in `node_modules` (which would not be the case if you installed normally then `npm link`ed).

Turbo87:
> You would need to ensure there is no version of klassy present in node_modules (which would not be the case if you installed normally then npm linked).

you're right, I'm stupid... I see how this would break now. Since `klassy` was a dependency of `ember-test-helpers` and not of `ember-cli-qunit` the resolver call would fail. Since we're sub-1.0 this will only require a minor version bump though to comply with semver.

rwjblue:
> Since we're sub-1.0 this will only require a minor version bump though to comply with semver.

Ya. I was thinking 0.6.0 (I sometimes incorrectly refer to this as a "major" bump when sub 1.0, but what I'm really meaning is that we need to *not* match for users with `^0.5.x`).

Turbo87:
🎉 

will adjust the other libs to account for this change later today unless someone else beats me to it 😉

Issue #195, opened 2017-02-05
Deprecation Warning thrown Even when `needs:[]` is passed


blimmer:
I first noticed this and reported it upstream with ember-mocha (https://github.com/emberjs/ember-mocha/issues/129), but I think it's actually an issue with a change made to test-module-for-component.

This issue doesn't show with current ember-qunit and ember-test-helpers 
```
├─┬ ember-cli-qunit@2.2.5
│ └─┬ ember-qunit@0.4.22
│   └── ember-test-helpers@0.5.34
```

but between 0.5.34 and 0.6.x, a call to `super` was added before checking the options passed:

https://github.com/emberjs/ember-test-helpers/blob/d5c50d0058e141aff0b3b7ed72932e6eca6e6ece/lib/ember-test-helpers/test-module-for-component.js#L29

This call mutates the options hash and deletes the `needs` key:

https://github.com/emberjs/ember-test-helpers/blob/d5c50d0058e141aff0b3b7ed72932e6eca6e6ece/lib/ember-test-helpers/test-module.js#L50

Which then causes the deprecation to fire:

https://github.com/emberjs/ember-test-helpers/blob/d5c50d0058e141aff0b3b7ed72932e6eca6e6ece/lib/ember-test-helpers/test-module-for-component.js#L38-L45

I have a proof of concept exposing this error here:

https://github.com/blimmer/ember-cli-example-app-for-github/compare/ember-mocha/issue-129

I'm not sure what the right fix is, maybe moving the call to super below the checks?

rwjblue:
Thanks for the detailed write up, will try to poke at this tomorrow...

Turbo87:
@blimmer the issue with calling `super()` later is that you have no access to `this` before the `super()` call happened and that's also the reason why I had to move the call.

PR #196, opened 2017-02-13
Deprecation Warning thrown with `needs:[]`

canufeel:
- Fix `moduleForComponent` unit test deprecation being raised even if `needs` is provided.
- Add a test case.

Fixes: #195

### DIFF
--- a/tests/test-module-for-component-test.js
+++ b/tests/test-module-for-component-test.js
@@ -9,6 +9,8 @@ var $ = Ember.$;
 
 var Service = Ember.Service || Ember.Object;
 
+const TEST_TYPE_DEPRECATION_ID = 'ember-test-helpers.test-module-for-component.test-type';
+
 function moduleForComponent(name, description, callbacks) {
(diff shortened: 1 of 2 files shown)


tests/test-module-for-component-test.js
```diff
@@ -298,6 +298,30 @@ test('it allows missing callbacks', function() {
...
+    deprecations = [];
+  }
+});
+
+test('deprecation is not raised', function() {
+  let deprecationId = 'ember-test-helpers.test-module-for-component.test-type';
+  setupRegistry();
+  testModule = new TestModuleForComponent('pretty-color', { needs: ['x:foo'] });
+  notOk(deprecations.find(function(deprecation) {
+    return deprecation.length === 3 && deprecation[2].id === deprecationId;
```
Turbo87: would it make sense to only collect the `id` properties in the mocked `deprecate()` function above to simplify the logic here?

canufeel: You are probably right. It would simplify the overall logic

PR #200, opened 2017-04-02
Updates to ensure ember-data from npm works properly.

rwjblue:


Turbo87:
I don't get it... why is this necessary?

PR #202, opened 2017-04-06
[FEATURE] Export helpers that add help tracking requests

cibernox:
This very simple addition already helps with 80% of what I want to 
accomplish. I think that all networking libraries should increment
the counter of pending ajax requests.

If this two utilities are expored networking libraries can use them.

I decided to leave the `ajax` terminology in the options the waiter
can receive because it is a defacto synonym of “request”.

### DIFF
--- a/lib/ember-test-helpers/wait.js
+++ b/lib/ember-test-helpers/wait.js
@@ -4,33 +4,30 @@ import Ember from 'ember';
 
 const jQuery = Ember.$;
 
-var requests;
-function incrementAjaxPendingRequests(_, xhr) {
-  requests.push(xhr);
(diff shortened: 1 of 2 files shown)


rwjblue:
This seems like a good refactor either way, but I don't think it solves the underlying issue that you raised https://github.com/emberjs/rfcs/issues/218 for.

cibernox:
@rwjblue I removed the exports for now to make it just refactor, and if I can convince you I'll make them exports and add a babel preprocessor.

cibernox:
Do not merge. It turns out it was necessary to keep track of the XHR requests in an array. I'm going to close it.

PR #205, opened 2017-04-17
Autoregister event dispatcher in unit tests

cibernox:
This allows `ember-native-dom-event-dispatcher` to be registered and hence unit tests to run without jQuery.

### DIFF
--- a/addon-test-support/test-module-for-component.js
+++ b/addon-test-support/test-module-for-component.js
@@ -111,6 +111,10 @@ export default class extends TestModule {
       thingToRegisterWith.register(layoutName, layout);
       thingToRegisterWith.injection(this.subjectName, 'layout', layoutName);
     }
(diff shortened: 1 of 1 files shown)


cibernox:
@rwjblue Yep, it was relatively simple.

rwjblue:
Looks like we have an issue with the canary build (but not beta). It's likely not an issue here (it would be an issue upstream in ember), but we should try to confirm.

cibernox:
I triggered the build again to check that it wasn't random, and it isn't.

rwjblue:
I opened https://github.com/emberjs/ember.js/pull/15199 to address the failure in 2.14 beta and 2.15 canary. I'll keep the underlying issue https://github.com/emberjs/ember.js/pull/15198 open until we confirm that all is well.

rwjblue:
Rebased onto 0.7.x changes (migration to a "real" addon), I believe this should pass now.

PR #212, opened 2017-08-18
[WIP] Workaround example

aureliosaraiva:
https://github.com/emberjs/ember.js/issues/15590

```javascript
it('its not working', function() {
    const modelFake = EmberObject.create({
      fullName: 'foo bar'
    });

    set(this.getContext(), 'model', modelFake);

    this.render(hbs`{{my-component model=model}}`);

    expect(this.$()).to.have.length(1);
  });
```


Turbo87:
Unfortunately using the context like that has several side-effects that IMHO we should avoid. Since `this.set(...)` works as expected I don't see any reason to support `set(this, ...)`.

PR #215, opened 2017-10-08
Reduce direct jQuery usage in tests.

rwjblue:
* add test scenarios that run without jQuery (defaulting to jQueryless setup)
* refactor all tests to avoid jQuery usage
* embed custom eventing test helpers (until we can bring ember-native-dom-helpers into the fold)

### DIFF
--- a/tests/unit/test-module-for-component-test.js
+++ b/tests/unit/test-module-for-component-test.js
@@ -14,6 +14,7 @@ import hasEmberVersion from 'ember-test-helpers/has-ember-version';
 import { setResolverRegistry } from '../helpers/resolver';
 import wait from 'ember-test-helpers/wait';
 import qunitModuleFor from '../helpers/qunit-module-for';
+import hasjQuery from '../helpers/has-jquery';
 import hbs from 'htmlbars-inline-precompile';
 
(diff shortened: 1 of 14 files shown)


tests/unit/test-module-for-component-test.js
```diff
@@ -760,7 +763,7 @@ moduleForComponent('Component Integration Tests: willDestoryElement', {
       'component:my-component': Component.extend({
         willDestroyElement() {
           var stateIndicatesInDOM = this._state === 'inDOM';
-          var actuallyInDOM = $.contains(document, this.$()[0]);
+          var actuallyInDOM = document.contains(this._element);
 
           QUnit.config.current.assert.ok(
```
trentmwillis: Should probably break this up into two assertions so you can get more granular feedback when it fails.

rwjblue: Ya, great point!

rwjblue:
Updated with lots more changes. Tests are passing locally without jQuery (and scenarios were added to ember-try config to test with and without jQuery).  Playing a bit of whack-a-mole with various issues in CI, but generally speaking I think this pretty close...

PR #218, opened 2017-10-10
Reorganize repo internals to prepare for RFC 232 implementation.

rwjblue:
Moves all of the future deprecated code into a `legacy-0-6-x` subfolder (while preserving import paths) in order to ease implementation of emberjs/rfcs#232.


Turbo87:
@rwjblue you one nitpick: it's RFC 232, not 230 😉

rwjblue:
Ugh. I keep referring to the wrong one 🤦‍♂️  Thank you for pointing that out.

PR #220, opened 2017-10-15
Basic implementation for new testing API.

rwjblue:
* Add `teardownContext` function.
* Add `setupContext` function.
* Add `setupRenderingContext`.
* Add `teardownRenderingContext`.

This is intended to be used to implement the new QUnit API via a small ember-qunit wrapper something like this:

```js
import { setupContext, teardownContext } from 'ember-test-helpers';

export default function(hooks) {
 hooks.beforeEach(function() {
   setupContext(this);
 });

  hooks.afterEach(function() {
   teardownContext(this);
 });
}
```

TODO:

- [ ] Port over more tests from current `test-module-for-component-test.js` file (to have better coverage of the various edge cases seen)
- [x] Fix issues with older Ember in rendering tests
  - [x] Ember 2.8
  - [x] Ember 2.4
- [x] Review (likely @Turbo87)
- [x] Add AJAX + RSVP setup and tear down.
- [ ] Assertions upon cleanup around scheduled async timers and whatnot


addon-test-support/build-owner.js
```diff
@@ -0,0 +1,104 @@
...
+    __registry__: registry,
+    __container__: null,
+  });
+
+  container = registry.container({ owner: owner });
+  owner.__container__ = container;
+
+  // TODO: this manual Ember Data setup should be removed in favor of
+  // running the applications `initializers` to ensure the container is
+  // properly setup, this would only need to be done once per test suite
```
Turbo87: running the initializers automatically would change the semantic of the tests though which might make the migration path a little harder 🤔

addon-test-support/setup-context.js
```diff
@@ -0,0 +1,40 @@
+import { run } from '@ember/runloop';
+import { set, setProperties, get, getProperties } from '@ember/object';
+import buildOwner from './build-owner';
+
+/*
+ * Responsible for:
+ *
+ * * create an owner object and set it on the provided context (e.g. this.owner)
+ * * setup this.set, this.setProperties, this.get, and this.getProperties to the provided context
```
Turbo87: might want to use `-` here instead of `*` since the comment block itself already uses `*` 😉

addon-test-support/setup-context.js
```diff
@@ -0,0 +1,40 @@
...
+    return ret;
+  };
+
+  context.get = function(key) {
+    return get(context, key);
+  };
+
+  context.getProperties = function(...args) {
+    return getProperties(context, args);
+  };
```
Turbo87: should we also support `toggleProperty()`?

tests/unit/setup-context-test.js
```diff
@@ -0,0 +1,110 @@
...
+        typeof owner.lookup,
+        'function',
+        'has expected lookup interface'
+      );
+
+      if (hasEmberVersion(2, 12)) {
+        assert.equal(
+          typeof owner.factoryFor,
+          'function',
+          'has expected lookup interface'
```
Turbo87: `'has expected factoryFor interface'`?

tests/unit/setup-context-test.js
```diff
@@ -0,0 +1,110 @@
...
+        );
+      }
+    });
+
+    test('it uses the default resolver if no override specified', function(
+      assert
+    ) {
```
Turbo87: prettier doing some strange formatting here? should we increase the max line length to something like 100 or 120?

tests/unit/setup-context-test.js
```diff
@@ -0,0 +1,110 @@
...
+      let { owner } = context;
+      let instance = owner.lookup('service:foo');
+      assert.equal(instance.isFoo, true, 'uses the default resolver');
```
Turbo87: wouldn't the `lookup()` call already fail and cause an exception?

addon-test-support/setup-rendering-context.js
```diff
@@ -0,0 +1,135 @@
...
+  let dispatcher =
+    owner.lookup('event_dispatcher:main') || Ember.EventDispatcher.create();
+  dispatcher.setup({}, '#ember-testing');
+
+  let OutletView = owner.factoryFor
+    ? owner.factoryFor('view:-outlet')
+    : owner._lookupFactory('view:-outlet');
+  let OutletTemplate = owner.lookup('template:-outlet');
+  let toplevelView = OutletView.create(); // TODO: stash this somewhere?
+  let hasOutletTemplate = !!OutletTemplate;
```
Turbo87: I usually prefer `Boolean(OutletTemplate)` because it's less cryptic

addon-test-support/setup-rendering-context.js
```diff
@@ -0,0 +1,135 @@
...
+    });
+  }
+
+  context.render = function render(template) {
+    if (!template) {
+      throw new Error('you must pass a template to `render()`');
+    }
+
+    return resolve().then(function asyncRenderSetup() {
+      let templateFullName = 'template:-undertest-' + ++templateId;
```
Turbo87: I understand what it does, but IMHO the following is easier to read and I don't think raw performance is *that* important here

```js
templateId += 1;
let templateFullName = `template:-undertest-${templateId}`;
```

addon-test-support/setup-rendering-context.js
```diff
@@ -0,0 +1,135 @@
...
+      toplevelView.setOutletState(outletState);
+    });
+  }
+
+  context.render = function render(template) {
+    if (!template) {
+      throw new Error('you must pass a template to `render()`');
+    }
+
+    return resolve().then(function asyncRenderSetup() {
```
Turbo87: has `this.render()` been async before?

addon-test-support/setup-rendering-context.js
```diff
@@ -0,0 +1,135 @@
...
+
+          // using next here because the actual rendering does not happen until
+          // the renderer detects it is dirty (which happens on backburner's end
+          // hook), see the following implementation details:
+          //
+          // * [view:outlet](https://github.com/emberjs/ember.js/blob/f94a4b6aef5b41b96ef2e481f35e07608df01440/packages/ember-glimmer/lib/views/outlet.js#L129-L145) manually dirties its own tag upon `setOutletState`
+          // * [backburner's custom end hook](https://github.com/emberjs/ember.js/blob/f94a4b6aef5b41b96ef2e481f35e07608df01440/packages/ember-glimmer/lib/renderer.js#L145-L159) detects that the current revision of the root is no longer the latest, and triggers a new rendering transaction
+
+          // TODO: should we resolve with the element? might be nice
+          // for easier assertions...
```
Turbo87: I don't see the benefit if we have `this.element` at the same time 🤔

addon-test-support/setup-rendering-context.js
```diff
@@ -0,0 +1,135 @@
...
+
+  Object.defineProperty(context, 'element', {
+    enumerable: true,
+    configurable: true,
+    get() {
+      return element;
+    },
+  });
+
+  if (jQuery) {
```
Turbo87: should we check `window.jQuery` here instead of relying on the import being transpiled as it is right now?

addon-test-support/setup-rendering-context.js
```diff
@@ -0,0 +1,135 @@
...
+    context.$ = function $(selector) {
+      // emulates Ember internal behavor of `this.$` in a component
+      // https://github.com/emberjs/ember.js/blob/v2.5.1/packages/ember-views/lib/views/states/has_element.js#L18
+      return selector ? jQuery(selector, element) : $(element);
```
Turbo87: shouldn't that be `jQuery(element)` instead of `$(element)`?

addon-test-support/setup-rendering-context.js
```diff
@@ -0,0 +1,135 @@
...
+    };
+  }
+
+  context.clearRender = function clearRender() {
+    return new Promise(function async_clearRender(resolve) {
```
Turbo87: why are we using `resolve().then()` above and a raw `Promise` here?

addon-test-support/teardown-rendering-context.js
```diff
@@ -0,0 +1 @@
+export default function() {}
```
Turbo87: guess this is for future extensibility? should this call `ctx.clearRender()`?

tests/unit/setup-rendering-context-test.js
```diff
@@ -0,0 +1,140 @@
...
+
+  hooks.afterEach(function() {
+    teardownContext(this);
+  });
+
+  test('render exposes an `.element` property', function(assert) {
+    return this.render(hbs`<p>Hello!</p>`).then(() => {
+      assert.equal(this.element.textContent, 'Hello!');
+    });
+  });
```
Turbo87: since we're using headless Chrome we should be able to use native async/await in the tests, right?

tests/unit/setup-rendering-context-test.js
```diff
@@ -0,0 +1,140 @@
...
+
+    assert.equal(
+      this.element,
+      undefined,
+      'precond - this.element is not set sync after this.render'
+    );
+
+    return renderPromise.then(() => {
+      assert.equal(this.element.textContent, 'Hello!');
+    });
```
Turbo87: This seems redundant since the `.then()` above has already checked the same thing?!

tests/unit/setup-rendering-context-test.js
```diff
@@ -0,0 +1,140 @@
...
+      });
+  });
+
+  if (hasjQuery()) {
+    test('this.$ is exposed when jQuery is present', function(assert) {
+      return this.render(hbs`<p>Hello!</p>`).then(() => {
+        assert.equal(this.$().text(), 'Hello!');
+      });
+    });
+  }
```
Turbo87:
```
(hasjQuery() ? test : skip)('this.$ ...
```

tests/unit/setup-rendering-context-test.js
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  test('can invoke JS only components', function(assert) {
+    return this.render(hbs`{{js-only}}`).then(() => {
+      assert.ok(
+        this.element.querySelector('.js-only'),
+        'element found for js-only component'
+      );
+    });
+  });
```
Turbo87: do we have/need a testcase for template+JS components?

tests/unit/setup-rendering-context-test.js
```diff
@@ -0,0 +1,140 @@
...
+
+  test('can pass arguments to helper from context', function(assert) {
+    this.set('name', 'james');
+
+    return this.render(hbs`{{jax name}}`).then(() => {
+      assert.equal(this.element.textContent, 'james-jax');
+    });
+  });
```
Turbo87: should we check the same for passing args to components or is it doing the same thing internally?

addon-test-support/build-owner.js
rwjblue: Ya, my thinking is basically that we should _not_ need to hard code customized setup for ember-data here.  There are basically two options to accomplish that goal:

1. Run initializers (but not instance-initializers) _only once_ for the full suite, just to build up the fallback registry for lookups (this essentially emulates the way fastboot works for example by creating a single registry that is shared amongst many different app instances).
2. Remove the need for ember-data to have these initializers.  

I think both are actually valid options...

Turbo87: second option seems easier to me

maybe we could also make it configurable via an option if the initializers should be run for the test or not?

addon-test-support/setup-context.js
rwjblue: We _could_, but it wasn't included in the initial RFC (or the now "legacy" system) so I'd like to hold off for now.  I'm not saying it requires an RFC, just that I want to implement the current thing without having a moving target...

addon-test-support/setup-rendering-context.js
rwjblue: Nope, it is a change specifically mentioned as a large change in the emberjs/rfcs#232.

addon-test-support/setup-rendering-context.js
rwjblue: Ya, seems good, I'll remove the TODO.

tests/unit/setup-context-test.js
rwjblue: No, lookup will return `undefined` if it does not find a result.

tests/unit/setup-rendering-context-test.js
rwjblue: Yes, we can, but we have to take care not to use it in non-test parts.  I think I have a nice eslint rule for this, will try it out...

tests/unit/setup-rendering-context-test.js
rwjblue: ya, still working to copy over more test cases.  The internal implementation is essentially the same for all types of things, but there were actual regressions that were being tested in the "legacy" tests that I want to carry forward...

tests/unit/setup-rendering-context-test.js
rwjblue: it is doing the same, but I will definitely add tests for it

tests/unit/setup-context-test.js
Turbo87: > No, owner.lookup will return undefined if it does not find a factory to instantiate.

exactly, and then evaluating `instance.isFoo` would result in an `undefined has no isFoo property` error instead of a failing assertion.

addon-test-support/build-owner.js
rwjblue: @Turbo87 - Just to clarify (because your last comment seemed to indicate that we aren't on the same page), I am **not** suggesting that we run initializers for each test.  I am saying that we should run them once for the entire test suite, then use the resulting registry as our fallback registry in this script (notice how we setup the fallback registry system here?).  The only purpose for `initializers` at this point (since sometime late in 1.x) is to setup the registry (e.g. add automatic injections, register manual factories, etc), this "registry setup" work is _exactly_ what we actually want to run.  Conversely, the `instance-initializers` are specifically intended to be able to work with instances (e.g. they can `.lookup`) from the container, which is _exactly_ what we want to prevent.

addon-test-support/build-owner.js
```diff
@@ -42,11 +39,7 @@ export default function(resolver = getResolver()) {
   function register(name, factory) {
     let thingToRegisterWith = registry || container;
 
-    if (
-      !(container.factoryFor
-        ? container.factoryFor(name)
-        : container.lookupFactory(name))
-    ) {
+    if (!(container.factoryFor ? container.factoryFor(name) : container.lookupFactory(name))) {
```
Turbo87: maybe we should extract an `existingFactory` variable here to make the condition easier to read

addon-test-support/setup-rendering-context.js
rwjblue: ya, fixed this up a bit (using `next` always now to ensure "async land").

addon-test-support/teardown-rendering-context.js
rwjblue: ya, it got content in a later commit

tests/unit/setup-context-test.js
rwjblue: Sorry, I don't grok what you mean.  Are you saying that I should remove the `assert.equal(instance.isFoo, true` line? Or something else?

addon-test-support/build-owner.js
Turbo87: > The only purpose for initializers at this point (since sometime late in 1.x) is to setup the registry

Yeah, guess I'm just a little worried about people misusing these things for other purposes

tests/unit/setup-context-test.js
Turbo87: I guess what I'm saying is

```
assert.ok(owner.lookup('service:foo'))
```

would probably be sufficient here and result in a proper failing assertion instead of throwing an Error

rwjblue: Ah, I see.  Ya, I wanted to make sure that the "default" resolver was used (which has `isFoo` set). There is a test down below that also has a `service:foo` but not `isFoo: true`.

PR #223, opened 2017-10-16
Expose `settled` helper function.

rwjblue:
* Rename `wait` to `settled`
* Ensure existing `import wait from 'ember-test-helpers/wait'` code works properly
* Export `settled` as public API from the main `index.js`.
* Move existing `wait` tests into `legacy-0-6-x` folder
* Duplicate and refactor existing `wait` tests to the new API style (`setupRenderingContext`, async/await, etc)

The primary reasoning for the rename from `wait` to `settled`:

```js

test('can do xyz', async function(assert) {
  await this.render(hbs`{{some-thing}}`);
  await click('.some-thing');
  await wait(); // <- WAT?!?!!?
});
```

This looks much better:

```js
test('can do xyz', async function(assert) {
  await this.render(hbs`{{some-thing}}`);
  await click('.some-thing');
  await settled();
});
```


addon-test-support/settled.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export default function settled(_options) {
+  let options = _options || {};
+  let waitForTimers = options.hasOwnProperty('waitForTimers') ? options.waitForTimers : true;
+  let waitForAJAX = options.hasOwnProperty('waitForAJAX') ? options.waitForAJAX : true;
+  let waitForWaiters = options.hasOwnProperty('waitForWaiters') ? options.waitForWaiters : true;
+
+  return new EmberPromise(function(resolve) {
+    let watcher = self.setInterval(function() {
```
Turbo87: should this use the new `./global` util instead of `self`?

addon-test-support/wait.js
```diff
@@ -1,98 +1,7 @@
...
-    }, 10);
-  });
-}
+export {
+  default,
+  _setupAJAXHooks,
+  _setupPromiseListeners,
+  _teardownAJAXHooks,
+  _teardownPromiseListeners,
+} from './settled';
```
Turbo87: this is fine for now, I'm just wondering if we should deprecate importing this module at some point?

tests/unit/settled-test.js
```diff
@@ -0,0 +1,270 @@
...
+
+    fireEvent(this.element.querySelector('div'), 'click');
+
+    await settled({ waitForAJAX: false });
+
+    assert.equal(testContext.element.textContent, 'Local Data!');
+  });
+
+  test('it waits for Ember test waiters', async function(assert) {
+    await this.render(hbs`{{x-test-5}}`);
```
Turbo87: since `render()` without `this` is apparently meant to be the documented default should we use it here too?

rwjblue: Ya, but I didn’t want to do that here since I wasn’t sure how you felt about the other PR yet.

addon-test-support/settled.js
rwjblue: Definitely, will update.

addon-test-support/wait.js
rwjblue: Ya, totally agreed. Basically, I want to introduce all the new functionality _without_ deprecating anything at first. Then as we get more sure of the new system (e.g. after a little while of using the new API’s by default) we add in all the deprecations. 

I want to be very careful about deprecation spew leading in to Ember 3.0...

PR #224, opened 2017-10-17
Continue to flesh out more tests for new API's.

rwjblue:
Carries over many tests from the previous (now "legacy-0-6-x" suite) into the new system. This is largely to ensure that the regressions previously caught by these tests, are still checked for long after the legacy-0-6-x suite is gone.

### DIFF
--- a/tests/unit/setup-context-test.js
+++ b/tests/unit/setup-context-test.js
@@ -1,5 +1,5 @@
 import { module, test } from 'qunit';
-import Service from '@ember/service';
+import Service, { inject as injectService } from '@ember/service';
 import { setupContext, getContext } from 'ember-test-helpers';
 import hasEmberVersion from 'ember-test-helpers/has-ember-version';
 import { setResolverRegistry, createCustomResolver } from '../helpers/resolver';
@@ -90,9 +90,33 @@ module('setupContext', function(hooks) {
         })
       );
 
-      let subject = context.owner.lookup('service:Foo');
+      let subject = context.owner.lookup('service:foo');
 
       assert.equal(subject.someMethod(), 'hello thar!');
+    });
+
+    test('can access a service instance (instead of this.inject.service("thing") in 0.6)', function(
+      assert
+    ) {
+      context.owner.register('service:bar', Service.extend());
+      context.owner.register(
+        'service:foo',
+        Service.extend({
+          bar: injectService(),
+          someMethod() {
+            this.set('bar.someProp', 'derp');
+          },
+        })
+      );
+
+      let subject = context.owner.lookup('service:foo');
+      let bar = context.owner.lookup('service:bar');
+
+      assert.notOk(bar.get('someProp'), 'precond - initially undefined');
+
+      subject.someMethod();
+
+      assert.equal(bar.get('someProp'), 'derp', 'property updated');
     });
   });
 
--- a/tests/unit/setup-context-test.js
+++ b/tests/unit/setup-context-test.js
@@ -79,6 +79,21 @@ module('setupContext', function(hooks) {
     test('it calls setContext with the provided context', function(assert) {
       assert.equal(getContext(), context);
     });
+
+    test('can be used for unit style testing', function(assert) {
+      context.owner.register(
+        'service:foo',
+        Service.extend({
+          someMethod() {
+            return 'hello thar!';
+          },
+        })
+      );
+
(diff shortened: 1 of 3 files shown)


tests/unit/setup-context-test.js
```diff
@@ -79,6 +79,21 @@ module('setupContext', function(hooks) {
...
+          someMethod() {
+            return 'hello thar!';
+          },
+        })
+      );
+
+      let subject = context.owner.lookup('service:Foo');
+
+      assert.equal(subject.someMethod(), 'hello thar!');
+    });
```
Turbo87: should we test `this.inject('foo')` here too?

tests/unit/setup-context-test.js
```diff
@@ -79,6 +79,21 @@ module('setupContext', function(hooks) {
...
+      context.owner.register(
+        'service:foo',
+        Service.extend({
+          someMethod() {
+            return 'hello thar!';
+          },
+        })
+      );
+
+      let subject = context.owner.lookup('service:Foo');
```
Turbo87: I suppose the uppercase `Foo` was deliberate? could probably use a comment in that case

tests/unit/setup-rendering-context-test.js
```diff
@@ -115,4 +117,187 @@ module('setupRenderingContext', function(hooks) {
...
+    );
+    this.owner.register(
+      'template:components/x-foo',
+      hbs`<button {{action 'clicked'}}>Click me!</button>`
+    );
+
+    await this.render(hbs`{{x-foo}}`);
+
+    assert.equal(this.element.textContent, 'Click me!', 'precond - component was rendered');
+    click(this.element.querySelector('button'));
```
Turbo87: I assume at some point passing in just a CSS selector will also be supported?

tests/unit/setup-rendering-context-test.js
```diff
@@ -115,4 +117,187 @@ module('setupRenderingContext', function(hooks) {
...
+      })
+    );
+
+    await this.render(hbs`{{my-input value=value}}`);
+
+    let input = this.element.querySelector('input');
+    input.value = '1';
+
+    fireEvent(input, 'change');
+    assert.equal(this.get('value'), '1');
```
Turbo87: precondition assertion might make sense here

tests/unit/setup-rendering-context-test.js
```diff
@@ -115,4 +117,187 @@ module('setupRenderingContext', function(hooks) {
...
+      hbs`<button {{action 'clicked'}}>{{foo}}</button>`
+    );
+
+    await this.render(hbs`{{my-component foo=foo bar=bar}}`);
+    click(this.element.querySelector('button'));
+
+    await this.clearRender();
+
+    assert.equal(this.get('foo'), 'updated!');
+    assert.equal(this.get('bar'), 'updated bar!');
```
Turbo87: why two properties? because one of them is rendered and the other is not? might need a comment 🤔

tests/unit/setup-context-test.js
rwjblue: Haha, nope, not intentional. Will fix.

tests/unit/setup-context-test.js
rwjblue: Well, that API is no longer present (see examples and reasoning in the RFC), but I will add an example to both these tests and the rendering tests that will be a better demo of the go forward way to do it...

tests/unit/setup-rendering-context-test.js
rwjblue: Ya, these event helpers are only internal test things for now. My next RFC will be for adding the real ones here (and they will essentially have the same API as ember-native-dom-helpers).

tests/unit/setup-rendering-context-test.js
rwjblue: Kk, sounds good. The idea was to confirm that two way binding works regardless of if the property was rendered.

tests/unit/setup-rendering-context-test.js
Turbo87: would that mean deprecating `ember-native-dom-helpers`?

rwjblue: I don’t know, but possibly. The main differences from what we need here is that we do not want to defer to Ember’s own internal application test globals. It’s definitely possible to do this transition a number of ways...

PR #229, opened 2017-10-20
Ensure ember-testing-container is reset properly.

### LINKED ISSUE #228: New `setupRenderingTest` seems to recreate the `#ember-testing-container` for every test
Run the tests of an app with several tests. P.e. this branch of `ember-basic-dropdown`: https://github.com/cibernox/ember-basic-dropdown/pull/349

Check the dom after the suite has finished: 
screen shot 2017-10-19 at 01 01 32

rwjblue:
The changes made in 7ad93d2c introduced a bug in the way that the testing DOM is cleaned up between tests. The error was quite simple: we were storing the `storedValue = container.outerHTML` before starting the test then resetting `container.innerHTML = storedValue`.  This created a DOM node leak for each test processed.

This fixes the issue for both the new testing API's in 0.7 and the older pre-existing system. It also adds some post-test checks to ensure that the DOM is indeed what we expect after each test is completed.

Fixes #228.

### DIFF
--- a/addon-test-support/legacy-0-6-x/abstract-test-module.js
+++ b/addon-test-support/legacy-0-6-x/abstract-test-module.js
@@ -104,15 +104,22 @@ export default class {
   }
 
   setupTestElements() {
+    let testElementContainer = document.querySelector('#ember-testing-container');
+    if (!testElementContainer) {
+      testElementContainer = document.createElement('div');
+      testElementContainer.setAttribute('id', 'ember-testing-container');
+      document.body.appendChild(testElementContainer);
+    }
+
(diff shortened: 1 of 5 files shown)


Turbo87:
Why exactly do we need two divs?

rwjblue:
We have had them historically (I believe the container is used for styling the preview window stuff), this library just ignored the fact that we had two which allows this little “trick” (using innerHTML of the container div to ensure the attributes of the ember-testing div are fully reset).

FWIW, in qunit land (not sure how mocha does this) we should probably just embed the container _inside_ the qunit-fixture div and all of this cleanup would be taken care of for us.

Issue #234, opened 2017-11-05
Use `Application.buildInstance` if possible to create `owner` for use in tests.


rwjblue:
`Application.prototype.buildInstance` is _exactly_ what is done in "reality" to build the owner object. We should migrate to a setup where the host application sets up the application under test (essentially replacing what is done via`setResolver`) in its `test-helper.js` and then for each test we  do something like `this.owner = previouslyCreatedApplicationInstance.buildInstance();`.

This would essentially provide the following benefits:

* Run initializers removing the (fairly large) divide between "acceptance" and "integration" tests
  * Automatically run initializers once (for the entire test suite)
  * Run instance-initializers per-test
* Share the registry for all tests (avoiding wasted work per test)
* Align our setup process for all test types in preparation for handling acceptance tests in the same manner.
* Remove the need for significant amounts of private API usage around `Ember._RegistryProxyMixin` / `Ember._ContainerProxyMixin`

PR #237, opened 2017-11-05
Use `Application.create().buildInstance()` if possible.

rwjblue:
When the host project properly registers their application, the testing context will now boot a lightweight instance of the application (similar to what FastBoot does).

In this scenario both `initializers` and `instance-initializers` will run so no more manual initializer setup shenanigans (as [documented here fore mirage](http://www.ember-cli-mirage.com/docs/v0.3.x/manually-starting-mirage/)). This has been a very long standing point of confusion for folks while testing, and I'm **very** excited to finally have a path forward that is both performant and intuitive.

For those curious, `initializers` will run only once for the entire test suite (unlike acceptance tests prior to https://github.com/emberjs/rfcs/pull/268), and the `instance-initializers` will run once per test (which also mirrors how FastBoot works internally).

This new system nearly identically matches how _real_ applications work in the wild, and is a much less brittle platform for our continued testing "unification".

Fixes https://github.com/emberjs/ember-test-helpers/issues/234

### DIFF
--- a/tests/dummy/app/resolver.js
+++ b/tests/dummy/app/resolver.js
@@ -1,3 +1,12 @@
 import Resolver from 'ember-resolver';
 
-export default Resolver;
+export let registry = Object.create(null);
+export function setRegistry(newRegistry) {
+  registry = newRegistry;
+}
+
+export default Resolver.extend({
+  resolve(fullName) {
+    return registry[fullName] || this._super(...arguments);
+  },
+});
--- a/addon-test-support/application.js
+++ b/addon-test-support/application.js
@@ -1,0 +1,9 @@
+var __application__;
+
+export function setApplication(application) {
+  __application__ = application;
+}
+
+export function getApplication() {
+  return __application__;
+}
--- a/addon-test-support/build-owner.js
+++ b/addon-test-support/build-owner.js
@@ -4,29 +4,37 @@ import global from './global';
 import ApplicationInstance from '@ember/application/instance';
 import Application from '@ember/application';
 import EmberObject from '@ember/object';
+import { Promise } from 'rsvp';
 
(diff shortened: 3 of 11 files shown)


Turbo87:
@rwjblue looks like CI is unhappy with this change

rwjblue:
Yep! Still WIP for now. I have a few things to do first:

- [ ] Add two test runs (one with no application set, and another with it set)
- [ ] Fix tests that provide a custom resolver per module (likely by allowing to provide a custom application instead)
- [ ] Fix some timing related failures (I believe these are cascading failures, but need to confirm)

rwjblue:
@Turbo87 - This is ready for another round of review.

tests/dummy/app/resolver.js
```diff
@@ -1,3 +1,12 @@
...
+export let registry = Object.create(null);
+export function setRegistry(newRegistry) {
+  registry = newRegistry;
+}
+
+export default Resolver.extend({
+  resolve(fullName) {
+    return registry[fullName] || this._super(...arguments);
+  },
+});
```
Turbo87: is that going to be required for all apps?

rwjblue:
> I have to admit that I have no clue of what I'm looking at here so this is quite hard for me to review 

No problem, thanks for taking a look!

tests/dummy/app/resolver.js
rwjblue: No, absolutely not!  This is _only_ a thing needed to test things internally here.  Basically the `setResolverRegistry` test helper function (just here in the tests of this repo) resets the list of "known" modules so that we can test things like `this.owner.register` "wins" over what is resolved, and whatnot.

PR #238, opened 2017-11-08
Random cleanup...

rwjblue:


tests/test-helper.js
```diff
@@ -43,7 +43,9 @@ QUnit.testDone(function({ module, name }) {
   // this is used to ensure that no tests accidentally leak `Ember.testing` state
   if (Ember.testing) {
     throw new Error(
-      `Ember.testing should be reset after test has completed. ${module}: ${name} did not reset Ember.testing`
+      `Ember.testing should be reset after test has completed. ${module}: ${
+        name
+      } did not reset Ember.testing`
```
Turbo87: ugh... how is that "prettier"?! :-/

rwjblue: Hehe. I felt the same way. Unfortunately I was getting trolled by ember-try using npm to install bower (which ultimately ended up forcing _all_ packages to the latest version regardless of useYarn flag) so I couldn’t dig in to how to prevent this. :cry:

PR #239, opened 2017-11-08
Use `setTimeout` for "forced" async.

rwjblue:
Using `Ember.run.next` has a negative impact on developer ergonomics because we loose access to the automatic stack stitching that is done by Chrome (which ends up being very nice when working with `async` / `await`.

Other than improved ergonomics if a failure occurs during setup or teardown, there should be no perceived changes from this PR...

### DIFF
--- a/addon-test-support/setup-context.js
+++ b/addon-test-support/setup-context.js
@@ -1,4 +1,4 @@
-import { run, next } from '@ember/runloop';
+import { run } from '@ember/runloop';
 import { set, setProperties, get, getProperties } from '@ember/object';
 import buildOwner from './build-owner';
 import { _setupPromiseListeners } from './ext/rsvp';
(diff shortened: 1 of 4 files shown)


Turbo87:
CI is unhappy... 😞 

> the automatic stack stitching that is done by Chrome

not sure what this means. do you have a screenshot?

> Other than improved ergonomics if a failure occurs during setup or teardown, there should be no perceived changes from this PR...

what do you mean by "improved ergonomics"?

rwjblue:
Ya, I’ll grab a screenshot to show what I mean shortly.

rwjblue:
> Using Ember.run.next has a negative impact on developer ergonomics because we loose access to the automatic stack stitching that is done by Chrome (which ends up being very nice when working with async / await.

I believe that I may have been suffering from the poor ergonomics that https://github.com/emberjs/ember.js/pull/15871 is addressing.  Closing for now, will revisit if this isn't solved by an Ember bump...

PR #248, opened 2017-11-30
Expose `isSettled` utility function.

rwjblue:
This is a simple refactoring of the existing `settled` helper that exposes an `isSettled()` function which can be used generally to confirm that there are is no pending async.

One example use case is for `ember-qunit` (and also `ember-mocha`) to implement something like:

```js
import { isSettled } from '@ember/test-helpers';

QUnit.testDone(function(results) {
  if (!isSettled) {
    throw new Error(`${results.name} leaks async`);
  }
});
```


rwjblue:
😩  ember-beta is failing now due to https://github.com/emberjs/ember.js/issues/15831 😭

PR #251, opened 2017-12-04
Refactor `isSettled` to allow better insight into current state.

rwjblue:
`isSettled()` will return boolean, and `getSettledState()` will return an object with a number of properties that you can use to see _why_ `isSettled()` returns `false`...

Also:

* Add actual unit tests for `settled`.
* Extract `tests/helpers/ajax.js` to ease testing requests.
* Move existing settled tests to clearly be "integration".

### DIFF
--- a/addon-test-support/@ember/test-helpers/index.js
+++ b/addon-test-support/@ember/test-helpers/index.js
@@ -11,5 +11,5 @@ export {
 export { default as teardownContext } from './teardown-context';
 export { default as setupRenderingContext, render, clearRender } from './setup-rendering-context';
 export { default as teardownRenderingContext } from './teardown-rendering-context';
-export { default as settled, isSettled } from './settled';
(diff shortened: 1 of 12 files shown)


rwjblue:
For any reviewers, its probably going to be easier to review as separate commits...

rwjblue:
The failures here are from moving the tests around (from `tests/unit` to `tests/integration`). Which means..... we have order dependent tests 😭 

https://github.com/emberjs/ember-test-helpers/pull/252 starts tackling issues (identified via `http://localhost:4200/?seed` and re-run until passing routinely)...

rwjblue:
Rebased to include the changes from #252...

Turbo87:
looks like this fails on Ember 2.4 for some reason 🤔

rwjblue:
Ya, it has to do with a bug in 2.4, but I’m going to have to track it down and guard for it.

rwjblue:
FINALLY

PR #252, opened 2017-12-04
Fix order dependence in tests.

rwjblue:
* Refactor cleanup assertions to use global failure after test run.
* Avoid `before` and `after` when randomizing test order.
* Ensure `setupRegistry` is called _before_ it is needed each time.
* Simplify precondition assertion in test-module-for-acceptance tests.
* Make randomized seed testing easier.
* Ensure `unsetContext()` is called during teardown.


rwjblue:
I was able to run `http://localhost:4200/tests/?seed&hidepassed` around 20 times without failing after these changes.

PR #254, opened 2017-12-05
Upgrade default Ember version to 2.17.

rwjblue:
This gives us that sweet sweet error throwing goodness...


rwjblue:
Ya, prettier@1.9 was released and since we ignore lockfiles with ember-try we are always drifting (even though we have a `yarn.lock` ourselves).  I'm working on this situation in a follow up PR...

rwjblue:
See https://github.com/emberjs/ember-test-helpers/commit/e5e71a49460b8623b0127474e1add2f2f28f90ee for the prettier@1.9 related changes.  IMHO, they are all good changes...

PR #255, opened 2017-12-05
Use build stages to fail early...

rwjblue:


Turbo87:
fancy, I definitely support this 👍

package.json
```diff
@@ -25,7 +25,9 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember try:each"
+    "test": "ember try:each",
+    "test:default": "ember test",
```
Turbo87: I actually prefer `test: ember test` and `test:all: ember try:each` 🤔

rwjblue: ya, same, I'll update to that

rwjblue:
@Turbo87 - OK, I'm done playing around with this.  Ready for review again...

.travis.yml
```diff
@@ -36,22 +61,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile
```
rwjblue: I removed this because `ember try:one` (when `useYarn: true` is specified in config) already does `yarn install --no-lockfile --ignore-engines` when it is installing the scenario.

package.json
```diff
@@ -25,7 +25,9 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember try:each"
+    "test": "ember test",
+    "test:all": "ember try:each",
+    "lint": "eslint addon-test-support config tests"
```
Turbo87: this seems to be missing `*.js`

.travis.yml
Turbo87: if it is runnin `yarn install --no-lockfile` and `node_modules` already exists, are you sure it updates all deps to the latest possible versions?

rwjblue: No, I'm unsure.  I'll add an explicit job to the first stage that is explicity testing transitive deps with the default set ("current" ember-source).

package.json
rwjblue: Great point, I'll copy from my other PR in ember-cli.

.travis.yml
rwjblue: OK, updated to add a single `--no-lockfile` job to the first stage. This way we don't have to ignore the lockfile for all jobs, and hopefully with this setup we can easily identify failures due to transient deps vs failures due to _actual_ repo changes.

PR #258, opened 2017-12-06
Bring over DOM helper implementation.

rwjblue:
TODO:

- [x] Add all the helpers in the RFC
  - [x] `click`
  - [x] `tap`
  - [x] `triggerKeyEvent`
  - [x] `triggerEvent`
  - [x] `fillIn`
  - [x] `focus`
  - [x] `blur`
  - [x] `waitUntil`
  - [x] `waitFor`
- [x] Add unit tests for DOM Helpers
  - [x] `click`
  - [x] `tap`
  - [x] `triggerKeyEvent`
  - [x] `triggerEvent`
  - [x] `fillIn`
  - [x] `focus`
  - [x] `blur`
  - [x] `waitUntil`
  - [x] `waitFor`
- [x] Fix tests when running with jQuery (current CI builds without jQuery are passing, but jQuery builds are failing)...
- [x] Account for https://github.com/cibernox/ember-native-dom-helpers/pull/104


addon-test-support/@ember/test-helpers/dom/-get-element.js
```diff
@@ -0,0 +1,16 @@
...
+    selectorOrElement instanceof Document ||
+    selectorOrElement instanceof HTMLElement ||
+    selectorOrElement instanceof SVGElement
+  ) {
+    return selectorOrElement;
+  }
+
+  let rootElement = getContext().element;
+
+  return rootElement.querySelector(selectorOrElement);
```
Turbo87: should we typecheck if `selectorOrElement` is a string?

addon-test-support/@ember/test-helpers/dom/-is-focusable.js
```diff
@@ -0,0 +1,14 @@
+const FOCUSABLE_TAGS = ['INPUT', 'BUTTON', 'LINK', 'SELECT', 'A', 'TEXTAREA'];
```
Turbo87: can a `<link>` actually be focused?

addon-test-support/@ember/test-helpers/dom/blur.js
```diff
@@ -0,0 +1,37 @@
...
+import settled from '../settled';
+import isFocusable from './-is-focusable';
+
+/*
+  @method blur
+  @param {String|HTMLElement} selector
+  @return {RSVP.Promise}
+  @public
+*/
+export default function blur(selectorOrElement) {
```
Turbo87: does it make sense that `blur()` takes a selector? should it default to `document.activeElement`?

addon-test-support/@ember/test-helpers/dom/focus.js
```diff
@@ -0,0 +1,41 @@
...
+  @public
+*/
+export default function focus(selectorOrElement) {
+  if (!selectorOrElement) {
+    throw new Error('Must pass an element or selector to `focus`.');
+  }
+
+  let element = getElement(selectorOrElement);
+
+  if (isFocusable(element)) {
```
Turbo87: should this throw an error or warning if the element is not focusable?

addon-test-support/@ember/test-helpers/dom/blur.js
```diff
@@ -0,0 +1,37 @@
+import getElement from './-get-element';
+import fireEvent from './fire-event';
+import settled from '../settled';
+import isFocusable from './-is-focusable';
+
+/*
```
Turbo87: doc comments should use `/**` instead of `/*`. this seems to be the case for the majority of the doc comments in this PR.

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -0,0 +1,23 @@
+import { merge } from '@ember/polyfills';
+import triggerEvent from './trigger-event';
+
+/**
+ * @public
+ * @param selector
+ * @param type
+ * @param keyCode
+ * @param modifiers
```
Turbo87: the parameters are missing type information

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -0,0 +1,23 @@
...
+export function keyEvent(
+  selectorOrElement,
+  type,
+  keyCode,
+  modifiers = { ctrlKey: false, altKey: false, shiftKey: false, metaKey: false }
+) {
+  return triggerEvent(
+    selectorOrElement,
+    type,
+    merge({ keyCode, which: keyCode, key: keyCode }, modifiers)
```
Turbo87: can we extract that into an `options` variable so that the `triggerEvent()` call stays single-line?

addon-test-support/@ember/test-helpers/wait-until.js
```diff
@@ -0,0 +1,22 @@
...
+      let value = callback();
+      if (value) {
+        resolve(value);
+      } else if (time < timeout) {
+        setTimeout(tick, 10);
+      } else {
+        reject(waitUntilTimedOut);
+      }
+    }
+    setTimeout(tick, 10);
```
Turbo87: should this call `tick()` directly?

tests/unit/helpers/wait-until-test.js
```diff
@@ -0,0 +1,40 @@
...
+module('DOM helper: waitUntil', function() {
+  test('waits until the provided function returns true', async function(assert) {
+    let current = false;
+    assert.step('before invocation');
+    let waiter = waitUntil(() => current).then(() => {
+      assert.step('after resolved');
+
+      assert.verifySteps(['before invocation', 'after invocation', 'after resolved']);
+    });
+    assert.step('after invocation');
```
Turbo87: TIL 🤔

Turbo87:
@cibernox can you have a look at this too?

addon-test-support/@ember/test-helpers/dom/-get-element.js
rwjblue: I thought about that, we _can_ but I was unsure if we could do beat things by allowing Object#toString to be leveraged...

I do not feel strongly though...

addon-test-support/@ember/test-helpers/dom/blur.js
rwjblue: Good catch. Will update.

addon-test-support/@ember/test-helpers/dom/blur.js
rwjblue: Seems reasonable...

addon-test-support/@ember/test-helpers/dom/focus.js
rwjblue: Absolutely

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
rwjblue: Good catch

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
rwjblue: Yep, that would look much better indeed.

addon-test-support/@ember/test-helpers/wait-until.js
rwjblue: I did think about that, but doing that would allow it to resolve sync if wrapped in a run-loop, and we _never_ want to resolve sync.

rwjblue:
One of the major missing things here are the unit and integration tests for the new helpers. I’m still working on porting those. At the moment, the only tests are the ones that we had previously (in setup rendering context and moduleForComponent) that needed events. This is obviously not enough coverage.

addon-test-support/@ember/test-helpers/dom/blur.js
Turbo87: we should check if that works properly with headless browsers and unfocused tabs though

cibernox:
I'll check this more thoroughly tomorrow, but looks good at first glance.

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -0,0 +1,225 @@
...
+  } else if (FILE_SELECTION_EVENT_TYPES.indexOf(type) > -1 && element.files) {
+    event = buildFileEvent(type, element, options);
+  } else {
+    event = buildBasicEvent(type, options);
+  }
+
+  element.dispatchEvent(event);
+}
+
+/*
```
Turbo87: `/**`

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -0,0 +1,225 @@
...
+  delete options.cancelable;
+
+  // bubbles and cancelable are readonly, so they can be
+  // set when initializing event
+  event.initEvent(type, bubbles, cancelable);
+  merge(event, options);
+  return event;
+}
+
+/*
```
Turbo87: `/**`

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -0,0 +1,225 @@
...
+  @method buildMouseEvent
+  @param {String} type
+  @param {Object} (optional) options
```
Turbo87: usejsdoc.org has some special syntax for declaring optional parameters

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -0,0 +1,54 @@
...
+  @param {Boolean} modifiers.ctrlKey
+  @param {Boolean} modifiers.altKey
+  @param {Boolean} modifiers.shiftKey
+  @param {Boolean} modifiers.metaKey
+  @return {Promise<void>}
+*/
+export default function triggerKeyEvent(target, eventType, keyCode, modifiers = DEFAULT_MODIFIERS) {
+  if (!target) {
+    throw new Error('Must pass an element or selector to `triggerKeyEvent`.');
+  }
```
Turbo87: we don't seem to have this check in any of the other helpers above 🤔

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!eventType) {
+    throw new Error(`Must provide an \`eventType\` to \`triggerKeyEvent\``);
+  }
+
+  if (KEYBOARD_EVENT_TYPES.indexOf(eventType) === -1) {
+    throw new Error(
+      `Must provide an \`eventType\` of ${KEYBOARD_EVENT_TYPES.join(
+        ', '
+      )} to \`triggerKeyEvent\` but you passed \`${eventType}\`.`
```
Turbo87: is this prettier doing something strange?

addon-test-support/@ember/test-helpers/wait-until.js
```diff
@@ -0,0 +1,29 @@
...
+    let time = -10;
+    function tick() {
+      time += 10;
+      let value = callback();
+      if (value) {
+        resolve(value);
+      } else if (time < timeout) {
+        // using `setTimeout` directly to allow fake timers
+        // to intercept
+        setTimeout(tick, 10);
```
Turbo87: we should capture the `setTimeout` reference so that Timecop et al can do their thing without disturbing us

tests/unit/dom/blur-test.js
```diff
@@ -0,0 +1,68 @@
...
+module('DOM Helper: blur', function(hooks) {
+  let element;
+
+  hooks.beforeEach(async function(assert) {
+    // used to simulate how `setupRenderingTest` (and soon `setupApplicationTest`)
+    // set context.element to the rootElement
+    this.element = document.querySelector('#qunit-fixture');
+
+    // create the element and focus in preparation for blur testing
+    element = buildInstrumentedElement('input');
```
Turbo87: having `this.element` and `element` be two different things seems a little confusing 🤔

tests/unit/dom/blur-test.js
```diff
@@ -0,0 +1,68 @@
...
+    }, /Element not found when calling `blur\('#foo-bar-baz-not-here-ever-bye-bye'\)`/);
+  });
+
+  test('bluring via selector with context set', async function(assert) {
+    setContext(this);
+    await blur(`#${element.id}`);
+
+    assert.verifySteps(['blur', 'focusout']);
+    assert.notEqual(document.activeElement, element, 'activeElement updated');
+  });
```
Turbo87: what about "bluring via selector without context set"?

tests/unit/dom/blur-test.js
```diff
@@ -0,0 +1,68 @@
...
+import { focus, blur, setContext, unsetContext } from '@ember/test-helpers';
+import { buildInstrumentedElement } from '../../helpers/events';
+
+module('DOM Helper: blur', function(hooks) {
+  let element;
+
+  hooks.beforeEach(async function(assert) {
+    // used to simulate how `setupRenderingTest` (and soon `setupApplicationTest`)
+    // set context.element to the rootElement
+    this.element = document.querySelector('#qunit-fixture');
```
Turbo87: I'm wondering if we should use an explicit `context` object here instead of `this` 🤔

tests/unit/legacy-0-6-x/test-module-for-component-test.js
```diff
@@ -475,8 +475,9 @@ test('it supports updating an input', function(assert) {
   let input = this._element.querySelector('input');
   input.value = '1';
 
-  fireEvent(input, 'change');
-  assert.equal(this.get('value'), '1');
+  return triggerEvent(input, 'change').then(() => {
+    assert.equal(this.get('value'), '1');
+  });
```
Turbo87: seems like we should `assert.expect(1)` here to make sure it actually runs correctly

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
rwjblue: As far as I can tell all of the helpers throw an error if the `target` / `selectorOrElement` is not provided.  Which once did I miss?

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
rwjblue: Yes, indeed.

tests/unit/dom/blur-test.js
rwjblue: I'm largely just trying to leverage the normal system of "rooting the selectors" that would happen in an app.  In those circumstances `this.element` (aka `getContext().element`) is the root element of the app / component test / etc. 

I think a good compromise here might be to use an explicit context _other_ than `this` which has the root element (`#qunit-fixture`) and to not call the local variable `element` here.

I'll work towards that...

cibernox:
I said I would review but this is pretty massive 😵
Wish me luck

addon-test-support/@ember/test-helpers/dom/-get-element.js
```diff
@@ -0,0 +1,22 @@
+import { getContext } from '../setup-context';
+
+export default function getElement(selectorOrElement) {
+  if (
+    selectorOrElement instanceof Window ||
+    selectorOrElement instanceof Document ||
+    selectorOrElement instanceof HTMLElement ||
```
cibernox: This can be simplified. At least, `HTMLElement` and `SVGElement` are both of subclasses of `Element`, so that is a simplification.
Also, both `Element` and `Document` are subclasses of `Node`, but I'm not sure we want to be so generic, as there is other things that inherit from it that might not be valid targets, like `DocumentFragment`s.

addon-test-support/@ember/test-helpers/dom/-is-focusable.js
```diff
@@ -0,0 +1,15 @@
+import isFormControl from './-is-form-control';
+import isContentEditable from './-is-content-editable';
+
+const FOCUSABLE_TAGS = ['LINK', 'A'];
```
cibernox: I really don't know why I listed `LINK` as a focusable tag originally. I think it was a mindfart, it should be gone.

addon-test-support/@ember/test-helpers/dom/-is-form-control.js
```diff
@@ -0,0 +1,11 @@
+const FORM_CONTROL_TAGS = ['INPUT', 'BUTTON', 'SELECT', 'TEXTAREA'];
+
+export default function isFormControl(el) {
+  let { tagName, type } = el;
+
+  if (type === 'hidden') {
+    return false;
+  }
```
cibernox: I think we could improve the logic here. I don't think that disabled buttons or inputs should be focusable.
I found this nice table we can copy from, even if we're not soooo exhaustive: https://allyjs.io/data-tables/focusable.html

addon-test-support/@ember/test-helpers/dom/blur.js
```diff
@@ -0,0 +1,49 @@
...
+  // fire `blur` event via native event.
+  if (browserIsNotFocused) {
+    fireEvent(element, 'blur', { bubbles: false });
+    fireEvent(element, 'focusout');
+  }
+}
+
+/**
+  @method blur
+  @param {String|HTMLElement} [target=document.activeElement] the element to blur
```
cibernox: Perhaps `{String|Element}`, as SVG elements might also be focused and blurred.

addon-test-support/@ember/test-helpers/dom/click.js
```diff
@@ -0,0 +1,32 @@
+import getElement from './-get-element';
+import fireEvent from './fire-event';
+import { __focus__ } from './focus';
+import settled from '../settled';
+import isFocusable from './-is-focusable';
+import { nextTickPromise } from '../-utils';
+
+/**
+  @method click
+  @param {String|HTMLElement} target
```
cibernox: `{String|Element}`

addon-test-support/@ember/test-helpers/dom/blur.js
```diff
@@ -0,0 +1,49 @@
...
+import settled from '../settled';
+import isFocusable from './-is-focusable';
+import { nextTickPromise } from '../-utils';
+
+/**
+  @private
+  @method __blur__
+  @param {HTMLElement} element
+*/
+export function __blur__(element) {
```
cibernox: Why have private methods that are not async and don't return a promise? Perf?

addon-test-support/@ember/test-helpers/dom/fill-in.js
```diff
@@ -0,0 +1,44 @@
...
+import isFormControl from './-is-form-control';
+import isContentEditable from './-is-content-editable';
+import { __focus__ } from './focus';
+import settled from '../settled';
+import fireEvent from './fire-event';
+import { nextTickPromise } from '../-utils';
+
+/*
+  @method fillIn
+  @param {String|HTMLElement} target
```
cibernox: `{String|Element}`

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -0,0 +1,225 @@
...
+  @param {Object} [options]
+  @return {Event}
+  @private
+*/
+function buildMouseEvent(type, options = {}) {
+  let event;
+  try {
+    event = document.createEvent('MouseEvents');
+    let eventOpts = merge(merge({}, DEFAULT_EVENT_OPTIONS), options);
+    event.initMouseEvent(
```
cibernox: IDK if this targets Ember 3.0, but if it does, then we don't need this as the `MouseEvent` constructor works in IE11+

addon-test-support/@ember/test-helpers/dom/-get-element.js
rwjblue: Confirm. I plan to refactor to use `HTMLElement.prototype.nodeType` which should “just work”.

addon-test-support/@ember/test-helpers/dom/blur.js
rwjblue: Basically just for reuse where they are used in another async helper already (e.g. focus is used by click).

addon-test-support/@ember/test-helpers/dom/fire-event.js
rwjblue: I don’t plan to change browser support as part of this effort, however we should add a comment to indicate it can be removed if IE8/IE9 are not supported.

cibernox: 9 and 10. Probably the same will happen to `KeyboardEvent`

rwjblue:
> I said I would review but this is pretty massive

Sorry about that....

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -0,0 +1,226 @@
...
+  'mouseover',
+];
+const FILE_SELECTION_EVENT_TYPES = ['change'];
+
+/**
+  @method fireEvent
+  @param {Element} element
+  @param {String} type
+  @param {Object} [options]
+
```
cibernox: Add `@return {Event}`

addon-test-support/@ember/test-helpers/dom/fire-event.js
cibernox: TODO: Add support for pointer events? That might unify tap/click.

addon-test-support/@ember/test-helpers/dom/trigger-event.js
```diff
@@ -0,0 +1,33 @@
+import getElement from './-get-element';
+import fireEvent from './fire-event';
+import settled from '../settled';
+import { nextTickPromise } from '../-utils';
+
+/**
+  @method triggerEvent
+  @param {String|HTMLElement} target
```
cibernox: `{String|Element}`

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -0,0 +1,54 @@
...
+const DEFAULT_MODIFIERS = Object.freeze({
+  ctrlKey: false,
+  altKey: false,
+  shiftKey: false,
+  metaKey: false,
+});
+
+/**
+  @public
+  @param {String|HTMLElement} target
```
cibernox: `{String|HTMLElement}`

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -0,0 +1,54 @@
...
+  @param {'keydown' | 'keyup' | 'keypress'} eventType
+  @param {String} keyCode
+  @param {Object} [modifiers]
+  @param {Boolean} [modifiers.ctrlKey=false]
+  @param {Boolean} [modifiers.altKey=false]
+  @param {Boolean} [modifiers.shiftKey=false]
+  @param {Boolean} [modifiers.metaKey=false]
+  @return {Promise<void>}
+*/
+export default function triggerKeyEvent(target, eventType, keyCode, modifiers = DEFAULT_MODIFIERS) {
```
cibernox: I've noticed a name change here. In ember-native-dom-helpers it's named just `keyEvent`, which I copied from the current acceptance helper: https://emberjs.com/api/ember/2.15/classes/Ember.Test/methods/keyEvent?anchor=keyEvent&show=inherited%2Cprotected%2Cprivate%2Cdeprecated

I don't oppose to the name change, but it won't be a drop-in replacement for ember-native-dom-helpers.

cibernox: When I initially copied `keyEvent`, I continued to use `keyCode` as the main way to set the key, but it seems to be [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and the new alternatives are [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) and [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key), specially the last one.

Perhaps we might want to start using it, or maybe assume that integers mean `keyCode` and strings mean `key`.

rwjblue: Aye, agreed.  The change was intentional (`keyEvent` "seemed" odd to me when what you are doing is triggering a keyboard event).

I think we can mitigate or at least ease some of the migration concerns by adding a reexport to `ember-native-dom-helpers`...

rwjblue: Agreed, I'll try to review the various docs and update accordingly...

cibernox:
I do have a question. Should this PR also add shims to methods provided by Ember?
I think that it would feel weird to do `import { click } from '@ember/test-helpers'` but not do the same for `visit`, `currentURL`, `currentPath` and `currentRouteName`.

cibernox:
Once this is published out of a beta version I plan to deprecate all the helpers implemented here. If I checked things right, the only two helpers you haven't implemented are `scrollTo` and `selectFiles`. Everything else will be deprecated.

Turbo87:
@cibernox correct me if I'm wrong but `find` and friends are also not implemented, right?

cibernox:
@Turbo87 you're completely right. I my mind I was making a difference between helpers and finders, but that's just how my brain works.

cibernox:
FWIW, although I believe that most of the people should be using things like `qunit-dom`, perhaps adding `find` and `findAll` here could be a good idea. I feel that otherwise people will continue to use the built-in `find` global, and be confused why `find('button:contains("Submit")')` works but `click('button:contains("Submit")')` doesn't.

Turbo87:
@cibernox I tend to agree that having something like find/findAll in here would probably make things a little easier. IMHO we can land that in a separate PR though so that we can finally merge this monster here 😉

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
Turbo87: e.g. `fillIn` and `tap` do not have this check and delegate to `getElement` directly

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
Turbo87: can we fix this by extracting `KEYBOARD_EVENT_TYPES.join(', ')` into a variable?

addon-test-support/@ember/test-helpers/dom/focus.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  let element = getElement(target);
+  if (!element) {
+    throw new Error(`Element not found when calling \`focus('${target}')\`.`);
+  }
+
+  if (!isFocusable(element)) {
+    throw new Error(`${target} is not focusable`);
+  }
```
Turbo87: I think I raised this issue before but maybe it got lost, I'm a little worried about:

- throwing sync errors from functions that return a Promise. from what I've read about Promises so far, the best practice for async functions is to always reject instead of throwing sync errors.

- assertions about the input params are running before the actual focus code. as in, while the `nextTickPromise()` is running the element might be remove from the page which would result in an exception. maybe we should move those conditions into the `.then()` block 🤔

tests/unit/legacy-0-6-x/test-module-for-component-test.js
```diff
@@ -498,11 +501,14 @@ test('it supports dom triggered focus events', function(assert) {
...
-  blur(input);
-  assert.equal(input.value, 'focusout');
+  return focus(input)
+    .then(() => {
+      assert.equal(input.value, 'focusin');
+      return blur(input);
+    })
+    .then(() => {
+      assert.equal(input.value, 'focusout');
+    });
```
Turbo87: y u no async/await?

addon-test-support/@ember/test-helpers/dom/focus.js
rwjblue: Yes, you are correct. I experimented with how async functions work, and confirmed your logic. I will update to rejections before landing.

tests/unit/legacy-0-6-x/test-module-for-component-test.js
rwjblue: Haha. I’m trying to touch those “legacy” tests as little as possible. Basically they are acting as my canary in the coal mine around if I broke existing `moduleFor*` APIs.

rwjblue:
@Turbo87 - Updated to properly reject when arguments are invalid formats (if you wanted to review that).

rwjblue:
@Turbo87 - Updated to properly reject when arguments are invalid formats (if you wanted to review that).

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -0,0 +1,227 @@
...
+/**
+  @method fireEvent
+  @param {Element} element
+  @param {String} type
+  @param {Object} [options]
+  @returns {Event}
+
+  @private
+*/
+export default function fireEvent(element, type, options = {}) {
```
Turbo87: why is this not using `nextTickPromise()`?

addon-test-support/@ember/test-helpers/wait-until.js
```diff
@@ -0,0 +1,36 @@
+import { Promise } from 'rsvp';
+
+import { nextTick } from './-utils';
+
+export default function(callback, options = {}) {
+  let timeout = 'timeout' in options ? options.timeout : 1000;
+  let waitUntilTimedOut = new Error('waitUntil timed out');
```
Turbo87: this could use a comment that the `Error` is created here to have the original stacktrace

addon-test-support/@ember/test-helpers/dom/fire-event.js
rwjblue: Because it is private, not exported from the entry point, and only called from _within_ a nextTickPromise already. I can add some underscores too though 😜...

addon-test-support/@ember/test-helpers/wait-until.js
rwjblue: Sounds good. Will do that.

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
rwjblue: Created https://github.com/emberjs/ember-test-helpers/issues/265 to track this.

rwjblue:
> Perhaps adding find and findAll here could be a good idea. I feel that otherwise people will continue to use the built-in find global, and be confused about why `find('button:contains("Submit")')` works but `click('button:contains("Submit")')` doesn't.

I'm not opposed to bringing `find` and `findAll` over to ease the transition from both `ember-native-dom-helpers` and Ember's own legacy acceptance testing system, but I definitely do not want them to be the way we teach folks. IMHO, using `this.element.querySelector` or `this.element.querySelectorAll` (even though more verbose) is much better than a "magical" `find` in the long run...

Turbo87:
🎉

cibernox:
As soon as there is a beta version I'll start testing it.

chriskrycho:
This is awesome.

But, uhhh… we want typings for these, and I'm gonna get 'em somehow. 😏 The question is: do we…

- add types by making this actually a TS library?
- add types by adding an `index.d.ts` file here?
- add types by putting them on DefinitelyTyped?

My inclination is if we can do the first, that'd be awesomest. I'm happy to help with that process, including doing whatever work to get the build stuff working right.

Issue #245, opened 2017-11-15
Timecop is freezing tests


Turbo87:
We are using https://github.com/matteodepalo/ember-cli-timecop in a few component integration tests and until now this worked fine. We tried to convert those tests to use the new testing APIs, but now `Timecop.freeze()` will block the `async render(...)` call from continuing since it mocks the `setTimeout()` function.

An issue similar to this has existed in the past where Backburner.js would not continue running when `setTimeout()` is mocked and frozen, but was fixed by getting a reference to the `setTimeout()` function on startup and then using that reference instead of the temporary mocked global.

Turbo87:
Since https://github.com/emberjs/ember-test-helpers/pull/239 was not merged yet it seems that we are not actually using `setTimeout()` at all yet, so it must something else using it 🤔

rwjblue:
@Turbo87 - Well, `Ember.run.next` ultimately does use `setTimeout` internally (and we are using `run.next` now).

rwjblue:
If you have the time, could you put together a repo I could use for testing / experimenting with this?

Turbo87:
@rwjblue https://github.com/Turbo87/__timecop-issue

Turbo87:
Interestingly https://github.com/jamesarosen/Timecop.js/issues/24#issuecomment-35451318 notes that:

> Timecop doesn't modify the global `setTimeout` and `setInterval` implementations.

I'm wondering what is causing these freezes then 🤔

rwjblue:
The issue is exactly what you mentioned in the initial description.  https://github.com/BackburnerJS/backburner.js/commit/a6fe19eb26c2b7ed8360de19ca0cc9ae5b05f297 changed `Date.now` to no longer be eagerly captured. The `Ember.run.next`'s that we are calling (in `setupContext` and `setupRenderingContext`) create a scenario where `settled()` would never resolve (since it internally used `Date.now()` to determine if it should run our scheduled callback).

However, I also tested this with the changes from https://github.com/emberjs/ember-test-helpers/pull/239 and it does pass. I suspect that any "real world" scenario would still fail though. Essentially _any_ usage of `run.next` or `run.later` in an app would cause this specific regression...

Turbo87:
How did this work before?

rwjblue:
[...]

PR #262, opened 2017-12-12
settled: Ramp up timeout values for isSettled() check

### LINKED ISSUE #260: Speed up settled()
It seems that each call to `await settled()` is unnecessarily slowing down the test runs since it will *always* wait at least 10ms before resolving. I'd like to propose some sort of exponential backoff for the timer interval that starts with 0ms, then 1ms, 2ms, 5ms, 10ms, 10ms, 10ms, or something like that.

For a test suite of 1000 async tests using about 5 async test helpers each this amounts to 50000ms that might potentially be wasted so this is actually a huge deal 🤔

Turbo87:
Resolves #260

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -107,17 +107,28 @@ export function isSettled(options) {
   return true;
 }
 
+const TIMEOUTS = [0, 1, 2, 5];
+const MAX_TIMEOUT = 10;
+
 export default function settled(options) {
   return new EmberPromise(function(resolve) {
-    let watcher = global.setInterval(function() {
-      let settled = isSettled(options);
-      if (settled) {
-        // Stop polling
-        global.clearInterval(watcher);
-
-        // Synchronously resolve the promise
-        run(null, resolve);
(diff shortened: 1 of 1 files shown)


addon-test-support/@ember/test-helpers/settled.js
```diff
@@ -107,17 +107,28 @@ export function isSettled(options) {
...
-        // Synchronously resolve the promise
-        run(null, resolve);
+    function scheduleCheck(counter) {
+      let timeout = TIMEOUTS[counter];
+      if (timeout === undefined) {
+        timeout = MAX_TIMEOUT;
       }
-    }, 10);
+
+      global.setTimeout(function() {
```
rwjblue: FWIW, I _think_ this is related to your `timecop` issue.  I'll fix to use `nextTick` (which is captured eagerly much earlier) once #258 lands.

Turbo87: except, it's not: https://github.com/emberjs/ember-test-helpers/issues/245#issuecomment-351068278 🤔

Turbo87: should still be changed though, I agree

PR #264, opened 2017-12-13
Wait for pending AJAX in acceptance tests.

rwjblue:
Ember internally tracks AJAX requests in the same way that we do here for legacy style "acceptance" tests using the `ember-testing.js` asset provided by emberjs/ember.js itself. When `@ember/test-helpers`'s `settled` utility is used in a legacy acceptance test context any pending AJAX requests are not properly considered during the `isSettled` check below.

This utilizes a local utility method present in Ember since around 2.8.0 to properly consider pending AJAX requests done within legacy acceptance tests.

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -8,7 +8,32 @@ import global from './global';
 
 // TODO: refactor to use `nextTick` from #258
 const SET_TIMEOUT = global.setTimeout;
+
+// Ember internally tracks AJAX requests in the same way that we do here for
+// legacy style "acceptance" tests using the `ember-testing.js` asset provided
+// by emberjs/ember.js itself. When `@ember/test-helpers`'s `settled` utility
+// is used in a legacy acceptance test context any pending AJAX requests are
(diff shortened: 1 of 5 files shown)


Turbo87:
@rwjblue do you think it's possible to write tests for this?

rwjblue:
yes, I _think_ so, but I am not 100% certain.  I am working on the failing test (the failure is legit) then I'll take a stab at adding a reasonable facsimile of a regression test...

rwjblue:
Just pushed up the fix for the prior failure (leaky state in the legacy acceptance testing system itself), and a regression test. The test was somewhat gnarly to add but I confirmed that the test failed prior to the changes in `addon-test-support/` and passes afterwards.

rwjblue:
FYI - I added a guard to prevent the new regression test from running on Ember versions older than 2.8.  It is _possible_ that I could have made it work, but I'm not sure and this issue has sucked enough time already. 😡   We can always try to address if someone provides a compelling case (and hopefully PR 😝 ), but since Ember 2.4 is already sunsetted I think its ok if some features don't work properly...

Turbo87:
awesome work @rwjblue ❤️

PR #269, opened 2017-12-15
Use `nextTick` and `nextTickPromise` throughout codebase.

rwjblue:
Updates `setupContext` and `setupRenderingContext` to use `nextTickPromise` (simplifying the code nicely).

Also adds `settled()` calls after `render`, `clearRender`, `teardownContext`, and `teardownRenderingContext` to ensure that all run-loop business has been completed before returning.

Fixes https://github.com/emberjs/ember-test-helpers/issues/268.

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -4,10 +4,7 @@ import { Promise as EmberPromise } from 'rsvp';
 import jQuery from 'jquery';
 
 import Ember from 'ember';
-import global from './global';
-
-// TODO: refactor to use `nextTick` from #258
-const SET_TIMEOUT = global.setTimeout;
+import { nextTick, futureTick } from './-utils';
 
 // Ember internally tracks AJAX requests in the same way that we do here for
 // legacy style "acceptance" tests using the `ember-testing.js` asset provided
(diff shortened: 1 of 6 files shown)


rwjblue:
https://github.com/emberjs/ember-test-helpers/pull/269/files?w=1 makes this a bit easier to review, since most of the changes are indentation related (removed a level of nesting).

addon-test-support/@ember/test-helpers/settled.js
```diff
@@ -146,7 +143,7 @@ export default function settled(options) {
         timeout = MAX_TIMEOUT;
       }
 
-      global.setTimeout(function() {
+      nextTick(function() {
```
Turbo87: since this invocation is called with a non-0 timeout value does it make sense to call something called `nextTick` here?

rwjblue: I can export it with a different name. I agree it isn’t perfect. 

The basic gist that we need to do is ensure we use the eagerly cached `setTimeout` (which is what `nextTick` actually is).

Turbo87: sure, I understand why it's there and why it's used, I'm just not sure it is the best name. I like the name for the situation where you don't pass any timeout at all, but once you pass a timeout the name is not fitting anymore since it won't actually be the *next* tick 😉

rwjblue: Yep, totally agree. I’ll export it as `futureTick` also 😜 and use that here...

rwjblue: 👍 - Updated now.

PR #272, opened 2017-12-15
Make tests at least boot in IE and sort package.json alphabetically

cibernox:


vendor/monkey-patches.js
```diff
@@ -21,7 +21,7 @@
       // that happen _between_ acceptance tests will always share
       // `pendingRequests`.
       _Ember.Application.reopen({
-        willDestroy() {
+        willDestroy: function() {
```
Turbo87: with https://github.com/emberjs/ember-test-helpers/pull/271 this should not be necessary anymore

package.json
```diff
@@ -35,20 +35,21 @@
...
-    "ember-cli-test-loader": "^2.2.0",
     "ember-cli-shims": "^1.2.0",
+    "ember-cli-test-loader": "^2.2.0",
+    "ember-cli": "^2.17.1",
     "ember-data": "~2.17.0",
     "ember-debug-handlers-polyfill": "^1.0.3",
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-fetch": "^3.4.0",
     "ember-load-initializers": "^1.0.0",
+    "ember-maybe-import-regenerator": "^0.1.6",
```
Turbo87: we should probably use https://github.com/ember-cli/ember-maybe-import-regenerator-for-testing instead since the production code should not use async/await

package.json
```diff
@@ -35,20 +35,21 @@
...
     "ember-cli-eslint": "^4.2.1",
-    "ember-cli-htmlbars": "^2.0.1",
     "ember-cli-htmlbars-inline-precompile": "^1.0.0",
+    "ember-cli-htmlbars": "^2.0.1",
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-pretender": "^1.0.1",
-    "ember-cli-test-loader": "^2.2.0",
     "ember-cli-shims": "^1.2.0",
+    "ember-cli-test-loader": "^2.2.0",
+    "ember-cli": "^2.17.1",
```
Turbo87: I don't think that is the default order that npm and yarn use. I usually use https://github.com/keithamus/sort-package-json to sort this correctly.

package.json
rwjblue: Agreed, and just to be clear we do have linting setup to catch any `async` / `await` usage in `addon-test-support`...

tests/dummy/config/targets.js
```diff
@@ -1,4 +1,9 @@
 /* eslint-env node */
 module.exports = {
-  browsers: ['last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'],
+  browsers: [
+    'ie 11',
```
rwjblue: Lets do this conditionally? I really like debugging with native async/await, and we only need IE11 support in smaller cases.

I was thinking of something like: http://rwjblue.com/2017/10/30/async-await-configuration-adventure/#10limittargetsduringlocaldevelopmenttesting

PR #276, opened 2017-12-15
Fix almost all tests for ie11

cibernox:
I think there is still 2 failing tests, but I'm out for the day.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fill-in.js
+++ b/addon-test-support/@ember/test-helpers/dom/fill-in.js
@@ -22,8 +22,8 @@ export default function fillIn(target, text) {
     if (!element) {
       throw new Error(`Element not found when calling \`fillIn('${target}')\`.`);
     }
-
-    if (!isFormControl(element) && !element.isContentEditable) {
+    let isControl = isFormControl(element);
+    if (!isControl && !element.isContentEditable) {
       throw new Error('`fillIn` is only usable on form controls or contenteditable elements.');
     }
 
@@ -33,10 +33,10 @@ export default function fillIn(target, text) {
 
     __focus__(element);
 
-    if (element.isContentEditable) {
-      element.innerHTML = text;
-    } else {
+    if (isControl) {
       element.value = text;
+    } else {
+      element.innerHTML = text;
     }
 
     fireEvent(element, 'input');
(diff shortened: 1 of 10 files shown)


addon-test-support/@ember/test-helpers/dom/fill-in.js
```diff
@@ -22,8 +22,8 @@ export default function fillIn(target, text) {
     if (!element) {
       throw new Error(`Element not found when calling \`fillIn('${target}')\`.`);
     }
-
-    if (!isFormControl(element) && !element.isContentEditable) {
+    let isControl = isFormControl(element);
```
cibernox: It turns out that `isContentEditable` is true for `input` elements in IE. I did this fix but maybe we have to go back to check if the element has a `contenteditable=true` attribute

tests/dummy/config/targets.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-env node */
-const browsers = ['last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'];
+const browsers = ['ie 11', 'last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'];
```
rwjblue: We should probably revert this change (to test locally you can use `CI=true ember test --server` or `CI=true ember s -p 0`)...

tests/helpers/is-ie.js
```diff
@@ -0,0 +1 @@
+export default !window.ActiveXObject && 'ActiveXObject' in window;
```
rwjblue: This is great! 👍  I think we should add a comment explaining what it does.

Something like:

```js
// `window.ActiveXObject` returns undefined in IE11 (as well as non-IE browsers)
// whereas `"ActiveXObject" in window` returns `true` in all IE versions
// only IE11 will pass _both_ of these conditions
```

ro0gr: 😨

rwjblue:
OK, pushed some changes / tweaks to address my comments above.

IE11 PASSING

cibernox:
Now great. Now I need an IE10 vm

rwjblue:
Great work here @cibernox! Thank you!

PR #277, opened 2017-12-15
Fix tests when run under MS Edge.

rwjblue:


tests/unit/dom/blur-test.js
```diff
@@ -1,14 +1,17 @@
...
 
 let focusSteps = ['focus', 'focusin'];
 let blurSteps = ['blur', 'focusout'];
 
-if (isIE) {
+if (isIE11) {
   focusSteps = ['focusin', 'focus'];
   blurSteps = ['focusout', 'blur'];
+} else if (isEdge) {
+  focusSteps = ['focus', 'focusin'];
```
Turbo87: this matches the default, why did you overwrite it again?

tests/helpers/browser-detect.js
```diff
@@ -1,4 +1,7 @@
 // `window.ActiveXObject` returns undefined in IE11 (as well as non-IE browsers)
 // `"ActiveXObject" in window` returns `true` in all IE versions
 // only IE11 will pass _both_ of these conditions
-export default !window.ActiveXObject && 'ActiveXObject' in window;
+export const isIE11 = !window.ActiveXObject && 'ActiveXObject' in window;
```
Turbo87: should we explicitly check for `undefined` instead of falsyness?

rwjblue: I can try that, will confirm in a local VM...

tests/unit/dom/blur-test.js
rwjblue: stupidity, the only one I meant to override was the `blurSteps` (because thats the only difference).

tests/helpers/browser-detect.js
rwjblue: LOLOLOL, I tested changing this.  tldr IE11 makes no sense 🙃 

`window.ActiveXObject` is actually a function (not `undefined` or `false`), but `!!window.ActiveXObject` is false 😝 ...

PR #280, opened 2017-12-16
Cleanup browser detection and overrides.

rwjblue:
Mentioned in https://github.com/emberjs/ember-test-helpers/pull/277#pullrequestreview-83972554.


tests/helpers/browser-detect.js
```diff
@@ -1,4 +1,4 @@
-// `window.ActiveXObject` returns undefined in IE11 (as well as non-IE browsers)
+// `window.ActiveXObject` is "falsey" in IE11 (but not `undefined` or `false`)
```
Turbo87: what the ...!?

rwjblue: INDEED!

PR #282, opened 2017-12-20
Refactor `setupRenderingContext` so that element is stable after setup.

### LINKED ISSUE #281: this.element is not accessible until this.render() is settled
I try to test if the loading animation is rendered while the component is loading data.

The test looks something like:
```js
import { setupRenderingTest } from 'ember-qunit';
import { waitFor } from '@ember/test-helpers';
import hbs from 'htmlbars-inline-precompile';

module('Integration | Component | my-component', function(hooks) {
  setupRenderingTest(hooks);

  test('it renders spinner', async function(assert) {
    // Do not use `await` intentionally. 
    // Otherwise it waits until ajax request is finished 
    // which prevents me to test loader existence
    this.render(hbs`{{my-component
      itemId="1"
    }}`);

    await waitFor('.loading-box');

    assert.ok(1);
  });
});
```

When test execution reaches `waitFor` I get the following error:
> Must setup rendering context before attempting to interact with elements.

As far as I can see element is set only after rendering is complete:

https://github.com/emberjs/ember-test-helpers/blob/b0f037b7361388f051d0f32615fdedd2581ff4bf/addon-test-support/%40ember/test-helpers/setup-rendering-context.js#L139-L146

The first thing came into my mind is to move setting of element in the very beginning of rendering cause seems like it always points to the same container _'#ember-testing > .ember-view'_. However the comment regarding ember <= 2.4 makes me think it might be not the right solution.

rwjblue:
Previously (with `moduleForComponent`) eager rendering was avoided because it was not clear that rendering was _actually_ desired (and therefore the extra work would be wasted in cases that were not testing rendering).  Unlike `moduleForComponent` we can be certain that `setupRenderingContext` is only used in actual rendering contexts (its in the name :P ), so we do not need to be concerned with the performance.

This change is a compatible change (unless someone was _relying_ on the fact that `this.element` was undefined, which seems super weird) that ensures `this.element` is set immediately upon `setupRenderingContext` being called (instead of lazily when the first `render` is done). At that point `this.element` is essentially an empty element (roughly akin to having only `{{outlet}}` in an `application` template without an `index` template).

This unlocks the ability to test pending promises during rendering (via things like `waitUntil`).

Fixes #281

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.js
@@ -5,6 +5,7 @@ import global from './global';
 import { getContext } from './setup-context';
 import { nextTickPromise } from './-utils';
 import settled from './settled';
+import hbs from 'htmlbars-inline-precompile';
 
 export const RENDERING_CLEANUP = Object.create(null);
 
@@ -57,14 +58,17 @@ export default function(context) {
     let OutletView = owner.factoryFor
       ? owner.factoryFor('view:-outlet')
       : owner._lookupFactory('view:-outlet');
-    let OutletTemplate = owner.lookup('template:-outlet');
     let toplevelView = OutletView.create();
+    let OutletTemplate = owner.lookup('template:-outlet');
+    if (!OutletTemplate) {
+      owner.register('template:-outlet', hbs`{{outlet}}`);
+      OutletTemplate = owner.lookup('template:-outlet');
+    }
 
     // push this into the rendering specific cleanup bucket, to be ran during
     // `teardownRenderingContext` but before the owner itself is destroyed
     RENDERING_CLEANUP[contextGuid].push(() => toplevelView.destroy());
 
-    let hasOutletTemplate = Boolean(OutletTemplate);
     let outletState = {
       render: {
         owner,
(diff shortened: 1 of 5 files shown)


rwjblue:
@ro0gr - Can you review the test I added (in the second commit) to make sure it roughly emulates your scenario?

/cc @cibernox (I believe you mentioned this case the other day also)

addon-test-support/@ember/test-helpers/setup-rendering-context.js
```diff
@@ -78,95 +82,66 @@ export default function(context) {
...
+        return settled();
+      });
+    };
 
     if (global.jQuery) {
       context.$ = function $(selector) {
         // emulates Ember internal behavor of `this.$` in a component
         // https://github.com/emberjs/ember.js/blob/v2.5.1/packages/ember-views/lib/views/states/has_element.js#L18
-        return selector ? global.jQuery(selector, element) : global.jQuery(element);
+        return selector ? global.jQuery(selector, context.element) : global.jQuery(context.element);
```
cibernox: Unrelated, but I just learned that `context.element` is a jQuery element if you have jQuery. Is that intended?
I'd have expeted user to do the wrapping themselves if they want to, like

```
import $ from 'jquery';
// ...
  assert.equal($(this.element).hasClass('foo'));
```

rwjblue:
> it even seems to simplify the rendering logic quite a bit :)

Totally agreed. A number of conditionals removed...

addon-test-support/@ember/test-helpers/setup-rendering-context.js
Turbo87: @cibernox `this.$ = $(this.element)`, `this.element` is always a native DOM element

tests/integration/setup-rendering-context-test.js
```diff
@@ -0,0 +1,92 @@
...
+    let deferred = defer();
+    this.set('promise', deferred.promise);
+
+    // force the waiter to pause, emulating a pending AJAX or fetch request
+    this.isWaiterPending = true;
+
+    // Does not use `await` intentionally
+    let renderPromise = render(hbs`{{promise-wrapper promise=promise }}`);
+
+    await waitFor('.loading');
```
cibernox: This is exactly what I was looking for two days ago 🎉

addon-test-support/@ember/test-helpers/setup-rendering-context.js
rwjblue: @cibernox - That is very surprising indeed! As you can see (just above in the diff) this is how we set `context.element` now (was the same before but done via `Object.defineProperty`):

```js
    context.element = document.querySelector('#ember-testing > .ember-view');
```

How could that _ever_ be a jQuery element? Are you certain that this wasn't in a `moduleForComponent` test?

cibernox: Nevermind. I read this ternary completely wrong. I thought the line was setting `context.element` to `global.jQuery(context.element)`.

ro0gr:
@rwjblue yes, it does.

Issue #285, opened 2017-12-23
Documentation!


rwjblue:
We need to generate a nice set of detailed API docs for the functionality in this library. 

- [ ] Add in-line API docs for all public APIs in `addon-test-support/@ember/test-helpers`
- [x] Add API doc generator (likely publishing to GH Pages).

Turbo87:
@samselikoff would https://github.com/ember-learn/ember-cli-addon-docs already be an option?

rwjblue:
Landed docs for the DOM interaction helpers in https://github.com/emberjs/ember-test-helpers/pull/287.

samselikoff:
@Turbo87 yes, it is functional! We are trying to round it out but if you/someone wants to use it now I will be happy to discuss/pair with you.

You can look at https://ember-learn.github.io/ember-cli-addon-docs/ for reference (and the corresponding dummy app in the source code to see how its built)

PR #286, opened 2017-12-27
Add `documentation` dev dependency

Turbo87:
bildschirmfoto 2017-12-27 um 12 05 26

Resolves part of https://github.com/emberjs/ember-test-helpers/issues/285

I haven't tested the CI part yet, but in theory it should be configured so that it automatically builds and deploys the documentation to GitHub Pages whenever a tag is pushed.

Unfortunately http://emberjs.github.io/ember-test-helpers/ is currently directing to a emberjs.com 404 page so I'm not sure if we can actually use GitHub Pages at all 🤔 

/cc @locks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,16 @@ jobs:
         on:
           tags: true
           repo: emberjs/ember-test-helpers
+    - if: tag IS present
+      script: yarn documentation
+      deploy:
+        provider: pages
+        skip_cleanup: true
+        local_dir: docs
+        github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
+        on:
+          tags: true
+          repo: emberjs/ember-test-helpers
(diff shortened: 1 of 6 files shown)


rwjblue:
Thank you!

.travis.yml
```diff
@@ -58,6 +58,16 @@ jobs:
         on:
           tags: true
           repo: emberjs/ember-test-helpers
+    - if: tag IS present
+      script: yarn documentation
+      deploy:
+        provider: pages
+        skip_cleanup: true
+        local_dir: docs
+        github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
```
Turbo87: @rwjblue forgot to mention: this env var needs to be set by someone with sufficient privs (aka. not me 🤷‍♂️)

PR #287, opened 2017-12-27
Flesh out DOM interaction helper documentation.

rwjblue:


addon-test-support/@ember/test-helpers/dom/click.js
```diff
@@ -22,10 +17,12 @@ export function __click__(element) {
 }
 
 /**
-  @method click
-  @param {String|Element} target
-  @return {Promise<void>}
+  Clicks on the specified selector.
```
Turbo87: "specified target" might be better

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -15,15 +15,6 @@ const MOUSE_EVENT_TYPES = [
...
 
-/**
-  @method fireEvent
-  @param {Element} element
-  @param {String} type
-  @param {Object} [options]
-  @returns {Event}
-
-  @private
-*/
```
Turbo87: Why did you remove these? They might still be useful for type info

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -13,16 +13,18 @@ const DEFAULT_MODIFIERS = Object.freeze({
 });
 
 /**
+  Triggers a keyboad event on the specified selector.
```
Turbo87: Keyboard

addon-test-support/@ember/test-helpers/dom/fire-event.js
rwjblue: Hmm, unless we are going to type check ember-test-helpers internally, I don't think this provides any value. This entire file is private and only used to expose the actual public API's which _do_ have valid type information...

Turbo87: Some IDEs are able to parse these comments and provide hints. Might be useful for future contributors.

rwjblue: OK, will bring it back for the export (but not the inner / private helper functions)

PR #290, opened 2017-12-29
Tests: Make valid-jsdoc an error (instead of warning)

rwjblue:
Now that the documentation is updated, this makes it an error...

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,7 @@ module.exports = {
     {
       files: ['addon-test-support/**/*.js'],
       rules: {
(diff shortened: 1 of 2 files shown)


Turbo87:
FWIW `documentation` also has a `lint` command 🤔

rwjblue:
Hmm, what does it do?

rwjblue:
I'll add a lint script for it and add it to the initial linting build...

Turbo87:
I think it does something similar, just not sure why they recommend the ESLint rule in their docs... using both can't hurt though

PR #292, opened 2017-12-30
Refactor settled to leverage waitUntil.

rwjblue:
Migrates ramp up functionality to `waitUntil` and migrates `settled` to leverage `waitUntil`.

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -1,10 +1,10 @@
 import { run } from '@ember/runloop';
 
-import { Promise as EmberPromise } from 'rsvp';
 import jQuery from 'jquery';
 
 import Ember from 'ember';
-import { nextTick, futureTick } from './-utils';
+import { nextTick } from './-utils';
+import waitUntil from './wait-until';
 
 // Ember internally tracks AJAX requests in the same way that we do here for
 // legacy style "acceptance" tests using the `ember-testing.js` asset provided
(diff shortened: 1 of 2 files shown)


rwjblue:
Note: I didn't update documentation mostly because exactly _how_ we wait seems like an implementation detail and not really the actual API...

addon-test-support/@ember/test-helpers/settled.js
```diff
@@ -203,25 +200,10 @@ const MAX_TIMEOUT = 10;
...
-          scheduleCheck(counter + 1);
-        }
-      }, timeout);
-    }
-
-    scheduleCheck(0);
-  });
+  return waitUntil(
+    () => {
+      return isSettled(options);
```
Turbo87: this could use shorthand arrow function syntax to simplify the code

addon-test-support/@ember/test-helpers/wait-until.js
```diff
@@ -1,6 +1,8 @@
 import { Promise } from 'rsvp';
+import { futureTick } from './-utils';
 
-import { nextTick } from './-utils';
+const TIMEOUTS = [0, 1, 2, 5, 7];
```
Turbo87: 7?

rwjblue: Yeah, I added another entry. Does It seem bad?

addon-test-support/@ember/test-helpers/settled.js
rwjblue: Yep, sounds good. Will update.

PR #294, opened 2017-12-30
Update documentation to generate into README.

rwjblue:
* update docs script to inject into README
* remove old / outdated docs in README
* remove gh-pages auto-deployment

[Rendered README.md](https://github.com/rwjblue/ember-test-helpers/blob/interaction-helpers-root/README.md) for easier reviewing...


README.md
```diff
@@ -4,60 +4,583 @@ A test-framework-agnostic set of helpers for testing Ember.js applications.
...
+-   `mousedown`
+-   `focus`
+-   `focusin`
+-   `mouseup`
+-   `click`
+
+The exact listing of events that are triggered may change over time as needed
+to continue to emulate how actual browsers handle clicking a given element.
+
+**Parameters**
```
Turbo87: On mobile this renders the same as the h4 of the functions which makes it somewhat hard to read and find stuff 🤔

README.md
```diff
@@ -4,60 +4,583 @@ A test-framework-agnostic set of helpers for testing Ember.js applications.
...
 ## Collaborating
 
 ### Installation
 
-* `git clone <repository-url>` this repository
-* `cd ember-test-helpers`
-* `yarn install`
+-   `git clone <repository-url>` this repository
+-   `cd ember-test-helpers`
+-   `yarn install`
```
Turbo87: IMHO it's quite annoying that documentation is reformatting the whole file when it's only supposed to replace a single section. I've had the same issue in qunit-dom and decided to have it output in a API.md instead

documentation.yml
```diff
@@ -1,50 +1,50 @@
...
-  - triggerKeyEvent
-  - fillIn
+    children:
+      - click
+      - tap
+      - focus
+      - blur
+      - triggerEvent
+      - triggerKeyEvent
+      - fillIn
```
Turbo87: I tried this at first, but it looked strange in the HTML version. Is this still the case?

rwjblue: Yes, this _only_ works for the `md` version. If you do `children:` in the HTML formatted output its _really really_ weird.

README.md
rwjblue: Ya, agreed that its odd.

README.md
rwjblue: Yeah, I think that is also true on desktop.

rwjblue:
FWIW, I don't think this is a "final" solution. I just spent so much damned time on the docs, and I'd like to have _something_ to show for it while we figure out a deployment strat for the HTML formatted output...

Turbo87:
aye, I think I'd still prefer an API.md though 🤔

PR #295, opened 2017-12-30
Use "real" rootElement for DOM interaction helpers.

rwjblue:
Extracts a `getRootElement` utility function (used by `getElement` and `waitFor`).  The utility is private for now, but may be made public in the future (needs more docs at the least).

The basic gist is that `getRootElement` should search relative to the `rootElement` of the application (when `setApplication` was used), and _not_ scoped to the first `.ember-view` _inside_ that element. This ensures that elements added to the `rootElement` (e.g. via `ember-wormhole` / `liquid-wormhole`) can also be interacted with.

Also updates `setupRenderingContext` to use the applications designated `rootElement` (as opposed to _always_ using `#ember-testing`). This brings `setupRenderingContext` in line with `setupApplicationContext`, and now properly uses the specified `rootElement`.

Prior to this, when using `setApplication(...)` with a different `rootElement` set than `#ember-testing` the event dispatcher would be setup to listen on a different element than the one that is actually being rendered into...

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.js
@@ -111,8 +111,15 @@ export default function setupRenderingContext(context) {
     };
     toplevelView.setOutletState(outletState);
 
-    // TODO: make this id configurable
-    run(toplevelView, 'appendTo', '#ember-testing');
+    let rootElement;
+    // When the host app uses `setApplication` (instead of `setResolver`) the owner has
+    // a `rootElement` set on it with the element id to be used
+    if (owner._emberTestHelpersMockOwner) {
+      rootElement = '#ember-testing';
+    } else {
+      rootElement = owner.rootElement;
+    }
+    run(toplevelView, 'appendTo', rootElement);
 
     // ensure the element is based on the wrapping toplevel view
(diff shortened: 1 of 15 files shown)


addon-test-support/@ember/test-helpers/setup-rendering-context.js
```diff
@@ -111,8 +111,15 @@ export default function setupRenderingContext(context) {
...
-    // TODO: make this id configurable
-    run(toplevelView, 'appendTo', '#ember-testing');
+    let rootElement;
+    // When the host app uses `setApplication` (instead of `setResolver`) the owner has
+    // a `rootElement` set on it with the element id to be used
+    if (owner._emberTestHelpersMockOwner) {
+      rootElement = '#ember-testing';
+    } else {
+      rootElement = owner.rootElement;
+    }
```
Turbo87: Why is this not using the extracted helper function?

rwjblue: good point, I'll double check that all supported ember versions support calling `appendTo` with an actual element (as opposed to a selector string, which is being used here), but I totally agree that it makes sense to avoid double encoding this concept (and is why I removed the complete hard coding to begin with).

PR #296, opened 2017-12-31
Clean up and simplify setupRenderingContext

rwjblue:
Moves core functionality into the importable `render` / `clearRender` / etc utility functions and out of the `setupRenderingContext` directly.

This simplifies how the methods work together (since they no longer have access to _everything_ in closure scope), and sets the stage for an eventual deprecation of `this.render` and `this.clearRender`.

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.js
@@ -9,7 +9,38 @@ import hbs from 'htmlbars-inline-precompile';
 import getRootElement from './dom/get-root-element';
 
 export const RENDERING_CLEANUP = Object.create(null);
+const OUTLET_TEMPLATE = hbs`{{outlet}}`;
+const EMPTY_TEMPLATE = hbs``;
 
+/**
+  @private
+  @param {Ember.ApplicationInstance} owner the current owner instance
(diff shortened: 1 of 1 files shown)


addon-test-support/@ember/test-helpers/setup-rendering-context.js
```diff
@@ -70,123 +165,58 @@ export default function setupRenderingContext(context) {
...
+    })
+    .then(() => {
+      // ensure the element is based on the wrapping toplevel view
+      // Ember still wraps the main application template with a
+      // normal tagged view
+      //
+      // In older Ember versions (2.4) the element itself is not stable,
+      // and therefore we cannot update the `this.element` until after the
+      // rendering is completed
+      context.element = document.querySelector('#ember-testing > .ember-view');
```
Turbo87: I assume this will break once the RFC from @cibernox is implemented?

cibernox: It probably will 🤔

rwjblue: It just changes to use the rootElement directly (based on a feature detect):

```
context.element = getRootElement();
```

At which point the various interaction helpers (which were just changed to use `rootElement` instead of this) _and_ `this.element` will be back in sync...

Issue #300, opened 2018-01-15
Arbitrary arguments to `wait` cause error


simonihmig:
Seems since this [commit](https://github.com/emberjs/ember-test-helpers/commit/0a83ee3b822f59813baf71bc50938cc71e70f1b4) where `wait` delegates to `settled`, passing a non-object (e.g. boolean) to `wait` would cause a `Cannot use 'in' operator to search for 'waitForTimers' in true` error. Prior to this, this would work (the argument would simply have no effect).

Offending line: https://github.com/emberjs/ember-test-helpers/blob/9eb3e32f8340d01704e950f317b282f07073f1e9/addon-test-support/%40ember/test-helpers/settled.js#L171
Example: https://travis-ci.org/kybishop/ember-popper/jobs/328856339#L547
Caused by: https://github.com/kybishop/ember-popper/blob/5d4457588fd9f30f7a0a2aa34a28e86209ae5978/tests/integration/components/ember-popper/action-test.js#L46

rwjblue:
Great sleuthing. There is another reported issue around this same sort of thing (though I didn’t know it was related until you tracked down the problem). 

The fix here should be to check the arg is an object and not-null before checking for options...

PR #303, opened 2018-01-18
Assert `settled()` only takes an object as first argument

### LINKED ISSUE #283: Unsafe options usage in isSettled
This is definitely a fault on my end, but something I think is worth bringing up, at least for anyone else who may encounter this.

We have been using the library for years, however a few of our tests were using it incorrectly. This wasn't a problem in the past as it was a bit safer.

Typical usage
```
...
return wait().then(()=>{
    assert....
})
```

Which is correct usage, however a few of my tests used it incorrectly, for example `wait(3000)` thinking it would also wait 3 seconds. This wasn't a problem in the past as the code in question used `hasOwnProperty` which was safe on a variable which was a number. Now it uses the `in` keyword which breaks on non objects.

We would see errors such as `Uncaught TypeError: Cannot use 'in' operator to search for 'waitForTimers' in 500`

[how it was](https://github.com/emberjs/ember-test-helpers/blob/v0-6-x/lib/ember-test-helpers/wait.js#L55-L57)
[how it is now](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/settled.js#L112-L116)

Easiest fix for users: remove the number, and better understand how/what wait is actually doing
Easiest fix for library: explicity check options is an object, and not just undefined

snewcomer:
Fixes #283

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -202,8 +202,10 @@ export default function settled() {
   let options = arguments[0];
 
   if (arguments.length > 0) {
-    assert(`The 'settled' (formerly 'wait') testing helper only accepts an object. You passed ${options}`,
-      typeof options === 'object' && options !== null);
+    assert(
+      `The 'settled' (formerly 'wait') testing helper only accepts an object.  You passed ${options}`,
+      typeof options === 'object' && options !== null
+    );
   }
 
   return waitUntil(() => isSettled(options), { timeout: Infinity });
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -203,7 +203,7 @@ export default function settled() {
 
   if (options) {
     assert(
-      `The 'settled' (formerly 'wait') testing helper only accepts an object. You passed ${options}`,
+      `The 'settled' (formerly 'wait') testing helper only accepts an object if passed arguments. You passed ${options}`,
       typeof options === 'object'
     );
   }
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -201,10 +201,10 @@ export function isSettled() {
 export default function settled() {
   let options = arguments[0];
 
-  if (arguments.length > 0) {
+  if (options) {
     assert(
       `The 'settled' (formerly 'wait') testing helper only accepts an object. You passed ${options}`,
-      typeof options === 'object' && options !== null
+      typeof options === 'object'
     );
   }
 
--- a/addon-test-support/@ember/test-helpers/settled.js
+++ b/addon-test-support/@ember/test-helpers/settled.js
@@ -201,8 +201,10 @@ export function isSettled() {
 export default function settled() {
   let options = arguments[0];
 
-  assert(`The 'settled' (formerly 'wait') testing helper only accepts 'undefined' or an object. You passed ${options}`,
-    typeof options === 'undefined' || typeof options === 'object' && options !== null);
+  if (arguments.length > 0) {
+    assert(`The 'settled' (formerly 'wait') testing helper only accepts an object. You passed ${options}`,
+      typeof options === 'object' && options !== null);
+  }
 
   return waitUntil(() => isSettled(options), { timeout: Infinity });
 }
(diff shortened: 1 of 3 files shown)


rwjblue:
Also addresses https://github.com/emberjs/ember-test-helpers/issues/300. 

After thinking about this more, let’s just make this work without error for now. In the long run, I want to deprecate the options argument completely and making this an assertion now then removing it later feels odd to me...

snewcomer:
That makes sense.  So does that mean close the PR?  Or just `debug`/`deprecate`?

addon-test-support/@ember/test-helpers/settled.js
```diff
@@ -200,5 +201,8 @@ export function isSettled() {
 export default function settled() {
   let options = arguments[0];
 
+  assert(`The 'settled' (formerly 'wait') testing helper only accepts 'undefined' or an object. You passed ${options}`,
+    typeof options === 'undefined' || typeof options === 'object' && options !== null);
```
Turbo87: why is `undefined` a valid value? `settled()` and `settled(undefined)` are somewhat similar but I don't think we should encourage passing `undefined` here.

rwjblue: Because that’s what the value appears to be for `settled()`. I don’t think _actually_ passing undefined is useful in anyway.

rwjblue:
@snewcomer - Ya, if you are interested in updating that would be great. Basically we should avoid the `in` checks completely if the first arg is not `typeof obj === ‘object’ && obj !== null`...

addon-test-support/@ember/test-helpers/settled.js
Turbo87: yep, that is why I don't think we should mention it that way in the assertion message. also we could check if `arguments.length > 0`, because IMHO `settled(undefined)` should also throw the assertion.

rwjblue: Oh I see. Sure, that makes sense.

rwjblue:
Another option here that is growing on me, is to treat the change in `wait` behavior as a regression but keep `settled` with the assertion that you originally submitted in this PR (with the tweaks that @turbo87 mentioned). In order to do this, `wait` changes from a simple reexport to something that sanitizes the input....

Thoughts?

snewcomer:
Ah great points!  The line below had to be refactored b/c as a callback w/ `apply`, undefined would be the argument.

https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/setup-application-context.js#L24

And I like the idea of changing the wait helper to sanitize the input.  Would it just simply ignore the passed in `3000` for example?

snewcomer:
Pinging this again.  The test is actually passing I believe.

screen shot 2018-01-31 at 7 16 31 am

kybishop:
@snewcomer correct! The linter is failing, but the tests themselves are good: https://travis-ci.org/emberjs/ember-test-helpers/jobs/333478168#L500

You should be able to test fixes with `yarn lint`

snewcomer:
@rwjblue pinging ya on this one if you still think is relevant :)

addon-test-support/@ember/test-helpers/setup-application-context.js
```diff
@@ -21,7 +21,9 @@ export function visit() {
     .then(() => {
       context.element = document.querySelector('#ember-testing > .ember-view');
     })
-    .then(settled);
+    .then(() => {
+      return settled();
+    });
```
Turbo87: you can still use short form: `.then(() => settled());`

addon-test-support/@ember/test-helpers/settled.js
```diff
@@ -198,10 +198,8 @@ export function isSettled() {
   @public
   @returns {Promise<void>} resolves when settled
 */
-export default function settled() {
-  let options = arguments[0];
-
-  if (arguments.length > 0) {
+export default function settled(options) {
```
Turbo87: this was `documentation` will see that there is a parameter and document it as such, but as mentioned in https://github.com/emberjs/ember-test-helpers/issues/314 this parameter is kinda private so we should keep `let options = arguments[0];` instead.

snewcomer: @Turbo87 Another good point.  How about now 😆 

https://github.com/emberjs/ember-test-helpers/pull/303/commits/22d1538396d5bcbb1ddb685d969179fc05fd01c9

rwjblue:
I took a stab at an alternative strategy in https://github.com/emberjs/ember-test-helpers/pull/317, what do y'all think?

PR #305, opened 2018-01-21
Support testing without an application template wrapper.

cibernox:
Is this the best way to check if a feature flag is enabled?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,7 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
     - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
+    - env: EMBER_TRY_SCENARIO=ember-canary-without-application-wrapper
 
     # runs deploy if running on a specific tag
     - stage: deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
     - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
-    - env: EMBER_TRY_SCENARIO=ember-canary-without-application-wrapper
+    - env: EMBER_TRY_SCENARIO=ember-without-application-wrapper
 
     # runs deploy if running on a specific tag
     - stage: deploy
--- a/addon-test-support/@ember/test-helpers/setup-application-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-application-context.js
@@ -19,7 +19,12 @@ export function visit() {
       return owner.visit(...arguments);
     })
     .then(() => {
-      context.element = document.querySelector('#ember-testing > .ember-view');
+      // eslint-disable-next-line
+      if (EmberENV.FEATURES['ember-glimmer-remove-application-template-wrapper']) {
+        context.element = document.querySelector('#ember-testing');
+      } else {
+        context.element = document.querySelector('#ember-testing > .ember-view');
+      }
     })
     .then(settled);
 }
--- a/addon-test-support/@ember/test-helpers/setup-application-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-application-context.js
@@ -20,10 +20,10 @@ export function visit() {
     })
     .then(() => {
       // eslint-disable-next-line
-      if (EmberENV.FEATURES['ember-glimmer-remove-application-template-wrapper']) {
-        context.element = document.querySelector('#ember-testing');
-      } else {
+      if (EmberENV._APPLICATION_TEMPLATE_WRAPPER) {
         context.element = document.querySelector('#ember-testing > .ember-view');
+      } else {
+        context.element = document.querySelector('#ember-testing');
       }
     })
     .then(settled);
--- a/addon-test-support/@ember/test-helpers/setup-application-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-application-context.js
@@ -20,7 +20,7 @@ export function visit() {
     })
     .then(() => {
       // eslint-disable-next-line
-      if (EmberENV._APPLICATION_TEMPLATE_WRAPPER) {
+      if (EmberENV._APPLICATION_TEMPLATE_WRAPPER !== false) {
         context.element = document.querySelector('#ember-testing > .ember-view');
       } else {
         context.element = document.querySelector('#ember-testing');
--- a/addon-test-support/@ember/test-helpers/setup-application-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-application-context.js
@@ -1,3 +1,4 @@
+/* globals EmberENV */
 import { get } from '@ember/object';
 import { nextTickPromise } from './-utils';
 import { getContext } from './setup-context';
@@ -19,7 +20,6 @@ export function visit() {
       return owner.visit(...arguments);
     })
     .then(() => {
-      // eslint-disable-next-line
       if (EmberENV._APPLICATION_TEMPLATE_WRAPPER !== false) {
         context.element = document.querySelector('#ember-testing > .ember-view');
       } else {
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.js
@@ -197,10 +197,10 @@ export default function setupRenderingContext(context) {
       // and therefore we cannot update the `this.element` until after the
       // rendering is completed
       // eslint-disable-next-line
-      if (EmberENV.FEATURES['ember-glimmer-remove-application-template-wrapper']) {
-        context.element = getRootElement();
-      } else {
+      if (EmberENV._APPLICATION_TEMPLATE_WRAPPER) {
         context.element = getRootElement().querySelector('.ember-view');
+      } else {
+        context.element = getRootElement();
       }
 
       return context;
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.js
@@ -1,3 +1,4 @@
+/* globals EmberENV */
 import { guidFor } from '@ember/object/internals';
 import { run } from '@ember/runloop';
 import Ember from 'ember';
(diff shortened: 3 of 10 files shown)


addon-test-support/@ember/test-helpers/setup-application-context.js
```diff
@@ -19,7 +19,12 @@ export function visit() {
       return owner.visit(...arguments);
     })
     .then(() => {
-      context.element = document.querySelector('#ember-testing > .ember-view');
+      // eslint-disable-next-line
+      if (EmberENV.FEATURES['ember-glimmer-remove-application-template-wrapper']) {
```
rwjblue: This might work for now, but isn't a good long term solution (once it moves to beta this will be falsey).

I think what we need to do is render an empty template, detect if there are any `ember-view`s and then memoize (so we only have to do the extra rendering one time for the entire suite)/

rwjblue: @cibernox - Now that this has settled, we can use `EmberENV._APPLICATION_TEMPLATE_WRAPPER`. A `true` value means we need `#ember-testing > .ember-view` and a `false` value means we need `#ember-testing`.

addon-test-support/@ember/test-helpers/setup-rendering-context.js
```diff
@@ -196,7 +196,12 @@ export default function setupRenderingContext(context) {
       // In older Ember versions (2.4) the element itself is not stable,
       // and therefore we cannot update the `this.element` until after the
       // rendering is completed
-      context.element = getRootElement().querySelector('.ember-view');
+      // eslint-disable-next-line
+      if (EmberENV.FEATURES['ember-glimmer-remove-application-template-wrapper']) {
```
rwjblue: See above, should be able to use `EmberENV._APPLICATION_TEMPLATE_WRAPPER` here.

tests/dummy/config/environment.js
```diff
@@ -11,6 +11,7 @@ module.exports = function(environment) {
       FEATURES: {
         // Here you can enable experimental features on an ember canary build
         // e.g. 'with-controller': true
+        'ember-glimmer-remove-application-template-wrapper': true,
```
rwjblue: Lets remove this.

In order to test (until https://github.com/ember-cli/ember-try/pull/181 lands in a release) we will need to:

1. add an ember-try scenario for `ember-canary-without-application-wrapper`
2. add that scenario to .travis.yml
3. update `tests/dummy/config/environment.js` to do:

```js
EmberENV: {
  // ...snip other normal stuff...
  _APPLICATION_TEMPLATE_WRAPPER: process.env.EMBER_TRY_CURRENT_SCENARIO === 'ember-canary-without-application-wrapper' ? false : true
}
```

addon-test-support/@ember/test-helpers/setup-application-context.js
cibernox: I'll update it.

cibernox:
@rwjblue I think this is ready.
However, how some the feature flag has become an underscored prop in `EmberENV`. Is this FF still enabled the same way?

rwjblue:
> However, how some the feature flag has become an underscored prop in EmberENV. Is this FF still enabled the same way?

It is expected to be enabled via the `@ember/optional-features` addon which takes care of setting up the private environment flag. In the long run, all of this will be moved to be build time (so we don't ship both sets of implementation in Ember) but for now it is runtime (based on `EmberENV._APPLICATION_TEMPLATE_WRAPPER`).  

I'll update this addon in a followup PR to leverage `@ember/optional-features` and `ember-try`'s integration...

cibernox:
@rwjblue Do the failure make sense to you?

rwjblue:
Yes, I am working on fixing the canary failure in https://github.com/emberjs/ember.js/pull/16232. 

Also, I pushed some commits which cleanup the testing harness a bit (to more of what we expect "normal" addons to do going forward).

rwjblue:
ZOMG, finally green.

cibernox:
Looks great. :shipit:

.travis.yml
```diff
@@ -44,6 +44,7 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
     - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
+    - env: EMBER_TRY_SCENARIO=ember-canary-without-application-wrapper
```
Turbo87: can we drop the `canary` from the scenario name? I assume this functionality will eventually exist outside of canary and that currently it uses canary is only an implementation detail, right?

addon-test-support/@ember/test-helpers/setup-application-context.js
```diff
@@ -19,7 +19,12 @@ export function visit() {
       return owner.visit(...arguments);
     })
     .then(() => {
-      context.element = document.querySelector('#ember-testing > .ember-view');
+      // eslint-disable-next-line
```
Turbo87: I would prefer `/* global EmberENV */` because `eslint-disable-next-line` disables **all** linting rules

addon-test-support/@ember/test-helpers/setup-rendering-context.js
```diff
@@ -196,7 +196,12 @@ export default function setupRenderingContext(context) {
       // In older Ember versions (2.4) the element itself is not stable,
       // and therefore we cannot update the `this.element` until after the
       // rendering is completed
-      context.element = getRootElement().querySelector('.ember-view');
+      // eslint-disable-next-line
```
Turbo87: see above

package.json
```diff
@@ -54,9 +55,9 @@
     "ember-maybe-import-regenerator-for-testing": "^1.0.0",
     "ember-native-dom-event-dispatcher": "^0.6.3",
     "ember-resolver": "^4.0.0",
-    "ember-source": "~2.18.0",
+    "ember-source": "~3.0.0-beta.5",
     "ember-source-channel-url": "^1.0.1",
-    "ember-try": "^0.2.23",
+    "ember-try": "^1.0.0-beta.1",
```
Turbo87: offtopic: mmhhh.... I have thoughts about promoting this to 1.0.0...

tests/integration/setup-rendering-context-test.js
```diff
@@ -119,18 +119,10 @@ module('setupRenderingContext "real world"', function(hooks) {
...
-        this.element,
-        'precond - confirm that the rootElement is different from this.element'
-      );
-
       this.set('rootElement', rootElement);
 
       await render(hbs`{{#-in-element rootElement}}{{click-me-button}}{{/-in-element}}`);
 
-      // this will need to be modified / removed once RFC280 lands
-      assert.equal(this.element.textContent, '', 'no content is contained _within_ this.element');
```
Turbo87: why is this removed?

tests/integration/setup-rendering-context-test.js
```diff
@@ -119,18 +119,10 @@ module('setupRenderingContext "real world"', function(hooks) {
     test('can click on a sibling element of this.element (within the rootElement)', async function(assert) {
       let rootElement = document.getElementById('ember-testing');
 
-      assert.notEqual(
-        rootElement,
-        this.element,
-        'precond - confirm that the rootElement is different from this.element'
-      );
```
Turbo87: should we keep this assertion and run it conditionally instead?

package.json
rwjblue: haha, we can always make a 2.0 😸 If there are specific things wrong with ember-try, please make issues so we can discuss the details.

IMHO, there is basically no reason to not have the ability to signal "new API's are added" (normal minor version bump) from "bug fixes" (normal patch release).

tests/integration/setup-rendering-context-test.js
rwjblue: Because `this.element` (when the optional flag is enabled) _will_ have content. I can swap things around a bit and guard the assertions, but I think I'd rather have two tests instead.  One for "with a wrapper div" and one "without a wrapper"...

tests/integration/setup-rendering-context-test.js
rwjblue: ya, I'll make it two tests

package.json
Turbo87: https://github.com/ember-cli/ember-try/issues/155

tests/integration/setup-rendering-context-test.js
Turbo87: what content will be in there and why wasn't it in there before? I'm not sure I understand the assertion 🤔

rwjblue: The test (as it exists on master) is actually testing a few different things:

* `this.element` within the test context is **not** the `#ember-testing` div
* when rendering _only_ content that is "wormholed" (via `in-element` in this case), the actual `this.element` is still empty (because the wormholed content is somewhere else)
* that we _can_ interact (e.g. click) with elements that are within the `#ember-testing` div (even though they are above the `this.element`).

As you can tell some of these assertions are no longer needed with the changes from emberjs/rfcs#280.  Specifically:

> `this.element` within the test context is **not** the `#ember-testing` div

without the wrapper DIV, the `this.element` actually _is_ `#ember-testing`

>  when rendering _only_ content that is "wormholed", the actual `this.element` is still empty

due to the above, this is not true any longer. `this.element` will have content (the wormholed content)

These two changes are the assertions that were being removed (when you commented). I have changed things around a bit so that we have two different tests (one with an application template wrapper and one without it).

rwjblue:
Created https://github.com/emberjs/ember-optional-features/issues/10 to track swapping out the manual `EmberENV` checks in favor of something more reliable provided by `@ember/optional-features`.

tests/integration/setup-rendering-context-test.js
Turbo87: > that we can interact (e.g. click) with elements that are within the #ember-testing div (even though they are above the this.element).

given that we use `document.querySelector('#ember-testing > .ember-view')` how is `click()` able to interact with elements above `this.element`? last time I've tried it didn't work with wormholed content (yet) 🤔

rwjblue: This behavior was changed in https://github.com/emberjs/ember-test-helpers/pull/295. `click` and friends are always operating on the root element (`#ember-testing` generally speaking) which is **not** `this.element`.  The specific test that we are commenting on ATM is actually introduce _in_ that PR specifically to address the wormholed content scenario...

Turbo87: I see. Turns out we are still on v0.7.10 of `@ember/test-helpers` (though newest `ember-cli-qunit`) which is the reason why it didn't work when I tried it.

PR #307, opened 2018-01-25
Make sure initializers can modify the DOM

bendemboski:
The initializers were running after the snapshot of the root element was taken, meaning that if an initializer modified the contents of the root element, that modification would not persist beyond the first test. So we boot the app a little earlier to make sure initializers can safely modify the DOM.


rwjblue:
Thank you!

bendemboski:
You bet! I'm really liking this new test API, and happy to help out with it when I can!

PR #313, opened 2018-02-08
[FEATURE] Add ancestor for find and findAll

Serabe:
In ember-native-dom-helpers, find and findAll had an optional second
argument, a _context_ element for querySelector or querySelectorAll.

Given _context_ and _root element_ have already some meaning in
ember-test-helpers, I chose to call it _ancestor_ to prevent
misunderstoods.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/-get-element.js
+++ b/addon-test-support/@ember/test-helpers/dom/-get-element.js
@@ -5,9 +5,10 @@ import getRootElement from './get-root-element';
 
   @private
   @param {string|Element} target the element or selector to retrieve
+  @param {Element} [ancestor] optional root element
   @returns {Element} the target or selector
 */
-export default function getElement(target) {
+export default function getElement(target, ancestor) {
   if (
     target.nodeType === Node.ELEMENT_NODE ||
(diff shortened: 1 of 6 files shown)


rwjblue:
I don’t think we aim to be perfectly compatible, and the second argument seems superfluous to me...

Serabe:
The selector thing can get really complicated really fast.

```js
// Without ancestor element
const list = findAll('.my-selector');
list.forEach((obj, idx) => {
  find(`.my-selector:nth-of-type(${idx}) .my-child-selector`); // option 1 if you are lucky enough to be able to select it
  obj.querySelector('.my-child-selector'); // option 2, which works, but adds noise
});

// With ancestor element
list.forEach(obj => {
  find('.my-child-selector', obj);
});
```

In general, this allows people not having to worry when to use `find` and when to use `querySelector`, making the code look more uniform.

rwjblue:
IMHO, using `obj.querySelector` insde the loop would be much clearer

Serabe:
But for `finfAll`, people would need to use `Array.from(element.querySelectorAll(x));`

Serabe:
Closing this.

PR #316, opened 2018-02-10
Ensure tests pass properly for template-only-glimmer-components.

rwjblue:


tests/unit/legacy-0-6-x/test-module-for-component-test.js
```diff
@@ -389,6 +389,9 @@ moduleForComponent('Component Integration Tests', {
   integration: true,
   beforeSetup() {
     setResolverRegistry({
+      // registered manually to ensure template-only-glimmer-components tests pass
+      'component:my-component': Component,
```
Turbo87: I don't understand, why is this needed?

rwjblue: @Turbo87  Without this, the component in use is a template only component which (when the new optional feature is enabled) means that the behaviors being tested are a bit different.

Specifically the `{{name}}` in this components template will literally mean nothing because there is no `this` and `{{name}}` is implicitly `{{this.name}}`. The alternate solution to what I did here is to update the template to use `{{@name}}` instead, but that will _only_ work when running against Ember 3.1+ with the optional feature enabled.

Turbo87: but that means it is only necessary in this specific scenario, right? I think we should probably only add it in this scenario then 🤔

rwjblue: That's fine and I don't have a problem with doing that, but it doesn't really change anything. In the other scenarios an `Ember.Component` is already being created to back the template, by registering it here all the time we just make the behavior consistent across all of our versions and scenarios.

Turbo87: my understanding was that before, we were implicitly testing that "old-style" template-only components work correctly too

rwjblue: True, though there really is no difference prior to the new optional feature. Ember always used either the component JS file that you provided or `Ember.Component`. 

Either way, I'll update to keep things being tested in roughly the same manner...

rwjblue:
Ugh.  Looks like we are going to have an issue with the way our babel transpilation is working.  On older Ember versions the `hbs` helper will error when compiling `{{@names}}` (because Ember decided to disallow that functionality when it did glimmer integration for 2.10).

Turbo87:
ouch... that seems somewhat hard to fix 🤔

rwjblue:
The only thing I can think of is to have the babel plugin swallow errors, but obviously that is pretty bad for a whole swath of legit errors 🤔

rwjblue:
should be able to use ember-named-arguments-polyfill for this now...

rwjblue:
Still need to pick this back up and add the named arguments polyfill in devDeps.

PR #317, opened 2018-02-10
Remove simple alias for `wait` -> `settled`.

### LINKED ISSUE #283: Unsafe options usage in isSettled

### LINKED ISSUE #300: Arbitrary arguments to `wait` cause error

rwjblue:
The original goal of `settled` was to "rebrand" `wait` as `settled` without the 'baggage" (since `wait` had some legacy APIs that should not exist in `settled`). Unfortunately, the code refactoring that I had originally done that allowed them to be shared resulted in introducing bugs like those reported in #283.

This commit completely severs the implementation of `wait` from that of `settled`, by leveraging the (IMHO very nice) layers of abstraction that we have created since the original extraction.  Implementing `wait` with `waitUntil` and `getSettledState` makes this quite trivial now.

tldr of changes:

* `isSettled` no longer has to support the legacy `wait` contract (this was undocumented private behavior _only_ kept around to support the old semantics of `wait`)
* `wait` is implemented as its own `waitUntil` with the same semantics it has had since the 0.6.x series

Fixes #300
Fixes #283
Closes #303

### DIFF
--- a/addon-test-support/ember-test-helpers/wait.js
+++ b/addon-test-support/ember-test-helpers/wait.js
@@ -1,7 +1,48 @@
 export {
-  default,
   _setupAJAXHooks,
   _setupPromiseListeners,
   _teardownAJAXHooks,
   _teardownPromiseListeners,
 } from '@ember/test-helpers/settled';
+
+import { waitUntil, getSettledState } from '@ember/test-helpers';
+
+/**
+  Returns a promise that resolves when in a settled state (see `isSettled` for
+  a definition of "settled state").
+
+  @private
+  @deprecated
+  @param {Object} [options={}] the options to be used for waiting
+  @param {boolean} [options.waitForTimers=true] should timers be waited upon
+  @param {boolean} [options.waitForAjax=true] should $.ajax requests be waited upon
+  @param {boolean} [options.waitForWaiters=true] should test waiters be waited upon
+  @returns {Promise<void>} resolves when settled
+*/
+export default function wait(options = {}) {
+  if (typeof options !== 'object' || options === null) {
+    options = {};
+  }
+
+  return waitUntil(() => {
+    let waitForTimers = 'waitForTimers' in options ? options.waitForTimers : true;
+    let waitForAJAX = 'waitForAJAX' in options ? options.waitForAJAX : true;
+    let waitForWaiters = 'waitForWaiters' in options ? options.waitForWaiters : true;
+
+    let { hasPendingTimers, hasRunLoop, hasPendingRequests, hasPendingWaiters } = getSettledState();
+
+    if (waitForTimers && (hasPendingTimers || hasRunLoop)) {
+      return false;
+    }
+
+    if (waitForAJAX && hasPendingRequests) {
+      return false;
+    }
+
+    if (waitForWaiters && hasPendingWaiters) {
+      return false;
+    }
+
+    return true;
+  });
+}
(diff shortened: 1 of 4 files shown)


addon-test-support/ember-test-helpers/wait.js
```diff
@@ -1,7 +1,46 @@
...
+
+/**
+  Returns a promise that resolves when in a settled state (see `isSettled` for
+  a definition of "settled state").
+
+  @private
+  @deprecated
+  @returns {Promise<void>} resolves when settled
+*/
+export default function settled() {
```
Turbo87: I think this should be called `wait()`, not `settled()`

addon-test-support/ember-test-helpers/wait.js
```diff
@@ -1,7 +1,46 @@
...
+  let options = arguments[0];
```
Turbo87: we should use `function wait(options = {}) {` instead

addon-test-support/ember-test-helpers/wait.js
rwjblue: DOH! Indeed!

addon-test-support/ember-test-helpers/wait.js
rwjblue: kk, seems good

rwjblue: Hmm, actually its a tad different.  The old 0.6.x `wait` code avoided errors when something like a number was passed in, if I use default arguments here it will only correct for `undefined` specifically...

Turbo87: sure, the logic to handle non-objects would likely still need to exists, but I think putting `options = {}` in would still be good for documentation purposes

rwjblue: updated

PR #326, opened 2018-02-15
Add test for firing file change event twice

amk221:
Even though ember-test-helpers doesn't have a `selectFiles` helper (yet?) it does handle firing of events with files. However, it was not possible to select files more than once (not to be confused with selecting multiple files at once).

### DIFF
--- a/tests/unit/dom/select-files-test.js
+++ b/tests/unit/dom/select-files-test.js
@@ -1,0 +1,55 @@
+import { module, test } from 'qunit';
+import { triggerEvent, setupContext, teardownContext } from '@ember/test-helpers';
+import { buildInstrumentedElement } from '../../helpers/events';
+import hasEmberVersion from 'ember-test-helpers/has-ember-version';
+
+module('DOM Helper: selectFiles', function(hooks) {
+  if (!hasEmberVersion(2, 4)) {
+    return;
+  }
+
+  let context, element;
+
+  const textFile = new Blob(['Hello World'], { type: 'text/plain' });
+  textFile.name = 'text-file.txt';
+
+  const imageFile = new Blob([], { type: 'text/png' });
+  imageFile.name = 'image-file.png';
+
+  hooks.beforeEach(function() {
+    // used to simulate how `setupRenderingTest` (and soon `setupApplicationTest`)
+    // set context.element to the rootElement
+    context = {
+      element: document.querySelector('#qunit-fixture'),
+    };
+  });
+
+  hooks.afterEach(async function() {
+    if (element) {
+      element.parentNode.removeChild(element);
+    }
+
+    // only teardown if setupContext was called
+    if (context.owner) {
+      await teardownContext(context);
+    }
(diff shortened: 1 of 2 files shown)


tests/unit/dom/select-files-test.js
```diff
@@ -0,0 +1,53 @@
...
+import hasEmberVersion from 'ember-test-helpers/has-ember-version';
+
+module('DOM Helper: selectFiles', function(hooks) {
+  if (!hasEmberVersion(2, 4)) {
+    return;
+  }
+
+  let context, element;
+
+  const textFile = new Blob(['Hello World'], { type: 'text/plain' });
```
rwjblue: Do we need to guard this? It doesn't look like `Blob` is available in IE11 (and I want to make sure the test suite still passes there)...

tests/unit/dom/select-files-test.js
```diff
@@ -0,0 +1,53 @@
...
+  textFile.name = 'text-file.txt';
+
+  const imageFile = new Blob([], { type: 'text/png' });
+  imageFile.name = 'image-file.png';
+
+  hooks.beforeEach(function() {
+    // used to simulate how `setupRenderingTest` (and soon `setupApplicationTest`)
```
rwjblue: I think this is ok, but it might be better to _actually_ use `setupRenderingTest` (may end up being less brittle over time), what do you think?

tests/unit/dom/select-files-test.js
amk221: I could send in POJOs rather than skipping the test for IE entirely?

tests/unit/dom/select-files-test.js
amk221: I just copied an example from one of the existing helper tests. I'm not sure I have enough knowledge to get this PR in a good enough state to go in. I just made it to demonstrate what I need.

rwjblue: Ya, no worries about that. There really isn’t an issue with it the way it is. Just made me reconsider RE: the other tests like this

walter:
Thanks for the fix!

Note for others that might be experiencing this... I had this issue w/ `triggerEvent` for file upload. Upgraded to `0.7.20` and issue was resolved.

ajcolter:
@walter I'm having the same issue now. What did you upgrade to `0.7.20`? `ember-test-helpers` latest non-beta release is `0.6.3`.

rwjblue:
`ember-test-helpers` should be deprecated, I'll try to look into that...

ajcolter:
@rwjblue, my app is on Ember 2.18.2. I upgraded ember-cli-qunit to 4.3.2 for the new test framework. According to my yarn.lock file, ember-qunit depends on `@ember/test-helpers` at `^0.7.18`. Could this be upgraded easily to at least `0.7.20` to avoid this issue in the future? It was weird to me that I had to install it as a dependency myself (when I was also able to use it just fine in most tests).

Issue #335, opened 2018-03-01
CI test suite is broken with latest QUnit release


Turbo87:
The latest QUnit release is now verifying that when `assert.step()` was called you must also call `assert.verifySteps()` later (see https://github.com/qunitjs/qunit/pull/1252). Unfortunately that breaks our assumptions and the "instrumented element" test abstraction we use to verify the events that an element emits. In some tests we use these instrumented elements without verifying the event order at the end which is now causing issues due to the QUnit changes.

There are a few potential solutions:

- Run `assert.steps = []` in the `afterEach` blocks to work around the new assertion (`assert.steps` appears to be private API though)

- Use instrumented and regular elements depending on the test

- Run `assert.verifyStep()` at the end of all tests even though the results don't actually matter to that test

IMHO none of those solutions are really great, but we will need to find/choose one to unblock our CI 🤔 

/cc @rwjblue

snewcomer:
@Turbo87 Anything help I can provide or is this mainly a Robert question?  I might be in favor of # 3 though.

Turbo87:
@snewcomer I guess any solution is better than no solution, so if you can work on a PR that unblocks CI that would be wonderful :)

PR #336, opened 2018-03-03
make getApplication public

snewcomer:
One use case I have is I would like to call `setApplication(Application.create({ autoboot: false, custom: 'bar' });` or something similar.  However, this may be a bit dangerous because this sets the application for the rest of the test run.  In order to ensure further calls to get the application instance aren't this Application instance, I need to reset it with the old application instance using `this.oldAppInstance = getApplication()` before setting.

Lmk what you think!

Ref #335


snewcomer:
Should we have the tests passing before merging this?  Took a shot with #341 but may need some help.

rwjblue:
Thank you!

PR #340, opened 2018-03-12
Add some specific documentation for uploading files

jrjohnson:
Uploading files with an event is a specific code path and the task can
be tough to figure out without an example.

I couldn't think of a better way to document this, but it seemed necessary to me.  I've been using a construct like:

```
inputElement.triggerHandler({
 type: 'change',
 target: {
  files: {
   0: file,
   length: 1,
   item() { return file; }
  }
 }
});
```

So when converting from jQuery helpers to this addon it was difficult to see how to make file uploads work. Turns out some really nice work is done in `fire-event.js` - just needed to be documented.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/trigger-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/trigger-event.js
@@ -4,14 +4,25 @@ import settled from '../settled';
 import { nextTickPromise } from '../-utils';
 
 /**
-  Triggers an event on the specified target.
-
-  @public
-  @param {string|Element} target the element or selector to trigger the event on
-  @param {string} eventType the type of event to trigger
-  @param {Object} options additional properties to be set on the event
-  @return {Promise<void>} resolves when the application is settled
-*/
+ * Triggers an event on the specified target.
+ *
+ * @public
+ * @param {string|Element} target the element or selector to trigger the event on
+ * @param {string} eventType the type of event to trigger
+ * @param {Object} options additional properties to be set on the event
+ * @return {Promise<void>} resolves when the application is settled
+ *
+ * @example
+ * <caption>Using triggerEvent to Upload a file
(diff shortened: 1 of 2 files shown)


jrjohnson:
I'm assuming these test failures have nothing to do with this doc change.

Turbo87:
The API docs are automatically generated, so you would have to change this in the JS docs, not in the API docs directly.

jrjohnson:
Whoops. Thanks @Turbo87. I fixed that, it doesn't like like docs get built in Travis so I assume I was supposed to commit the change to `API.md` as well?

I had to add `*`'s to the docs. Otherwise the whitespace was stripped from the example.

addon-test-support/@ember/test-helpers/dom/trigger-event.js
```diff
@@ -4,14 +4,27 @@ import settled from '../settled';
...
+ * <caption>Using triggerEvent to Upload a file
+ * When using triggerEvent to upload a file the `eventType` must be `change` and  you must pass an
+ * array of [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) as `options`.</caption>
+ *
+ * ```javascript
+ * triggerEvent(
+ *   'input.fileUpload',
+ *   'change',
+ *   [new Blob(['Ember Rules!'])]
+ *  );
```
ro0gr: a minor indentation issue, this bracket should be moved one char left

jrjohnson: Thanks @ro0gr. Fixed.

jrjohnson:
@Turbo87 / @ro0gr  just wanted to check in. Is this something worth merging? I'm ok with closing it un-merged if it isn't.

rwjblue:
Yes! Absolutely!

rwjblue:
Thank you!

PR #341, opened 2018-03-12
call verifySteps at end of all tests

### LINKED ISSUE #335: CI test suite is broken with latest QUnit release

snewcomer:
Some tests do not have an `assert.verifyStep` call.  Need to find and add where applicable.

- [x] blur
- [ ] fillIn
- [ ] click
- [ ] focus
- [ ] tap

close #335

### DIFF
--- a/tests/unit/dom/blur-test.js
+++ b/tests/unit/dom/blur-test.js
@@ -61,6 +61,7 @@ module('DOM Helper: blur', function(hooks) {
       blur(`#foo-bar-baz-not-here-ever-bye-bye`),
       /Element not found when calling `blur\('#foo-bar-baz-not-here-ever-bye-bye'\)`/
     );
+    assert.verifySteps([]);
   });
 
   test('bluring via selector with context set', async function(assert) {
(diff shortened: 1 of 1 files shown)


snewcomer:
@Turbo87 I might be a bit confused as to what exactly I would need to do to get those tests passing...Specifically tried out `blur` first.

https://travis-ci.org/emberjs/ember-test-helpers/jobs/352423827

PR #346, opened 2018-03-19
Allow for custom error message in waitUntil helper

BitBrit:
We have need to distinguish between different waitUntil timeouts. To do this, it would be nice to be able to pass a custom error message to waitUntil which it will use upon promise reject. This small PR does that.


addon-test-support/@ember/test-helpers/wait-until.js
```diff
@@ -18,9 +18,10 @@ const MAX_TIMEOUT = 10;
 */
 export default function waitUntil(callback, options = {}) {
   let timeout = 'timeout' in options ? options.timeout : 1000;
+  let errorMsg = 'errorMessage' in options ? options.errorMessage : 'waitUntil timed out';
```
rwjblue: lets call this explicitly `timeoutMessage` (we may have other types of errors in the future, so I'd like to avoid using the more generic name for now)

PR #348, opened 2018-03-26
wait-for helper's count value is null by default.

### LINKED ISSUE #312: Documentation of the `count` option of the `waitFor` helper is wrong
The documentation suggests that the default is `1`, but in fact it is `null` which basically means "one or more" instead of "exactly one".

sly7-7:
Closes #312


API.md
```diff
@@ -41,10 +41,10 @@
     -   [setupRenderingContext](#setuprenderingcontext)
     -   [teardownRenderingContext](#teardownrenderingcontext)
     -   [setApplication](#setapplication)
-    -   [getApplication](#getapplication)
     -   [setupApplicationContext](#setupapplicationcontext)
     -   [teardownApplicationContext](#teardownapplicationcontext)
     -   [validateErrorHandler](#validateerrorhandler)
+-   [getApplication](#getapplication)
```
rwjblue: I assume this was updated due to running `yarn docs`?  We need to tweak the configuration to show `getApplication` under the `Test Framework APIs` section.  I think all we need to do is add it to this list https://github.com/emberjs/ember-test-helpers/blob/184bb747a2d08f5af86d545e2f4a79db96301f51/documentation.yml#L42

sly7-7: Yep, it was a updated as a side effect. Do you want a separate PR ?

Turbo87: yeah, I guess this PR should focus on the `waitFor` thing only

rwjblue:
Thank you!

PR #349, opened 2018-03-26
Protect context properties and methods from overwriting

### LINKED ISSUE #244: Add "owner" reserved word warning
After running [`ember-qunit-codemod`](https://github.com/rwjblue/ember-qunit-codemod) against our app I observed that a test was falling with the following message:

`TypeError: Cannot read property 'apply' of undefined`

The relevant potion of the affected test:

```
  test('it displays the project owner', async function(assert) {
    this.set('owner', 'some_username');
    await render(hbs`{{project owner=owner}}`);
    
    ...
    
  });
```

It looks like this is now falling because the `owner` property is now set within the test context (https://github.com/emberjs/ember-test-helpers/blame/0cacde9a1ba308409394b18f3705e60499955b45/addon-test-support/%40ember/test-helpers/setup-context.js#L90
).

I am not sure what may be the best way to handle this, but, I am thinking adding a check prior to setting `owner` may be helpful.

Turbo87:
Resolves #244

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-context.js
@@ -167,31 +167,46 @@ export default function(context, options = {}) {
       return buildOwner(getApplication(), getResolver());
     })
     .then(owner => {
-      context.owner = owner;
+      Object.defineProperty(context, 'owner', {
+        value: owner,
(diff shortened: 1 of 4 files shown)


Turbo87:
prettier....... 😭

Issue #351, opened 2018-03-29
find and findAll helpers do not use HTML element context properly


jbingham94:
It seems that both the `find` and `findAll` helpers do not properly use context if you pass in an HTML element.  For example:

```javascript
find('.my-selector', myHTMLElement); // I want to look for .my-selector within myHTMLElement
```

Instead of looking within myHTMLElement, the helpers seem to look within the first element with myHTMLElement's selector.  When I switch back to using `find` from ember-native-dom-helpers, the above code works properly.

Turbo87:
the reason for this is that the `find()` functions in this library here don't have a `context` parameter. see https://github.com/emberjs/ember-test-helpers/blob/master/API.md#find

jbingham94:
@Turbo87 Thanks for clearing that up. I don't feel like this would be super tough to implement, and it would definitely be a very helpful feature, as almost every find/findAll we do has some sort of context.  This limitation is pretty much the only thing stopping us from ditching ember-native-dom-helpers completely.

synaptiko:
Are you willing to reconsider adding `context`? It would match quite well with `assert.dom()` from `qunit-dom`.

Turbo87:
qunit-dom is doing a bit more and have a slightly different API though.

instead of `find('.my-selector', myHTMLElement);` you can just use `myHTMLElement.querySelector('.my-selector');`

rwjblue:
To reiterate from prior conversations on this, the `find` / `findAll` variants should generally only be used in tooling without access to the test context directly, user-land test code should totally use `this.element` directly.

The path forward here to prevent accidental trollage is to trigger an error when passing a second argument. As mentioned in https://github.com/emberjs/ember-test-helpers/issues/351#issuecomment-377321867.

synaptiko:
@rwjblue We recently refactored our addons and apps from Ember 2.18 to 3.5. Usage of `find`/`findAll` vs `this.element.querySelector/All` was one of the things we were considering but it wasn't clear to us that `find`/`findAll` is not intended to be used by user-land test code. Especially when the default behavior of https://github.com/simonihmig/ember-test-helpers-codemod is usage of `find`/`findAll`. There is an additional codemod to convert it to `this.element`. This was something which confused us.

rwjblue:
[...]

PR #354, opened 2018-04-02
#351 Adding context for find and findAll helpers

jbingham94:
I added an optional context param to both the find and findAll helpers.  This was a helpful feature in ember-native-dom-helpers, allowing one to write very clean and readable test loops.  I also tightened up the test code a bit for both helpers.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/-get-element.js
+++ b/addon-test-support/@ember/test-helpers/dom/-get-element.js
@@ -1,13 +1,14 @@
-import getRootElement from './get-root-element';
+import getContextElement from './get-context-element';
 
 /**
   Used internally by the DOM interaction helpers to find one element.
 
   @private
   @param {string|Element} target the element or selector to retrieve
+  @param {string|Element} [context] the context element or selector
   @returns {Element} the target or selector
 */
-export default function getElement(target) {
+export default function getElement(target, context) {
   if (
     target.nodeType === Node.ELEMENT_NODE ||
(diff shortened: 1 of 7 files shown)


rwjblue:
See conversation in https://github.com/emberjs/ember-test-helpers/pull/313, I do not believe this is a direction we want to go...

jbingham94:
Hey @rwjblue.  While I understand that it is theoretically possible to have some of the same behavior by using a more complicated selector, it's just not as elegant.  Also, it doesn't work in all cases.  Suppose you want to query within a specific element, not just the first element with a given selector.  Maybe you could use nth-of-type, but it just gets messier and messier the more you try to stuff in that big selector.  Honestly, you could do most things in this library (and definitely in the find and findAll helpers) just using built-in Javascript, but the point is to give developers a more concise way of writing things.  So I'm not really sure I follow your argument that the context parameter is superfluous

rwjblue:
In general folks should be using `element.querySelector`directly, and only using `find` as a last resort. If you have a context element already, just do `context.querySelector` or `context.querySelectorAll`. 

More generally though, this library is all about basic primitives. I think the primitives already exist to support this need. It seems quite trivial to create an addon that makes a `findFrom` or `findWithContext` that works _exactly_ how you want here, and have that addon leverage `find` or `findAll` from `@ember/test-helpers` (or `getRootElement()` and `.querySelector` directly).

jbingham94:
@rwjblue Thanks for the explanation!  I understand your argument from the other PR better now.  I will close this out.

PR #356, opened 2018-04-05
add waitFor msg

snewcomer:
Opens up `waitFor` to have a `timeoutMessage`.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/wait-for.js
+++ b/addon-test-support/@ember/test-helpers/dom/wait-for.js
@@ -15,7 +15,7 @@ import { nextTickPromise } from '../-utils';
   @param {number} [options.count=null] the number of elements that should match the provided selector (null means one or more)
   @returns {Element|Array<Element>} the element (or array of elements) that were being waited upon
 */
-export default function waitFor(selector, { timeout = 1000, count = null } = {}) {
+export default function waitFor(selector, { timeout = 1000, count = null, timeoutMessage = 'waitFor timed out' } = {}) {
   return nextTickPromise().then(() => {
     if (!selector) {
       throw new Error('Must pass a selector to `waitFor`.');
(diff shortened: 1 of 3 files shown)


addon-test-support/@ember/test-helpers/dom/wait-for.js
```diff
@@ -15,7 +15,7 @@ import { nextTickPromise } from '../-utils';
   @param {number} [options.count=null] the number of elements that should match the provided selector (null means one or more)
   @returns {Element|Array<Element>} the element (or array of elements) that were being waited upon
 */
-export default function waitFor(selector, { timeout = 1000, count = null } = {}) {
+export default function waitFor(selector, { timeout = 1000, count = null, timeoutMessage = 'waitFor timed out' } = {}) {
```
snewcomer: I can revert the default and keep `waitUntil timed out` if you would like.

tests/unit/dom/wait-for-test.js
```diff
@@ -71,11 +71,11 @@ module('DOM Helper: waitFor', function(hooks) {
 
     let start = Date.now();
     try {
-      await waitFor('.something', { timeout: 100 });
+      await waitFor('.something', { timeout: 100, timeoutMessage: '.something timed out' });
```
rwjblue: Lets leave this test alone, and add another that is specifically for modifying the timeout message...

addon-test-support/@ember/test-helpers/dom/wait-for.js
rwjblue: Nah, I think this is fine...

rwjblue:
@snewcomer - 👋 looks like some linting failures here still but if you fix those and rebase against master, CI should be green again...

rwjblue:
Thank you!

PR #360, opened 2018-04-17
revert part of #349 - only new function read-only

### LINKED ISSUE #359: setting this.render is broken in 0.7.21
https://github.com/kellyselden/ember-test-setup/blob/master/addon-test-support/-setup-render.js#L9 was working on the assumption in 0.7.20 that I could override `this.render`. It was broken in https://github.com/emberjs/ember-test-helpers/pull/349. I understand why it was changed, but it sucks I can't have my desired API anymore.

kellyselden:
Closes #359. Revert part of #349

My thought is that we can mark new function API (owner, element) as read-only, and leave the ports (render, set, $, etc) as-is. If someone was overriding these previously, their tests were either already broken, or they had a reason for overriding (like me). It retains the original fix for #244. Is this an acceptable compromise?


Turbo87:
see https://github.com/emberjs/ember-test-helpers/issues/359#issuecomment-382278238

PR #361, opened 2018-04-18
Fix KeyboardEvent#key and allow `triggerKeyEvent` to receive keys instead of keycodes

### LINKED ISSUE #265: Update `triggerKeyEvent` to handle `code` and `key`.
From https://github.com/emberjs/ember-test-helpers/pull/258#discussion_r155963086:

> When I initially copied `keyEvent`, I continued to use `keyCode` as the main way to set the key, but it seems to be [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and the new alternatives are [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) and [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key), specially the last one.
>
> Perhaps we might want to start using it, or maybe assume that integers mean `keyCode` and strings mean `key`.

cibernox:
Closes #265

Now: `triggerKeyEvent(selector, 'keydown', 9)` and `triggerKeyEvent(selector, 'keydown', 'Tab')` are equivalent.
Also, the generated event in both cases has `e.keyCode === 9` and `e.key === 'Tab'`

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
@@ -12,13 +12,96 @@ const DEFAULT_MODIFIERS = Object.freeze({
   metaKey: false,
 });
 
+// This is not a comprehensive list, but it is better than nothing.
+const keyFromKeyCode = {
+  8: 'Backspace',
+  9: 'Tab',
+  13: 'Enter',
+  16: 'Shift',
+  17: 'Control',
+  18: 'Alt',
+  20: 'CapsLock',
+  27: 'Escape',
+  32: '',
+  37: 'ArrowLeft',
+  38: 'ArrowUp',
+  39: 'ArrowRight',
+  40: 'ArrowDown',
+  65: 'a',
+  66: 'b',
+  67: 'c',
+  68: 'd',
+  69: 'e',
+  70: 'f',
+  71: 'g',
+  72: 'h',
+  73: 'i',
+  74: 'j',
+  75: 'k',
+  76: 'l',
+  77: 'm',
+  78: 'n',
+  79: 'o',
+  80: 'p',
+  81: 'q',
+  82: 'r',
+  83: 's',
+  84: 't',
+  85: 'u',
+  86: 'v',
+  87: 'v',
+  88: 'x',
+  89: 'y',
+  90: 'z',
+  91: 'Meta',
+  93: 'Meta', // There is two keys that map to meta,
+  187: '=',
+  189: '-',
+};
+
+/**
+  Calculates the value of KeyboardEvent#key given a keycode and the modifiers.
+  Note that this works if the key is pressed in combination with the shift key, but it cannot
+  detect if caps lock is enabled.
+  @param {number} keycode The keycode of the event.
+  @param {object} modifiers The modifiers of the event.
+  @returns {string} The key string for the event.
+ */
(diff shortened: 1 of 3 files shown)


cibernox:
Test failures seems unrelated to this change 🤔

tests/unit/dom/trigger-key-event-test.js
```diff
@@ -123,4 +123,110 @@ module('DOM Helper: triggerKeyEvent', function(hooks) {
...
+    await checkKey(40, 'ArrowDown');
+    await checkKey(91, 'Meta');
+    await checkKey(93, 'Meta');
+    await checkKey(187, '=');
+    await checkKey(189, '-');
+    await checkKey(65, 'a');
+    await checkKey(90, 'z');
+    await checkKey(65, 'A', { shiftKey: true });
+    await checkKey(90, 'Z', { shiftKey: true });
+    assert.verifySteps([
```
cibernox: Btw, verifing steps like this is pretty annoying. Is there a way to avoid this?

localpcguy:
Just cause it may be helpful, @wesbos made a utility site at keycode.info and in the JS he has a pretty comprehensive mapping keyCodes to their English equivalent. 
https://github.com/wesbos/keycodes/blob/gh-pages/scripts.js#L3

It isn't mapped to key values directly, but a more English representation of what that keycode is, but I thought it might be helpful to ensure this provides a larger set of available keyCodes.

cibernox:
This should be green now

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -48,11 +131,17 @@ export default function triggerKeyEvent(target, eventType, keyCode, modifiers =
       );
     }
 
-    if (!keyCode) {
-      throw new Error(`Must provide a \`keyCode\` to \`triggerKeyEvent\``);
+    if (key === null || key === undefined) {
```
rwjblue: Can we instead check that they are as expected?

```js
let keyType = typeof key;
if (keyType !== 'number' && keyType !== 'string') {
  throw new Error(...);
}
```

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -12,21 +12,104 @@ const DEFAULT_MODIFIERS = Object.freeze({
...
+}
+
 /**
   Triggers a keyboard event on the specified target.
 
   @public
   @param {string|Element} target the element or selector to trigger the event on
   @param {'keydown' | 'keyup' | 'keypress'} eventType the type of event to trigger
-  @param {string} keyCode the keyCode of the event being triggered
+  @param {number|string} key the keyCode of the event being triggered
```
rwjblue: The docs here still refer to this as `the keyCode of the event being triggered`, but it seems that its now either they key or the keycode.

Can you also add more prose to the description above regarding this argument? Perhaps even an example of each way and a link to MDN or something?

tests/unit/dom/trigger-key-event-test.js
rwjblue: Don't build an "instrumented element", thats why you have to verify in this manner (because the instrumented elements also assert no other events are fired via steps).

cibernox:
Improved 😄

rwjblue:
Thank you @cibernox !

PR #366, opened 2018-04-20
Fix no-jquery scenarios

cibernox:
Since `ember-native-dom-event-dispatcher` is not required in ember 3.0, checking for its presence to decide wether or not to load jquery is not a good strategy anymore.


ember-cli-build.js
```diff
@@ -13,7 +17,11 @@ module.exports = function(defaults) {
     },
   };
 
-  if (defaults.project.findAddonByName('ember-native-dom-event-dispatcher')) {
+  if (
+    (project.findAddonByName('ember-native-dom-event-dispatcher') ||
+      emberChecker.isAbove('3.0.0')) &&
+    project.findAddonByName('ember-fetch')
```
Turbo87: we're not using `ember-fetch` anywhere though 🤔

cibernox: We do:
- https://github.com/emberjs/ember-test-helpers/blob/master/config/ember-try.js#L119-L124
- https://github.com/emberjs/ember-test-helpers/blob/master/tests/helpers/ajax.js#L16

Turbo87: 🤦‍♂️  nevermind...

PR #369, opened 2018-04-26
Clobber `self.find` when running rendering or application tests.

rwjblue:
Forgetting to import `find`, and then calling the global `find` function has become a very common pain point as folks migrate from the older system (where `find` was always globally clobbered to be our own special version) and the new system where you must import it.

This fixes many WAT's...

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-application-context.js
+++ b/addon-test-support/@ember/test-helpers/setup-application-context.js
@@ -1,9 +1,13 @@
 /* globals EmberENV */
+import { guidFor } from '@ember/object/internals';
 import { get } from '@ember/object';
 import { nextTickPromise } from './-utils';
+import global from './global';
 import { getContext } from './setup-context';
 import hasEmberVersion from './has-ember-version';
 import settled from './settled';
+
+export const APPLICATION_CLEANUP = Object.create(null);
 
 /**
   Navigate the application to the provided URL.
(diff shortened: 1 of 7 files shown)


Turbo87:
Does that interfere with old-style testing where Ember sets the `find` global?

cibernox:
Yass. If what @Turbo87 is not the case, this will save a lot of f**ks per hour as a whole.

rwjblue:
> Does that interfere with old-style testing where Ember sets the find global?

Notice that the `teardownApplicationContext`/`teardownRenderingContext` here reset `self.find` after each test.

And when using legacy style tests they do not call into this pathway. They would either not use ember-test-helpers code at all (the default `moduleForAcceptance` didn't use any code from this repo), or they would not run these `setupRenderingTest` / `setupApplicationTest` methods (e.g. `ember-mocha` which uses the `moduleForAcceptance` from this repo).

rwjblue:
Just pushed some tests demonstrating things working properly together...

kellyselden:
Would this be better solved via a eslint-plugin-ember rule in the included recommended? or is it not possible to de-global things?

Turbo87:
> or is it not possible to de-global things?

yes, that unfortunately seems to be the case. usually https://eslint.org/docs/rules/no-undef should take care of it, but since `find` is a legitimate global defined by `env: { browser: true }` we can't undefined it in ESLint. the only solution would be a dedicated rule that just looks for the `find` global and nothing else. might be worth it, idk.

Turbo87:
looks like https://eslint.org/docs/rules/no-restricted-globals might be an option 🤔

kellyselden:
@Turbo87 indeed it is. I forgot I had already added this to our linter https://github.com/peopleconnectus/eslint-config-peopleconnect-ember/blob/2be0e59b012da91cbb3379efa01d9d22c0c92087/index.js#L23

Issue #332, opened 2018-02-28
visit() throws TransitionAborted error


simonihmig:
As discussed in Slack, I was unexpectedly seeing failing tests where a transition was aborted. 
`ApplicationInstance#visit` is throwing it [here](https://github.com/emberjs/ember.js/blob/8be1a9d1d4fbbefda5623efa9c35ffadcf127802/packages/ember-application/lib/system/application-instance.js#L256-L257)

As I said I cannot share the app, but this is the human readable test spec (we have a rather exotic test setup with `ember-cli-yadda` 😉):

```gherkin
  Scenario: Access restricted page as invited user and login
    Given there exists a user with username "testname" and password "testpw"
    And I am an invited user
    And I am on the homepage
    When I go to the user profile page
    Then I should still be on the homepage
    And I should see the login modal
    When I insert "testname" as the username
    And I insert "testpw" as the password
    And I submit the login form
    Then I should be on the user profile page
```

The `I go to the user profile page` step is throwing the exception.

I quickly assembled this little repro, that resembles what we were doing in a very minimal way:
https://github.com/simonihmig/ember-visit-transition-abort

All the acceptance tests are passing, with the exception of this last one:
https://github.com/simonihmig/ember-visit-transition-abort/blob/master/tests/acceptance/auth-test.js#L36-L40

simonihmig:
@rwjblue did you have a chance to look at this? 

I am happy to work on a PR, if we agree what the solution should be. (this is the only remaining issue left for our app's test suite refactoring PR ;))

I am not fully aware of the inner workings, but it puzzles me that clicking on the `link-to` is different to using `visit`. (See [this](https://github.com/simonihmig/ember-visit-transition-abort/blob/master/tests/acceptance/auth-test.js#L24-L28) vs. [that](https://github.com/simonihmig/ember-visit-transition-abort/blob/master/tests/acceptance/auth-test.js#L36-L40))

spencer516:
@simonihmig — we've been running into this too. Just opened a PR with a failing test and a place to start: https://github.com/emberjs/ember-test-helpers/pull/373

esbanarango:
☝️ We're having the same issue when migrating to the new testing API.

NullVoxPopuli:
I have some `transitionTo`s happening in my afterModel hook -- which is intentional -- but in my test `visit` is raising a TransitionAborted exception.

The only way for my test to pass is:

```
            try {
              await visit(url);
            } catch(e) {
[...]

PR #373, opened 2018-05-08
TransitionAborted on a `visit()`

spencer516:
We ran into [this issue](https://github.com/emberjs/ember-test-helpers/issues/332) where an aborted transition (which was the behavior we expected) was causing the `visit()` helper to reject failing our test prematurely. 

_I don't think_ this was the case before when writing Acceptance Tests before (it may just be the nature of awaiting promises now); but this also feels unexpected or broken. 

I'm not sure of the implications of this fix to suppress the error if it's message is `TransitionAborted`. 

Are there times when we _would_ want that promise to reject when a transition aborts? Should this be configurable instead? 

Something like:

```js
visit('/home', { ignoreAbortedTransitions: true });
```


tests/unit/setup-application-context-test.js
```diff
@@ -127,4 +135,11 @@ module('setupApplicationContext', function(hooks) {
   test('bubbles up errors', function(assert) {
     assert.rejects(visit('/widgets'), /Model hook error from \/widgets/);
   });
+
+  test('`transitionAborted` errors are not treated as test errors', async function(assert) {
+    await visit('/purgatory');
+    await visit('/');
+
+    assert.equal(currentURL(), '/purgatory');
```
simonihmig: I might miss something obvious, but why is the current URL expected to be `/purgatory` here, after having visited `/`?

rwjblue:
Sorry for the super long delay here, I'd love to resurrect this PR would you have any time to rebase it?

spencer516:
@rwjblue — just rebased. 

However, this test is now failing for me in Safari :-P Chrome seems fine. 

I wasn't able to dig too far into it, but when I disable the Ember 3.4 check in [`nextTickPromise()`](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/-utils.ts#L84), the test is passing again. 

Something something RSVP and Native Promises commingling === Unhappy Safari? This type of error feels like there's a new promise getting created somewhere that rejects that ends up not being a part of the promise chain where the catch is added in this PR...maybe?

spencer516:
Cool beans. I just made a change to get the `2.4` -> `2.12` scenarios to pass — and rather than aborting the transition, I just re-transition. Test now passes in safari, but that feels like a cheat. 

Does `transition.abort()` create a rejected promise? 

For what it's worth, the `bubbles up errors` test is _also_ failing in Safari for, perhaps, a similar reason. 

image

rwjblue:
> Does transition.abort() create a rejected promise?

Yes, aborting a transition is essentially the same as rejecting a promise.

samselikoff:
We're running into this, any updates on this or workarounds for now?

patsy-issa:
Fwiw using `redirect()` to redirect instead of `before/after/model` does not cause `visit` to fail.

Issue #368, opened 2018-04-24
Throw or warn when the user passes a downcased key event


cibernox:
`triggerKeyEvent` now allows to pass a string as third argument, representing the `e.key`  of the dispatched event.

By the `KeyboardEvent#key` spec, the strings have to be capitalized (`Enter` instead of `enter`, `Tab` instead of `tab`, ...). But it's easy for users to forget and pass `enter`, which is invalid but they don't receive any warning or error.

We must help users by detect this mistake.

mydea:
Related to this, some of our tests were broken, where we had e.g. `triggerKeyEvent('keyup', '13')` - which used to work, but then stopped working. Removing the quotes made it work, but it was still pretty confusing, and took some time to figure out what was going on. I think for the sake of backwards-compatibility, it would be nice if it would detect "strings that are actually numbers" and just convert them to a number for you.

PR #379, opened 2018-05-24
Improve `triggerKeyEvent()` warnings

Turbo87:
Resolves https://github.com/emberjs/ember-test-helpers/issues/368

/cc @mydea

### DIFF
--- a/addon-test-support/@ember/test-helpers/-utils.js
+++ b/addon-test-support/@ember/test-helpers/-utils.js
@@ -35,3 +35,14 @@ export function runDestroyablesFor(object, property) {
 
   delete object[property];
 }
+
+/**
+ Returns whether the passed in string consists only of numeric characters.
+
+ @private
+ @param {string} n input string
+ @returns {boolean} whether the input string consists only of numeric characters
+ */
+export function isNumeric(n) {
+  return !isNaN(parseFloat(n)) && isFinite(n);
+}
--- a/addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
@@ -3,7 +3,7 @@ import getElement from './-get-element';
 import fireEvent from './fire-event';
 import settled from '../settled';
 import { KEYBOARD_EVENT_TYPES } from './fire-event';
-import { nextTickPromise } from '../-utils';
+import { nextTickPromise, isNumeric } from '../-utils';
 
 const DEFAULT_MODIFIERS = Object.freeze({
   ctrlKey: false,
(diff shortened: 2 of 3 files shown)


addon-test-support/@ember/test-helpers/-utils.js
```diff
@@ -35,3 +35,14 @@ export function runDestroyablesFor(object, property) {
...
+
+/**
+ Returns whether the passed in string consists only of numeric characters.
+
+ @private
+ @param {string} n input string
+ @returns {boolean} whether the input string consists only of numeric characters
+ */
+export function isNumeric(n) {
+  return !isNaN(parseFloat(n)) && isFinite(n);
```
cibernox: Keycodes are only integers. I think you should use parseInt

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
```diff
@@ -22,7 +22,7 @@ const keyFromKeyCode = {
   18: 'Alt',
   20: 'CapsLock',
   27: 'Escape',
-  32: '',
+  32: 'Space',
```
cibernox: Why this change? I think I remember getting those values myself and being surprised by the empty string, but it was the value I saw when debugging

Turbo87: I got that value from https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode 🤔

addon-test-support/@ember/test-helpers/-utils.js
Turbo87: this seems to be the accepted solution for a lot of these questions on StackOverflow with lots of people mentioning that `parseInt()` should not be used for some reason.

ultimately this is only used to catch the case where someone used `'13'` instead of `13`. if someone would put in `'13.0'` in the current state it would also warn, but if we changed it to only allow integers it would no longer show that warning, so I think it is actually better this way

addon-test-support/@ember/test-helpers/dom/trigger-key-event.js
cibernox: I created this which to test it and it seems to say otherwise: https://jsbin.com/qizayecoqe/edit?js,console,output

Turbo87: that is `' '`, not `''` though, so also something different. I'm fine with using `' '` instead.

cibernox: I thought I had a space, not an empty string. My bad.

Turbo87: updated to use `' '` now

PR #384, opened 2018-06-09
don't click when disabled

### LINKED ISSUE #383: click helper should ignore disabled inputs
`this.$('#foo').click()` and `document.getElementById('foo').click()` do nothing if the input is disabled. `await click('#foo')` still clicks disabled inputs. This is probably a bug.

kellyselden:
Closes #383.

Probably affects other helpers.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/click.js
+++ b/addon-test-support/@ember/test-helpers/dom/click.js
@@ -4,6 +4,7 @@ import { __focus__ } from './focus';
 import settled from '../settled';
 import isFocusable from './-is-focusable';
 import { nextTickPromise } from '../-utils';
+import isFormControl from './-is-form-control';
 
 /**
   @private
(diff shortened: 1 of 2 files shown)


addon-test-support/@ember/test-helpers/dom/click.js
```diff
@@ -59,7 +59,10 @@ export default function click(target) {
       throw new Error(`Element not found when calling \`click('${target}')\`.`);
     }
 
-    __click__(element);
+    if (!element.hasAttribute('disabled')) {
```
Turbo87: I think it should check the property instead of the attribute 🤔

kellyselden: updated

cibernox:
I'm confused. I understand that disabled inputs can't be filled, but can't they be clicked? That's surprising

Turbo87:
@cibernox I think you're confusing `disabled` with `readonly`

cibernox:
I'm not. I only expected disabled inputs to not be fillable or focusable, I didn't know they weren't clickable

addon-test-support/@ember/test-helpers/dom/click.js
```diff
@@ -59,7 +59,10 @@ export default function click(target) {
       throw new Error(`Element not found when calling \`click('${target}')\`.`);
     }
 
-    __click__(element);
+    if (!element.disabled) {
```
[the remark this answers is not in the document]
kellyselden: updated

rwjblue:
Thank you @kellyselden!

PR #386, opened 2018-06-15
Fix `waitFor` return value documentation

pcambra:


Turbo87:
@pcambra you're right that it returns a Promise, but that Promise does not resolve "when settled". It resolves when the element(s) appear on the page and it resolves with those elements. I think the correct return type would be `Promise<Element|Element[]>` in TypeScript, but no idea if JSDoc can handle that.

You will also need to adjust the documentation in the code itself: https://github.com/emberjs/ember-test-helpers/blob/3e91a29addb28d9a73e97f353c3bbc0985aa293d/addon-test-support/%40ember/test-helpers/dom/wait-for.js#L7-L17

... and then you can regenerate the documentation using `yarn docs`

pcambra:
Thanks for the pointers @Turbo87! 
Makes sense, but this is all very new to me, `{Promise<Element|Element[]>}` has generated some weird stuff :)

PR #387, opened 2018-06-19
Update buildMouseEvent to use MouseEvent constructor if it is present

### LINKED ISSUE #266: Update `fireEvent` to use `KeyboardEvent` and `MouseEvent` constructors.
From https://github.com/emberjs/ember-test-helpers/pull/258#discussion_r155955704.

> IDK if this targets Ember 3.0, but if it does, then we don't need this as the MouseEvent constructor works in IE11+. Probably the same will happen to KeyboardEvent

`@ember/test-helpers` does still support running on older browsers (it should support the full spectrum of browsers that the supported Ember versions support), which means that we will have to support IE9 and IE10 until we drop support for testing Ember 2.x (which will be quite a long time). However, we can try to detect that a given browser supports `new MouseEvent()` or `new KeyboardEvent()` and use the newer syntax whenever possible.

ggayowsky:
Closes #266 by using the `MouseEvent` constructor in `buildMouseEvent` if it is present. Otherwise, it falls back to the existing `initMouseEvent` function for compatibility with IE 9/10

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fire-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/fire-event.js
@@ -84,27 +84,39 @@ function buildBasicEvent(type, options = {}) {
 // eslint-disable-next-line require-jsdoc
 function buildMouseEvent(type, options = {}) {
   let event;
+  let eventOpts = merge(merge({}, DEFAULT_EVENT_OPTIONS), options);
   try {
-    event = document.createEvent('MouseEvents');
-    let eventOpts = merge(merge({}, DEFAULT_EVENT_OPTIONS), options);
-    event.initMouseEvent(
-      type,
-      eventOpts.bubbles,
-      eventOpts.cancelable,
-      window,
-      eventOpts.detail,
-      eventOpts.screenX,
-      eventOpts.screenY,
-      eventOpts.clientX,
-      eventOpts.clientY,
-      eventOpts.ctrlKey,
-      eventOpts.altKey,
-      eventOpts.shiftKey,
-      eventOpts.metaKey,
-      eventOpts.button,
-      eventOpts.relatedTarget
-    );
+    event = new MouseEvent(type, eventOpts);
   } catch (e) {
+    // left intentionally blank
+  }
+
+  if (!event) {
+    try {
+      event = document.createEvent('MouseEvents');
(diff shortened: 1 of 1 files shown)


addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -84,27 +84,39 @@ function buildBasicEvent(type, options = {}) {
...
-      eventOpts.ctrlKey,
-      eventOpts.altKey,
-      eventOpts.shiftKey,
-      eventOpts.metaKey,
-      eventOpts.button,
-      eventOpts.relatedTarget
-    );
+    event = new MouseEvent(type, eventOpts);
   } catch (e) {
+    // left intentionally blank
```
rwjblue: Hmm, aren’t some errors valid here? E.g. when `MouseEvent` is present but you pass invalid options or something?

rwjblue: maybe we could make a constant in module scope to track if `MouseEvent‘ is functional, then remove the try/catch inline?

ggayowsky: Was following the pattern from buildKeyboardEvent, but I agree that the solution checking if MouseEvent exists is better

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -1,5 +1,15 @@
 import { merge } from '@ember/polyfills';
 
+// eslint-disable-next-line require-jsdoc
+function checkMouseEventExistence() {
+  try {
```
cibernox: can you memoize this?

ggayowsky: Do you have an example of where else Memoization it is done in this repo? Or another example I can emulate?

rwjblue: Something like:

```js

const MOUSE_EVENT_CONSTRUCTOR = (() => {
  try {
    new MouseEvent(‘test’);
    return true;
  } catch(e) {
    return false;
  }
})();

ggayowsky:
Bump?

cibernox:
It looks good to me. /cc @rwjblue

PR #388, opened 2018-06-20
[In progress] Create visible() test helper

GlynnJKW:
From [this](https://github.com/GoogleChrome/puppeteer/issues/2629) discussion on the puppeteer github, I found out about the intersection observer API and thought it would be a good fit for ember tests. I've completed the helper and manually tested it on a project of my own, but I still need to add automated tests that cover more scenarios.

### DIFF
--- a/API.md
+++ b/API.md
@@ -25,6 +25,7 @@
 -   [Wait Helpers](#wait-helpers)
     -   [waitFor](#waitfor)
     -   [waitUntil](#waituntil)
+    -   [visible](#visible)
     -   [settled](#settled)
     -   [isSettled](#issettled)
     -   [getSettledState](#getsettledstate)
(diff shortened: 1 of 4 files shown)


GlynnJKW:
I think that the display: none or visibility: hidden case you mentioned is the most common, and there is currently no helper for those - you would have to manually check the style of the element it is being set on. 
The case which I originally developed this for was a test of an automatic scroller, where when a button was clicked the page would scroll down to the appropriate element. In cases such as that (where there is no/little change to the dom), I think this is the most elegant solution.

Turbo87:
I tend to agree with @cibernox. It might make more sense to offer this test helper in a separate addon. I would also propose to change it to an `isVisible()` helper function, so that it can work in harmony with `waitUntil()` (e.g. `waitUntil(() => isVisible('.foo'));`).

GlynnJKW:
That's fair - I can see that since the use cases for this aren't as general it would be better as an addon.
However, I think making it a synchronous helper like isVisible defeats the purpose of using an intersection observer - the other way of doing it (comparing bounding client rects) [forces reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a).

PR #391, opened 2018-06-29
add numbers to keyFromKeyCode

### LINKED ISSUE #390: trigger-key-event cannot trigger a text input of a number
keyFromKeyCode doesn't have numbers in the list, so trigger-key-event cannot trigger a text input of a number

keyCodes are 48-57 for 0-9 respectively

localpcguy:
Ran into an issue where I needed the numbers to in the keycode list as well (height/weight form). 

Fixes #390


ro0gr:
Adding a number key test-case seems good

rwjblue:
Ya, definitely needs a test. 

Also, would like @cibernox and @Turbo87 to double check.

cibernox:
Looks good to me, it just needs tests.
Those keyCodes are for the number above the keyboard, do you know if the keycodes of the numpad keys are different? Just curiosity.

localpcguy:
The keyCodes for number pad are the same. I'll get on the test(s)

localpcguy:
Added numbers to the tests. Let me know if there are any more concerns.

localpcguy:
@cibernox I wanted to clarify re:keycodes between the top row and number pad:

The ASCII code (`event.which`) is the same between both, and that is what this codes uses as the keyCode. The `event.code` value IS different (at least on MacOS), showing `Numpad1` vs. `Digit1` for the `1` key for example. That won't affect this code as we as using the `event.key` value when this code references `keyCode`, not the `event.code` value. See http://keycode.info/ from @wesbos for a simple way to see key information.

rwjblue:
Thank you!

PR #392, opened 2018-06-30
Include selector in waitFor()'s default timeout message

gabrielgrant:
Reduces ambiguity if a timeout occurs in a test with multiple waitFor()s


gabrielgrant:
Looks like a problem with eslint: 

```
$ eslint *.js addon addon-test-support app config lib node-tests server tests test-support
/home/travis/build/emberjs/ember-test-helpers/addon-test-support/@ember/test-helpers/dom/wait-for.js
  18:33  error  Replace `⏎··selector,⏎··{·timeout·=·1000,·count·=·null,·timeoutMessage·}·=·{}⏎` with `selector,·{·timeout·=·1000,·count·=·null,·timeoutMessage·}·=·{}`  prettier/prettier
✖ 1 problem (1 error, 0 warnings)
```
Didn't change this spacing/formatting, so not sure why that wasn't causing problems before, but just added a commit to fix it

rwjblue:
Thank you!

PR #395, opened 2018-07-11
docs(api): fix get settled state items formatting

knownasilya:


Turbo87:
@knownasilya that documented is automatically generated using `yarn docs`, so your changed would be overwritten again soon unless you edit the source code comments

PR #396, opened 2018-07-11
docs(api): try fixing getSettledState list

knownasilya:
Not sure if this will fix it, since it seems the list doesn't even get created in the generated content.


rwjblue:
I think you can do `yarn docs` to regenerate the API docs (so we can see if it fixes it)...

rwjblue:
Thank you!

PR #397, opened 2018-07-14
Add `typeIn` helper to trigger keyup, keypress and keyup events when filling inputs

mfeckie:
The current implementation of `fillIn` does not trigger keydown, keypress and keyup events.

I've recently been working on some stuff that uses these events.  One example being a formatted text input for bank account numbers that only allow certain characters.

I'm not sure if other folks would find this helpful so submitting as a PR for consideration / discussion.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/type-in.js
+++ b/addon-test-support/@ember/test-helpers/dom/type-in.js
@@ -1,0 +1,98 @@
+import { nextTickPromise } from '../-utils';
+import settled from '../settled';
+import getElement from './-get-element';
+import isFormControl from './-is-form-control';
+import { __focus__ } from './focus';
+import { Promise } from 'rsvp';
+import fireEvent from './fire-event';
+
+/**
+ * Mimics character by character entry into the target `input` or `textarea` element.
+ *
+ * Allows for simulation of slow entry by passing an optional millisecond delay
+ * between key events.
+
+ * The major difference between `typeIn` and `fillIn` is that `typeIn` triggers
+ * keyboard events as well as `input` and `change`.
+ * Typically this looks like `focus` -> `focusin` -> `keydown` -> `keypress` -> `keyup` -> `input` -> `change`
+ * per character of the passed text (this may vary on some browsers).
+ *
+ * @public
+ * @param {string|Element} target the element or selector to enter text into
+ * @param {string} text the test to fill the element with
+ * @param {Object} options {delay: x} (default 50) number of milliseconds to wait per keypress
+ * @return {Promise<void>} resolves when the application is settled
+ */
+export default function typeIn(target, text, options = { delay: 50 }) {
+  return nextTickPromise().then(() => {
+    if (!target) {
+      throw new Error('Must pass an element or selector to `typeIn`.');
+    }
+
+    const element = getElement(target);
+    if (!element) {
+      throw new Error(`Element not found when calling \`typeIn('${target}')\``);
+    }
+    let isControl = isFormControl(element);
+    if (!isControl) {
+      throw new Error('`typeIn` is only usable on form controls.');
+    }
+
+    if (typeof text === 'undefined' || text === null) {
+      throw new Error('Must provide `text` when calling `typeIn`.');
+    }
+
+    __focus__(element);
+
+    return fillOut(element, text, options.delay)
+      .then(() => fireEvent(element, 'change'))
+      .then(settled);
+  });
+}
+
+// eslint-disable-next-line require-jsdoc
+function fillOut(element, text, delay) {
+  return new Promise(resolve => {
+    executeEvents(element, text, delay).then(resolve);
+  });
+}
+
+// eslint-disable-next-line require-jsdoc
+function keyEntry(element, character) {
+  const charCode = character.charCodeAt();
+
+  const eventOptions = {
+    bubbles: true,
+    cancellable: true,
+    charCode,
+  };
+
+  const keyEvents = {
+    down: new KeyboardEvent('keydown', eventOptions),
+    press: new KeyboardEvent('keypress', eventOptions),
+    up: new KeyboardEvent('keyup', eventOptions),
+  };
+
+  return function() {
+    element.dispatchEvent(keyEvents.down);
+    element.dispatchEvent(keyEvents.press);
+    element.value = element.value + character;
+    fireEvent(element, 'input');
+    element.dispatchEvent(keyEvents.up);
+  };
+}
+
+// eslint-disable-next-line require-jsdoc
+function executeEvents(element, text, delay) {
+  const inputFunctions = text.split('').map(character => keyEntry(element, character, delay));
+  return inputFunctions.reduce((currentPromise, func) => {
+    return currentPromise.then(() => delayedExecute(func, delay));
+  }, Promise.resolve());
+}
+
+// eslint-disable-next-line require-jsdoc
+function delayedExecute(func, delay) {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  }).then(func);
+}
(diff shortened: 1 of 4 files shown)


mfeckie:
How about `fillUsingKeyEvents` ?

cibernox:
My preference is "typeIn"

mfeckie:
:heart	Less verbose and says what it does :) will use

mfeckie:
Looking at the codebase, it seems that `async/await` is explicitly disabled, which would make cleanly implementing a delay between each key input difficult.  Suggestions welcome.

Turbo87:
@mfeckie you can use promise chains instead

mfeckie:
OK folks, I've updated the PR to address all the feedback and suggestions so far.

I've also updated the documentation to go along with it.

mfeckie:
Any updates on the this folks?

tests/unit/dom/type-in-test.js
```diff
@@ -0,0 +1,133 @@
+import { module, test } from 'qunit';
+import { typeIn, setupContext, teardownContext } from '@ember/test-helpers';
+import { buildInstrumentedElement } from '../../helpers/events';
+import { isIE11 } from '../../helpers/browser-detect';
+import hasEmberVersion from 'ember-test-helpers/has-ember-version';
+
+let expectedEvents = [
+  'focus',
```
cibernox: After some investigation, I think the order of the events may be wrong. Check this jsbin: https://jsbin.com/zitazuxabe/edit?html,js,console,output

Seems that the sequence when typing "abc" is:
```
keydown | e.target.value is ""
keypress | e.target.value is ""
input | e.target.value is "a"
keyup | e.target.value is "a"
keydown | e.target.value is " a"
keypress | e.target.value is "a"
input | e.target.value is "ab"
keyup | e.target.value is "ab"
keydown | e.target.value is " ab"
keypress | e.target.value is "ab"
input | e.target.value is "abc"
keyup | e.target.value is "abc"
change | e.target.value is "abc"
```

Changes I request:

- `input` fires before `keyup`
- It's important that `keydown` and `keypress` are fired **before** the input gets the new value, and the rest are fired after
- The `change` event should only be fired once at the end.

cibernox:
@rwjblue how do you feel about this helper once it has been polished?

tests/unit/dom/type-in-test.js
mfeckie: Done

cibernox:
Code-wise, I think this is good. I'll defer to @rwjblue to decide wether we want this helper in this repo or in another one, but sounds good.

rwjblue:
Conceptually, I think it is a good helper to have. I haven't had much time to dig into the implementation just yet though (sorry).

mfeckie:
Anything I can do to help progress this good humans?

API.md
```diff
@@ -3,14 +3,15 @@
 ### Table of Contents
 
 -   [DOM Interaction Helpers](#dom-interaction-helpers)
+    -   [blur](#blur)
```
rwjblue: Can you reset the changes here to `API.md`? I'll regenerate it before publishing (this way the API.md on master roughly continues to match what actual users are using)...

addon-test-support/@ember/test-helpers/dom/type-in.js
```diff
@@ -0,0 +1,98 @@
...
+ * Typically this looks like `focus` -> `focusin` -> `keydown` -> `keypress` -> `keyup` -> `input` -> `change`
+ * per character of the passed text (this may vary on some browsers).
+ *
+ * @public
+ * @param {string|Element} target the element or selector to enter text into
+ * @param {string} text the test to fill the element with
+ * @param {number} millisecondDelay number of millisecons to wait per keypress
+ * @return {Promise<void>} resolves when the application is settled
+ */
+export default function typeIn(target, text, millisecondDelay = 0) {
```
rwjblue: Lets go with an `{ delay: 50 }` for the third arg

addon-test-support/@ember/test-helpers/dom/type-in.js
```diff
@@ -0,0 +1,98 @@
...
+
+    return fillOut(element, text, millisecondDelay)
+      .then(() => fireEvent(element, 'change'))
+      .then(settled);
+  });
+}
+
+// eslint-disable-next-line require-jsdoc
+function fillOut(element, text, delay) {
+  return new Promise(resolve => {
```
rwjblue: I don't understand this promise nesting

addon-test-support/@ember/test-helpers/dom/type-in.js
```diff
@@ -0,0 +1,98 @@
...
+      throw new Error('`typeIn` is only usable on form controls.');
+    }
+
+    if (typeof text === 'undefined' || text === null) {
+      throw new Error('Must provide `text` when calling `typeIn`.');
+    }
+
+    __focus__(element);
+
+    return fillOut(element, text, millisecondDelay)
```
rwjblue: Any reason we can't use `executeEvents` directly here? I don't see what `fillOut` provides us...

```js
return executeEvents(element, text, delay).then(() => fireEvent(element, 'change'))
```

addon-test-support/@ember/test-helpers/dom/type-in.js
```diff
@@ -0,0 +1,98 @@
...
+    let isControl = isFormControl(element);
+    if (!isControl) {
+      throw new Error('`typeIn` is only usable on form controls.');
+    }
+
+    if (typeof text === 'undefined' || text === null) {
+      throw new Error('Must provide `text` when calling `typeIn`.');
+    }
+
+    __focus__(element);
```
rwjblue: Does `__focus__` return a promise that we need to chain?

tests/unit/dom/type-in-test.js
rwjblue: Can you add this JSBin as an inline comment? It sure will make "reality" testing easier over time...

addon-test-support/@ember/test-helpers/dom/type-in.js
mfeckie: done

API.md
mfeckie: done

addon-test-support/@ember/test-helpers/dom/type-in.js
mfeckie: This is exactly the same as I found it in `fill-in.js` so I kept for consistency

addon-test-support/@ember/test-helpers/dom/type-in.js
mfeckie: Fill out packages up the promise tree.

addon-test-support/@ember/test-helpers/dom/type-in.js
mfeckie: The reason for the promise nesting is because `fillOut` takes each step and executes `keyDown, keyPress, input and keyUp` events for each character in the input string.  

We need to ensure that each promise is triggered in order.

Without this, everything essentially happens without delay.

mfeckie: Yep, your right.  Have removed.  Sorry for misunderstanding the original question!

makepanic:
`fillIn` supports [contenteditable as target elements](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/fill-in.js#L29).
Is it possible to add this to the `typeIn` helper too?

mfeckie:
I specifically excluded `contenteditable` as I don't believe it follows the same event pattern as native inputs.

Perhaps you could take a look after / if this PR gets merged?

mfeckie:
@rwjblue I think I've addressed all your feedback?  Anything outstanding?

mfeckie:
Bump

PR #398, opened 2018-07-14
add options to click, doubleClick, tap

BryanCrotaz:
Add options to public api which are passed down to the relevant MouseEvent and TouchEvent

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/click.js
+++ b/addon-test-support/@ember/test-helpers/dom/click.js
@@ -10,15 +10,15 @@ import isFormControl from './-is-form-control';
   @private
   @param {Element} element the element to click on
 */
-export function __click__(element) {
-  fireEvent(element, 'mousedown');
+export function __click__(element, options) {
+  fireEvent(element, 'mousedown', options);
 
   if (isFocusable(element)) {
     __focus__(element);
   }
 
-  fireEvent(element, 'mouseup');
-  fireEvent(element, 'click');
+  fireEvent(element, 'mouseup', options);
+  fireEvent(element, 'click', options);
 }
 
 /**
@@ -49,7 +49,7 @@ export function __click__(element) {
   @param {string|Element} target the element or selector to click on
   @return {Promise<void>} resolves when settled
 */
-export default function click(target) {
+export default function click(target, options = {}) {
   return nextTickPromise().then(() => {
     if (!target) {
       throw new Error('Must pass an element or selector to `click`.');
@@ -63,7 +63,7 @@ export default function click(target) {
     let isDisabledFormControl = isFormControl(element) && element.disabled === true;
 
     if (!isDisabledFormControl) {
-      __click__(element);
+      __click__(element, options);
     }
 
     return settled();
--- a/addon-test-support/@ember/test-helpers/dom/double-click.js
+++ b/addon-test-support/@ember/test-helpers/dom/double-click.js
@@ -9,19 +9,19 @@ import { nextTickPromise } from '../-utils';
   @private
   @param {Element} element the element to double-click on
 */
-export function __doubleClick__(element) {
-  fireEvent(element, 'mousedown');
+export function __doubleClick__(element, options) {
+  fireEvent(element, 'mousedown', options);
 
   if (isFocusable(element)) {
     __focus__(element);
   }
 
-  fireEvent(element, 'mouseup');
-  fireEvent(element, 'click');
-  fireEvent(element, 'mousedown');
-  fireEvent(element, 'mouseup');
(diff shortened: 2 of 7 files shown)


BryanCrotaz:
@rwjblue build failed with a curl error - can't see how to rerun tests - are you able to?

rwjblue:
Restarted

addon-test-support/@ember/test-helpers/dom/click.js
```diff
@@ -9,16 +9,17 @@ import isFormControl from './-is-form-control';
...
+export function __click__(element, options) {
+  fireEvent(element, 'mousedown', options);
 
   if (isFocusable(element)) {
     __focus__(element);
   }
 
-  fireEvent(element, 'mouseup');
-  fireEvent(element, 'click');
+  fireEvent(element, 'mouseup', options);
```
rwjblue: It seems really odd to me to have the same options for mousedown, mouseup, and click 🤔

addon-test-support/@ember/test-helpers/dom/double-click.js
```diff
@@ -8,20 +8,21 @@ import { nextTickPromise } from '../-utils';
...
-  fireEvent(element, 'mousedown');
-  fireEvent(element, 'mouseup');
-  fireEvent(element, 'click');
-  fireEvent(element, 'dblclick');
+  fireEvent(element, 'mouseup', options);
+  fireEvent(element, 'click', options);
+  fireEvent(element, 'mousedown', options);
+  fireEvent(element, 'mouseup', options);
+  fireEvent(element, 'click', options);
+  fireEvent(element, 'dblclick', options);
```
rwjblue: Same concern here. If we are going to allow event options I think we need a system that allows us to customize each of the events individually...

@mmun and I had discussed this somewhat recently, but I can’t recall if we came to a good conclusion 🤔

addon-test-support/@ember/test-helpers/dom/click.js
BryanCrotaz: The common options are going to be button, x, y. Which options would you want to be different for each event?

sdhull:
👋 Hi!  Absolutely love that you're working on this @BryanCrotaz, thank you!

I'm not sure what @rwjblue has in mind in terms of customizing "each of the events individually." What has been done in this PR makes perfect sense to me -- that is to say, you pass options through to `fireEvent`, which then constructs the event with some default options & merges the passed options in. That's what these helpers _should_ do IMO.

And specifically for [`mousedown`](https://developer.mozilla.org/en-US/docs/Web/Events/mousedown), [`mouseup`](https://developer.mozilla.org/en-US/docs/Web/Events/mouseup) and [`click`](https://developer.mozilla.org/en-US/docs/Web/Events/click), these events all appear to have the same properties according to MDN.

Perhaps @rwjblue is concerned with developers passing an option like `charCode` to a click event? Or `clientX` to a keyboard event? Personally I wouldn't worry about that in _this_ PR.

Maybe in a followup PR we could add some checks to warn developers they're trying to use options that don't make sense, but for simply enabling a developer to specify specific coordinates in a click (or double-click or tap), I would urge maintainers to merge this to enable ambitious testing of our ambitious apps 😎

---
tl;dr -- 🙏 please merge this and ship a patch release so I can get off of @BryanCrotaz's branch 😬

mmun:
@rwjblue This PR seems ok to me as is. My main concern would be passing the event options to unrelated kinds of events (e.g. passing mouse options to the focus event).

You could argue that the mousedown and mouseup events could have different options (if you move your mouse during the press), however even then I think it just makes more sense to manually trigger each event individually in your test.

tl;dr I think it's good as is.

BryanCrotaz:
I agree. The method is to simulate a click. If you want to simulate down-drag-up you wouldn’t use it - you’d definitely simulate individual events.

cibernox:
I think it makes sense to pass the same options for both mousedown/mouseup and click. For instance, if the user is pressing shift, that key will be pressed for all three events.

It is true that there _might_ be some situation where someone might want to simulate a `mousedown` in one x/y coordinate and a mouseup in a different coordinate, but in those situation I'd expect the developer to resort to more lower-level approach like

```hbs
await triggerEvent('mousedown', { screenX: 1, screenY: 1 } );
await triggerEvent('mouseup', { screenX: 1, screenY: 5 } );
```

addon-test-support/@ember/test-helpers/dom/double-click.js
bendemboski: Maybe allow `options` to either be a POJO or a function, and if a function, gets invoked with the name of the event and its return value is the options POJO?

PR #399, opened 2018-07-21
Standarise the `fireEvent` method interface

khalilovcmd:
## Why?

When using `triggerEvent` to fire a `change` event on `<input> element with type="file"`, [those lines](https://github.com/emberjs/ember-test-helpers/blob/8a3614657064efaf13dfe667fe4cfdc14f15f509/addon-test-support/@ember/test-helpers/dom/fire-event.js#L57-L58) call the `buildFileEvent` which forwards `options` with its same structure.

Though, `buildFileEvent` [accepts an array](https://github.com/emberjs/ember-test-helpers/blob/8a3614657064efaf13dfe667fe4cfdc14f15f509/addon-test-support/@ember/test-helpers/dom/fire-event.js#L183): `files = []`, the documentation doesn't clarify this difference.

I think it will be more natural not to break the interface here by introducing an array argument.

## Sample code

```js
let file = new File(['my-content'],'image.png', {
  type: 'image/png',
});
triggerEvent(selector, 'change', [file]); // will work
triggerEvent(selector, 'change',{ files: file }); // will not propagate the value to the event handler
```

## Questions

1. Should we start writing a test for this file? It think it is a sensitive one to skip tests on.

## Alternatives

We could also live with elaborating on this difference in the `API.md`, to avoid breaking changes.

----

cc: kudos to @mattdonnelly for pointing this out, and subsequently saving my day.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fire-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/fire-event.js
@@ -180,8 +180,9 @@ function buildKeyboardEvent(type, options = {}) {
 }
 
 // eslint-disable-next-line require-jsdoc
-function buildFileEvent(type, element, files = []) {
+function buildFileEvent(type, element, options = {}) {
   let event = buildBasicEvent(type);
+  let files = Array.isArray(options) ? options : options.files || [];
 
   if (files.length > 0) {
     Object.defineProperty(files, 'item', {
--- a/addon-test-support/@ember/test-helpers/dom/trigger-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/trigger-event.js
@@ -4,14 +4,39 @@ import settled from '../settled';
 import { nextTickPromise } from '../-utils';
 
 /**
-  Triggers an event on the specified target.
-
-  @public
-  @param {string|Element} target the element or selector to trigger the event on
-  @param {string} eventType the type of event to trigger
-  @param {Object} options additional properties to be set on the event
-  @return {Promise<void>} resolves when the application is settled
-*/
+ * Triggers an event on the specified target.
+ *
+ * For HTML `input` element of type 'file', if you wish to trigger `change`
+ * event, the options param can be passed as:
+ *   1. an array of `File` objects.
+ *   2. an object with key `files` which contains an array of `File` objects.
+ *
+ * @public
+ * @param {string|Element} target the element or selector to trigger the event on
+ * @param {string} eventType the type of event to trigger
+ * @param {Object} options additional properties to be set on the event
+ * @return {Promise<void>} resolves when the application is settled
+ *
+ * @example <caption>Usage</caption>
+ *
+ * import { setupRenderingTest } from 'ember-qunit';
+ * import { render, triggerEvent } from '@ember/test-helpers';
+ *
+ * module('awesome-sauce', function(hooks) {
+ *   setupRenderingTest(hooks);
+ *     test('does something awesome', async function(assert) {
+ *     await render(hbs`{{awesome-sauce}}`);
+ *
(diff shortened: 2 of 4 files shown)


addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -180,8 +180,9 @@ function buildKeyboardEvent(type, options = {}) {
 }
 
 // eslint-disable-next-line require-jsdoc
-function buildFileEvent(type, element, files = []) {
+function buildFileEvent(type, element, options = {}) {
   let event = buildBasicEvent(type);
+  let files = options.files || [];
```
rwjblue: As written, this is a breaking change. As mentioned in the PR description, anyone using `triggerEvent` for this today _must_ pass an array, if we are going to change this to `{ files: someArray }` as you suggest here, we definitely need to detect if `options` is already an array...

khalilovcmd: > if we are going to change this to { files: someArray } as you suggest here, we definitely need to detect if options is already an array...

@rwjblue that makes sense. My first thought is to disambiguate this in the documentation itself, but then as a novice user of this add on, I would have expected that the interface to accept `{ files: someArray }`, hence why I felt it might be a more natural (but breaking change) step to standarize it.

khalilovcmd:
Thank you @rwjblue for taking a look.

To summarise, I could take these two actions, as you suggested:

> Can you add additional examples to those inline docs to show how to trigger change on a file input?

1. I will add more examples to the documentation of the `triggerEvent` to elaborate on how to trigger a `change` event for a file.
2. Detect if `options` is already an array.

Does this sound like a good way to move forward?

rwjblue:
Yep, sounds good to me!

khalilovcmd:
@rwjblue does it make sense the way it looks now?

addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -180,8 +180,9 @@ function buildKeyboardEvent(type, options = {}) {
 }
 
 // eslint-disable-next-line require-jsdoc
-function buildFileEvent(type, element, files = []) {
+function buildFileEvent(type, element, options = {}) {
   let event = buildBasicEvent(type);
+  let files = Array.isArray(options) ? options : options.files || [];
```
rwjblue: Should we issue a deprecation or something when we do this coercion?

I'm just thinking that over time we _probably_ only want to support one form...

addon-test-support/@ember/test-helpers/dom/trigger-event.js
```diff
@@ -4,14 +4,39 @@ import settled from '../settled';
...
+ * import { render, triggerEvent } from '@ember/test-helpers';
+ *
+ * module('awesome-sauce', function(hooks) {
+ *   setupRenderingTest(hooks);
+ *     test('does something awesome', async function(assert) {
+ *     await render(hbs`{{awesome-sauce}}`);
+ *
+ *     // passing an array of files
+ *     let file = new File(['text file'], 'text.txt', { type: 'text/plain' });
+ *     await triggerEvent('input:file', 'change', [file]);
```
rwjblue: I think ultimately, we should only have one format here (probably the `{ files: [] }` one?)...

rwjblue:
@khalilovcmd - Had a chance to review those questions?

rwjblue:
Will need to address merge conflicts from docs changes made in https://github.com/emberjs/ember-test-helpers/pull/340

rwjblue:
@khalilovcmd - I'd still love to land this, any chance you mind have time to rebase and clean up soon?

PR #404, opened 2018-08-06
Fix getRootElement failing when Application.rootElement is an element

### LINKED ISSUE #403: getRootElement() fails when rootElement is an element
The `rootElement` documentation states it can be an element or a selector string, but `getRootElement()` only works with selector strings.

https://emberjs.com/api/ember/3.3/classes/Application/properties/rootElement?anchor=rootElement

We were testing out shadow DOM stuff in chrome and using an initializer to create the shadow DOM and then point `rootElement` to an element inside the shadow DOM. Unfortunately tests don't work because get-root-element.js only works when `rootElement` is a string that can be passed to a querySelector.

https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/get-root-element.js#L21

If it's useful I can build an Ember Twiddle showing the crash

apellerano-pw:
Fixes #403 

I used the same element detection logic that -get-element.js uses.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/get-root-element.js
+++ b/addon-test-support/@ember/test-helpers/dom/get-root-element.js
@@ -14,16 +14,24 @@ export default function getRootElement() {
     throw new Error('Must setup rendering context before attempting to interact with elements.');
   }
 
-  let rootElementSelector;
+  let rootElement;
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome, thank you!

PR #410, opened 2018-09-11
Use `assign` instead of deprecated `merge`

wagenet:


wagenet:
@cibernox doesn't look like the failing test is my fault, but can you verify?

rwjblue:
No, we need to remove `^` from ember-try in package.json

wagenet:
So is this good to merge?

rwjblue:
Rebased and updated to add ember-assign-polyfill as a dependency.

PR #427, opened 2018-10-13
Drop support for Node.js 4 and below

Turbo87:
Unblocks several dependency updates. This is a breaking change and requires a major version bump! (or a minor bump in this case as we're still pre-1.0)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "4"
+  - "6"
 
 sudo: false
 dist: trusty
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "qunit": "^2.5.1"
   },
   "engines": {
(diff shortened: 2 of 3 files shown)


Turbo87:
> unlock the ember-data version in `package.json`?

what do you mean by that? `ember-source` and `ember-cli` are also at v3.1.x currently and updating them to something else seems not directly related to dropping Node 4 support 🤔

rwjblue:
I was referring to the resolutions field in package.json (where I had added custom resolution overrides to avoid various Node 4 breaking packages) and the fact that I was forced to pin ember-data to a specific 2.18 release because later patch releases broke node 4 compatibility.

Seems _OK_ if you really don’t want to update them here, but they were only pinned due to Node 4 issues so it seemed reasonable to undo in a Node 4 dropping PR to me...

rwjblue:
Ya, seems ok, I’m just saying that the _reason_ I pinned it was due to node 4 issues. I’ll fix it in a follow up.

PR #428, opened 2018-10-15
Standardize params for calling triggerEvent on file input

mattdonnelly:
#### Why?

Following on from the feedback in this PR: https://github.com/emberjs/ember-test-helpers/pull/399

#### How?

I've updated `triggerEvent` method to accept an `options` param as an object containing a key `files` and issue a deprecation warning when passed an array

---

Closes #399

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fire-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/fire-event.js
@@ -1,4 +1,5 @@
 import { assign } from '@ember/polyfills';
+import { deprecate } from '@ember/application/deprecations';
 
 // eslint-disable-next-line require-jsdoc
 const MOUSE_EVENT_CONSTRUCTOR = (() => {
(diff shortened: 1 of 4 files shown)


addon-test-support/@ember/test-helpers/dom/fire-event.js
```diff
@@ -194,8 +195,23 @@ function buildKeyboardEvent(type, options = {}) {
...
+function buildFileEvent(type, element, options = {}) {
   let event = buildBasicEvent(type);
+  let files;
+  if (Array.isArray(options)) {
+    deprecate(
+      'Passing the `options` param as an array to `triggerEvent` for file inputs is deprecated. Please pass an object with a key `files` containing the array instead.',
+      false,
+      {
+        id: 'ember-test-helpers.trigger-event.options-blob-array',
+        until: '0.8.0',
```
mattdonnelly: Not sure what the process here is so I just went with the next major version

rwjblue: Yep, this seems good.

PR #429, opened 2018-10-16
Ensure view is set by default on generated mouse events

### LINKED ISSUE #426: Upgrade to 0.7.26 breaks d3-drag
After upgrading from 0.7.25, our tests started failing. It was really difficult to track down, but it seems to be related to changes in this addon.

Basically, we started getting a lot of errors like this: 

`Uncaught TypeError: Cannot read property \'document\' of null at http://localhost:7357/assets/vendor.js, line 78363'`

This seems to be tied to the `click` helper and https://github.com/d3/d3-drag.

Which somehow failed here: (it's hard to say exactly since this was always minified in our app, but we tried to trace the steps as good as possible): https://github.com/d3/d3-drag/blob/ec20d5e35bfc6ad2b4266494aef475c94cc2e6c3/src/nodrag.js#L16

Which is called here: https://github.com/d3/d3-drag/blob/ec20d5e35bfc6ad2b4266494aef475c94cc2e6c3/src/drag.js#L74  (or, I guess it could also have been this code: https://github.com/d3/d3-drag/blob/ec20d5e35bfc6ad2b4266494aef475c94cc2e6c3/src/drag.js#L55)

Basically, our test just did this:

```
await click('[data-test-d3-circle-element]');
```

Which lead it to blow up. It works if I do it manually, it seems the `click` listener has somehow changed to break d3. This used to work fine in 0.7.25.

mydea:
This PR adds `window` as default `view` on generated mouse events.

This fixes #426

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fire-event.js
+++ b/addon-test-support/@ember/test-helpers/dom/fire-event.js
@@ -93,7 +93,7 @@ function buildBasicEvent(type, options = {}) {
 // eslint-disable-next-line require-jsdoc
 function buildMouseEvent(type, options = {}) {
   let event;
-  let eventOpts = assign({}, DEFAULT_EVENT_OPTIONS, options);
+  let eventOpts = assign({ view: window }, DEFAULT_EVENT_OPTIONS, options);
   if (MOUSE_EVENT_CONSTRUCTOR) {
     event = new MouseEvent(type, eventOpts);
   } else {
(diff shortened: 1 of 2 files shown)


cibernox:
I had no idea the view property was important, I never saw any code that failed because of that.
LGTM

mydea:
I guess D3 is a bit special ;)

cibernox:
The failure looks random. Can you amend and push again?

rwjblue:
@ember/test-helpers 0.7.27 published 🎉

PR #448, opened 2018-11-01
CI: Use `name` property instead of `NAME` env var workaround

Turbo87:
didn't know this existed, but apparently ember-data is using this successfully... :)


rwjblue:
Hehe. It didn’t exist when we added the env var (early in the beta cycle). It was added just as stages went from beta -> stable.

PR #371, opened 2018-05-07
Add TypeScript declaration file.

jamescdavis:
This adds a TypeScript type declarations file based on https://github.com/emberjs/ember-test-helpers/blob/master/API.md

PR #450, opened 2018-11-01
Convert to TypeScript

Turbo87:
This PR starts the conversion of this addon towards TypeScript

Closes #371 

/cc @chriskrycho @mike-north @dfreeman


chriskrycho:
One quick note here (more later): depending on the sense of urgency you feel here, it may be worth waiting till ember-cli-typescript 2 is out of beta: at that point you won’t need to use the internals, and things will Just Work™. No worries either way; it was just the one bit that caught my attention on first skim.

(We don’t have a definite timeline for that, but I’m going to publish a blog post about it to get some more eyes on it; initial signs including from our largest app are very good.)

dfreeman:
The need for the `move-declaration-files` script isn't something we have a good answer for yet (in 1.x or the 2.x betas), though it's something I've been thinking about trying to address soon, as TypeScript is a good fit for lots of testing-oriented addons, but those addons also have a tendency to do things with path remapping.

I think there's a good chance we can look at the `paths` in `tsconfig.json` to be more intelligent about where we move declarations prior to publish, rather than baking in knowledge of "normal" addon layout. That would allow us to make sure all the `@ember/test-helpers` declarations are in the right place automatically.

Turbo87:
@rwjblue I've trimmed this down to the most important commits, and I think it's good to go. I got several follow-up changes lined up already and will open PRs for them once this one is merged.

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -158,8 +162,8 @@ export function getSettledState() {
   let pendingRequestCount = pendingRequests();
 
   return {
-    hasPendingTimers: Boolean(run.hasScheduledTimers()),
-    hasRunLoop: Boolean(run.currentRunLoop),
+    hasPendingTimers: Boolean((run as any).hasScheduledTimers()),
```
rwjblue: Doesn't `Ember.run` have stuff in `@types/ember`? Why do we have to cast here?

Turbo87: it does, but unfortunately those two are missing from the published types 😢

mike-north: @Turbo87 @rwjblue The only reason they're missing is because [there's no indication that they're part of Ember's public API](https://github.com/emberjs/ember.js/blob/d277c93bc0d357ff50e2c62ec1853794a9b82dd1/packages/%40ember/runloop/index.js#L337-L339). We look to the https://www.emberjs.com/api/ as the source of truth.

rwjblue: Ya, seems fine though this library (IMHO by necessity) will have to use what amount to private APIs for applications so that it can absorb any private API churn and expose _actually_ public stable APIs to our users.

mike-north: > Ya, seems fine though this library (IMHO by necessity) will have to use what amount to private APIs for applications so that it can absorb any private API churn and expose _actually_ public stable APIs to our users.

In the future, we may want to discuss the pros/cons of publishing an additional types package for authors of libraries like this, to layer on top of the public ember types. I have some quality/stability concerns around each author aiming to define/maintain this type info themselves.

rwjblue: Yeah. The issue is somewhat complex, the layers of “public” are quite different. For example:

* intended for consumption by apps
* intended for consumption by tooling
* intended for consumption by “privileged” addons

Things like the modifier and component manager APIs are good examples of things intended for addons but not apps. 

Things like default outlet templates / using outlet state / etc are examples of things Ember intentionally exposes for this addon to be able to provide its APIs, but those things should never be made exposed to other addons or apps...

jamescdavis:
This is a great start! I was going to comment on places where the types could be tightened up, but I see you're already on it in #452.

PR #452, opened 2018-11-01
Import type definitions from `@types/ember__test-helpers`

Turbo87:
This PR imports the type definitions from the `@types/ember__test-helpers` npm package, and applies them to our own TypeScript files.

Further cleanup of those types will happen in follow-up PRs.

### DIFF
--- a/addon-test-support/@ember/test-helpers/-utils.ts
+++ b/addon-test-support/@ember/test-helpers/-utils.ts
@@ -78,7 +78,7 @@ export const futureTick = setTimeout;
  @private
  @returns {Promise<void>} Promise which can not be forced to be ran synchronously
 */
-export function nextTickPromise() {
+export function nextTickPromise(): RSVP.Promise<void> {
   // Ember 3.4 removed the auto-run assertion, in 3.4+ we can (and should) avoid the "psuedo promisey" run loop configuration
   // for our `nextTickPromise` implementation. This allows us to have real microtask based next tick timing...
   if (hasEmberVersion(3,4)) {
--- a/addon-test-support/@ember/test-helpers/application.ts
+++ b/addon-test-support/@ember/test-helpers/application.ts
@@ -1,6 +1,8 @@
+import Application from '@ember/application';
+
 import { getResolver, setResolver } from './resolver';
 
-var __application__;
+var __application__: Application;
 
 /**
   Stores the provided application instance so that tests being ran will be aware of the application under test.
@@ -11,11 +13,11 @@ var __application__;
   @public
   @param {Ember.Application} application the application that will be tested
 */
-export function setApplication(application) {
+export function setApplication(application: Application): void {
   __application__ = application;
 
   if (!getResolver()) {
-    let Resolver = application.Resolver;
+    let Resolver = (application as any).Resolver;
     let resolver = Resolver.create({ namespace: application });
 
     setResolver(resolver);
@@ -28,6 +30,6 @@ export function setApplication(application) {
   @public
   @returns {Ember.Application} the previously stored application instance under test
 */
-export function getApplication() {
+export function getApplication(): Application {
   return __application__;
 }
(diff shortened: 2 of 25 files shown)


jamescdavis:
Can we tighten up and standardize `Context` to extend `TestContext` from the `ember-test-helpers` typedefs?:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f8721b6253475d6f8267a9cc16c5bde6089314c6/types/ember-test-helpers/index.d.ts#L36-L63

Turbo87:
Yeah, I have something like this already in a follow up commit that is not part of this PR

Turbo87:
@rwjblue any thoughts on this?

addon-test-support/@ember/test-helpers/-utils.ts
```diff
@@ -9,7 +9,7 @@ export const futureTick = setTimeout;
  @private
  @returns {Promise<void>} Promise which can not be forced to be ran synchronously
 */
-export function nextTickPromise() {
+export function nextTickPromise(): RSVP.Promise<void> {
```
rwjblue: Hmm, this seems odd, why can’t it be `RSVPPromise<void>`?

addon-test-support/@ember/test-helpers/application.ts
```diff
@@ -11,11 +13,11 @@ var __application__;
   @public
   @param {Ember.Application} application the application that will be tested
 */
-export function setApplication(application) {
+export function setApplication(application: Application): void {
   __application__ = application;
 
   if (!getResolver()) {
-    let Resolver = application.Resolver;
+    let Resolver = (application as any).Resolver;
```
rwjblue: Is there no type for `Ember.Resolver`?

addon-test-support/@ember/test-helpers/dom/-target.ts
```diff
@@ -0,0 +1,3 @@
+type Target = string | Element;
```
rwjblue: I think there is actually an `EventTarget` type, is that better here?

addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -139,8 +144,9 @@ export default function(context, options: any = {}) {
     .then(() => {
       let application = getApplication();
       if (application) {
-        return application.boot();
+        return application.boot().then(() => {});
```
rwjblue: This seems somewhat odd, are we just trying to make sure it doesn’t return something?

addon-test-support/@ember/test-helpers/-utils.ts
rwjblue: Also, I’d prefer to change this to native Promise...

Turbo87: > why can’t it be `RSVPPromise<void>`?

I could, but I just rebased and now everywhere else we also use `RSVP.Promise`

> I’d prefer to change this to native Promise...

me too, but iirc somewhere we use `.finally()`, which means we must use RSVP here until we refactor that away

addon-test-support/@ember/test-helpers/application.ts
Turbo87: there is, and I have commits in the queue that introduce it

addon-test-support/@ember/test-helpers/dom/-target.ts
Turbo87: `Target` in this case means "what you can pass into e.g. `click()`", not the `target` property of an `Event`

addon-test-support/@ember/test-helpers/setup-context.ts
Turbo87: yes, the API contract is that we return `void`. we could change it to `Promise<any>` but I'm not sure if that's a good idea 🤔

rwjblue: Yep, totally agree with you.

addon-test-support/@ember/test-helpers/-utils.ts
rwjblue: Ok, I’m happy to work on removing that finally after this lands. At least we have some TS flow analysis support to track down where that happens now....

PR #457, opened 2018-11-02
Improve teardownContext tests

wagenet:
This should fail when running against newer version of Ember due to an
apparent regression introduced in Ember 3.2.


wagenet:
The question now is do we remove the new test and the XHR semaphore one for now?

wagenet:
Turns out that the following will fail as of Ember 3.2:

```js
await settled();
assert.ok(isSettled());
```

wagenet:
See #458.

wagenet:
Feel free to close this if it seems superfluous after #458.

rwjblue:
Ya, #458 seems better for the underlying issue here. Closing...

PR #458, opened 2018-11-02
Adds a basic failing test for `settled`

wagenet:
As of Ember 3.2, `isSettled()` is no longer `true`, even immediately
after `await settled()`. This is because `waitUntil` uses an RSVP
Promise which resolves with a runloop. So when `waitUntil`'s promise
is about to resolve, things _are_ settled, but as soon as we actually
resolve, a new runloop is created which causes `isSettled()` to be
`false`.

### DIFF
--- a/addon-test-support/@ember/test-helpers/-utils.ts
+++ b/addon-test-support/@ember/test-helpers/-utils.ts
@@ -1,5 +1,74 @@
 /* globals Promise */
-import { Promise as RSVPPromise } from 'rsvp';
+
+import RSVP from 'rsvp';
+import hasEmberVersion from './has-ember-version';
+
+export class _Promise extends RSVP.Promise<void> {}
+
+const ORIGINAL_RSVP_ASYNC: Function = RSVP.configure('async');
+
+/*
+  Long ago in a galaxy far far away, Ember forced RSVP.Promise to "resolve" on the Ember.run loop.
+  At the time, this was meant to help ease pain with folks receiving the dreaded "auto-run" assertion
+  during their tests, and to help ensure that promise resolution was coelesced to avoid "thrashing"
+  of the DOM. Unfortunately, the result of this configuration is that code like the following behaves
+  differently if using native `Promise` vs `RSVP.Promise`:
+
+  ```js
+  console.log('first');
+  Ember.run(() => Promise.resolve().then(() => console.log('second')));
+  console.log('third');
+  ```
+
+  When `Promise` is the native promise that will log `'first', 'third', 'second'`, but when `Promise`
+  is an `RSVP.Promise` that will log `'first', 'second', 'third'`. The fact that `RSVP.Promise`s can
+  be **forced** to flush synchronously is very scary!
+
+  Now, lets talk about why we are configuring `RSVP`'s `async` below...
+
+  ---
+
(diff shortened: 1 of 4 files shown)


tests/integration/settled-test.js
```diff
@@ -144,6 +146,17 @@ module('settled real-world scenarios', function(hooks) {
     await teardownContext(this);
   });
 
+  test('basic behavior', async function(assert) {
+    await settled();
+
+    assert.ok(
+      isSettled(),
+      `out of balance xhr semaphores are cleaned up on teardown: ${JSON.stringify(
```
rwjblue: Can you tweak this message (it really isn't about semaphores here)...

scalvert:
:(

tests/integration/settled-test.js
wagenet: Oops! Bad copy paste :)

wagenet:
Hmm... I can't reproduce the error locally, and it seem unrelated to the change.

addon-test-support/@ember/test-helpers/-utils.ts
```diff
@@ -1,5 +1,43 @@
...
+
+      backburner.schedule('actions', () => {
+        adapter && adapter.asyncEnd();
+        callback(promise);
+      });
+    } else {
+      backburner.schedule('actions', () => callback(promise));
+    }
+  }
+});
```
Turbo87: Sorry, but... what does this do?

addon-test-support/@ember/test-helpers/-utils.ts
```diff
@@ -1,5 +1,43 @@
...
+    }
+  } else {
+    // this is the default behaviors from Ember:
+    //   https://github.com/emberjs/ember.js/blob/v3.4.6/packages/ember-testing/lib/ext/rsvp.js
+
+    // if schedule will cause autorun, we need to inform the test adapter
+    if (Ember.testing && !backburner.currentInstance) {
+      let adapter = Ember.Test.adapter as any;
+
+      adapter && adapter.asyncStart();
```
Turbo87: can we use a proper if condition here?

addon-test-support/@ember/test-helpers/-utils.ts
```diff
@@ -1,5 +1,43 @@
...
+      // use real native promise semantics whenever possible
+      Promise.resolve().then(() => callback(promise));
+    }  else {
+      // fallback to using RSVP's natural `asap` (**not** the fake
+      // one configured by Ember...)
+      RSVP.asap(callback, promise);
+    }
+  } else {
+    // this is the default behaviors from Ember:
+    //   https://github.com/emberjs/ember.js/blob/v3.4.6/packages/ember-testing/lib/ext/rsvp.js
```
Turbo87: does this mean we're copying stuff out of Ember? What if that code changes in the future and gets out of sync?

krisselden: @rwjblue can we not just copy the existing config async out instead of copying it?

rwjblue: Yep, great idea. Will update (along with some additional comments explaining the overall issue).

PR #463, opened 2018-11-07
Improve type declarations

Turbo87:
/cc @mike-north @chriskrycho @dfreeman @jamescdavis

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fire-event.ts
+++ b/addon-test-support/@ember/test-helpers/dom/fire-event.ts
@@ -17,6 +17,11 @@ const DEFAULT_EVENT_OPTIONS = { bubbles: true, cancelable: true };
 export const KEYBOARD_EVENT_TYPES = tuple('keydown', 'keypress', 'keyup');
 export type KeyboardEventType = typeof KEYBOARD_EVENT_TYPES[number];
 
+// eslint-disable-next-line require-jsdoc
+export function isKeyboardEventType(eventType: any): eventType is KeyboardEventType {
+  return KEYBOARD_EVENT_TYPES.indexOf(eventType) > -1;
+}
+
 const MOUSE_EVENT_TYPES = tuple(
   'click',
   'mousedown',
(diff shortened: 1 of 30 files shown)


addon-test-support/@ember/test-helpers/dom/fire-event.ts
```diff
@@ -11,8 +13,16 @@ const MOUSE_EVENT_CONSTRUCTOR = (() => {
   }
 })();
 const DEFAULT_EVENT_OPTIONS = { bubbles: true, cancelable: true };
-export const KEYBOARD_EVENT_TYPES = Object.freeze(['keydown', 'keypress', 'keyup']);
-const MOUSE_EVENT_TYPES = [
+
+export const KEYBOARD_EVENT_TYPES = tuple('keydown', 'keypress', 'keyup');
```
dfreeman: Totally possible I'm missing something, but what are you getting out of forcing these constants to be tuples rather than arrays?

dfreeman: Ahh, I see—forcing that to be a tuple also narrows the types of the elements as a side effect. What threw me was the fact that you don't ever actually care about the ordering of the elements, which is the normal scenario where you'd use a tuple.

The actual difference is probably negligible (except maybe unless someone else stumbles across this and is also confused), but you _can_ get the type system [to infer the narrowest possible element type without converting to a tuple](http://www.typescriptlang.org/play/#src=%2F%2F%20You%20can%20have%20arrays%20of%20literal%20types%20without%20forcing%20them%20to%20be%20tuples%3A%0Alet%20array%3A%20Array%3C'a'%20%7C%20'b'%20%7C%20'c'%3E%20%3D%20%5B'a'%2C%20'b'%2C%20'c'%5D%3B%0Atype%20Elements%20%3D%20typeof%20array%5Bnumber%5D%3B%0A%0A%0A%2F%2F%20And%20you%20can%20get%20a%20function%20application%20to%20infer%20that%0Afunction%20narrowArray%3CT%20extends%20string%20%2F*%20%7C%20number%20%7C%20...*%2F%3E(...items%3A%20T%5B%5D)%20%7B%0A%20%20%20%20return%20items%3B%0A%7D%0A%0Alet%20another%20%3D%20narrowArray('a'%2C%20'b'%2C%20'c')%3B%0Atype%20MoreElements%20%3D%20typeof%20another%5Bnumber%5D%3B).

dfreeman: Is there something reasonable to map _to_ though? In the DOM typings they're mapping the strings to the event interfaces themselves, but I don't see an equivalent "other half" here.

Turbo87:
as discussed with @rwjblue last week I'll merge this for now and we can keep iterating on the types afterwards

PR #465, opened 2018-11-07
Ensure typeIn has correct key option

mydea:
Currently, the `typeIn()` helper will only set the `charCode` option of the `KeyboardEvent`. This PR also adds the correct `key` option.


rwjblue:
@mydea - Have you had a chance to poke at my idea of reusing `triggerKeyEvent`s logic?

mydea:
I've pushed a commit making `typeIn` use `triggerKeyEvent`!

PR #481, opened 2018-11-21
Allow ember-cli-htmlbars-inline-precompile 2.x and 1.x

mydea:
See #468 

This PR allows ember-cli-htmlbars-inline-precompile 2.x as well as 1.x, to continue supporting Ember <= 2.12.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -28,6 +28,9 @@ module.exports = function() {
               'ember-fetch': null,
               'ember-data': '~3.1.0',
             },
+            dependencies: {
+              'ember-cli-htmlbars-inline-precompile': '^1.0.5',
+            },
           },
         },
         {
@@ -45,6 +48,9 @@ module.exports = function() {
               'ember-source': null,
               'ember-fetch': null,
             },
+            dependencies: {
+              'ember-cli-htmlbars-inline-precompile': '^1.0.5',
+            },
           },
         },
         {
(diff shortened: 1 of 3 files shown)


mydea:
Hmm, not entirely sure how to get the tests to pass - any pointers?

mydea:
Hmm, that fixes the 2.12 test, but not the ones before that.. not sure what the cause is there, any ideas?

config/ember-try.js
```diff
@@ -24,6 +24,7 @@ module.exports = function() {
           },
           npm: {
             devDependencies: {
+              'ember-cli-htmlbars-inline-precompile': '~1.0.5',
```
rwjblue: I think this needs to be moved to `dependencies` (not `devDependencies`), otherwise we would have `dependencies` of `2.0.0` and `devDependencies` of `1.0.5` and my guess is that `yarn` chooses `dependencies` to "win" over dev deps.

Also can you change these to `^1.0.5`

config/ember-try.js
```diff
@@ -68,6 +71,7 @@ module.exports = function() {
           name: 'ember-lts-2.12',
           npm: {
             devDependencies: {
+              'ember-cli-htmlbars-inline-precompile': '~1.0.5',
```
rwjblue: I don't think we need any modification to this one (since 2.12 already comes from npm and is supported by 2.0.0 of ember-cli-htmlbars-inline-precompile).

mydea:
Yeah, that did the tick - I added it to `dependencies` instead, now all tests pass!

rwjblue:
Thank you @mydea!

Issue #411, opened 2018-09-12
Not waiting for router transitions


simonihmig:
### Problem

`settled()` and all other async test helpers that use it do not wait for a model hook implemented using native async/await. 

### Test scenario:

In an application test there is an `await click()`, which triggers a route transition. The model hook is implemented using native async/await. The test continues to run beyond the await, even before the model hook has resolved (which means the new route hasn't rendered yet, though the test expects that)

### Solution

As suggested by @rwjblue on Discord: wait for active router transitions, as is the case in old `ember-testing`: https://github.com/emberjs/ember.js/blob/615daa2eeef971e578de75e6319baca3b0ec0508/packages/ember-testing/lib/helpers/wait.js#L45-L49

### Reproduction

https://github.com/simonihmig/async-model-reproduction

First test uses an explicit delay and passes. Second just `await`s `click()` and fails: https://github.com/simonihmig/async-model-reproduction/blob/master/tests/acceptance/foo-test.js

rwjblue:
Thank you for reporting! I can't believe I missed this in the initial development 😢 .

simonihmig:
@rwjblue getting totally confused now. I just pushed another commit to the repo, with a bar route that returns an RSVP promise. And even that is not working!? So seems not even related to native Promises/async!?? 

I mean this *has always been working*, I am getting totally lost now... 🤔

Anything that has changed, that would explain this?

dexturr:
I have run into this issue as well when trying to upgrade a project. 

I think this is an issue in ember-source as on my project if I only change ember-source from 3.1.1 -> 3.2.0 this issue appears with the same version of ember-test-helpers and every other dependency.

rwjblue:
Yep, this issue will be closed when the issue is fixed. 😸

rwjblue:
Fixed in https://github.com/emberjs/ember-test-helpers/pull/482, released in @ember/test-helpers@1.2.1.

lolmaus:
I'm on `@ember/test-helpers` 1.3.2 and transitions are still not being waited for.

E. g. I have this with [ember-router-helpers](https://github.com/rwjblue/ember-router-helpers/):

    <button {{action (transition-to "foo")}}>

When I do `await click('button')` in tests, it doesn't wait for the transition to finish.

I had to fork the `transition-to` helper and wrap it with a custom `@waitForPromise` decorator which registers a test waiter based on returned promise resolution.

[...]

PR #482, opened 2018-11-22
Wait for pending route transitions as part of settledness check

AlexTraher:
This aims to resolve an issue where `settled()` isn't waiting for router transitions. See here: [#411](https://github.com/emberjs/ember-test-helpers/issues/411).

This is a combined effort from myself and @dexturr.

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.ts
+++ b/addon-test-support/@ember/test-helpers/settled.ts
@@ -3,6 +3,8 @@ import { run } from '@ember/runloop';
 import jQuery from 'jquery';
 
 import Ember from 'ember';
+
+import { getContext } from './setup-context'
 import { nextTick } from './-utils';
 import waitUntil from './wait-until';
 
@@ -39,6 +41,23 @@ function pendingRequests() {
   let internalRequestsPending = _internalPendingRequests();
 
   return localRequestsPending + internalRequestsPending;
+}
+
+/**
+  @private
+  @returns {boolean} if there are any pending router transitions
+*/
+function hasPendingTransitions() {
+  const context = getContext();
+  const owner = context && context.owner;
+
+  if (!owner) {
+    return false;
+  }
+
+  const router = owner.lookup('router:main');
+  const isLoading = router._routerMicrolib && !!router._routerMicrolib.activeTransition;
+  return isLoading;
 }
 
 /**
@@ -147,6 +166,7 @@ export interface SettledState {
   hasPendingTimers: boolean;
   hasPendingWaiters: boolean;
   hasPendingRequests: boolean;
+  hasPendingTransitions: boolean;
   pendingRequestCount: number;
 }
 
@@ -177,6 +197,7 @@ export function getSettledState(): SettledState {
     hasRunLoop: Boolean((run as any).currentRunLoop),
     hasPendingWaiters: checkWaiters(),
     hasPendingRequests: pendingRequestCount > 0,
+    hasPendingTransitions: hasPendingTransitions(),
     pendingRequestCount,
   };
 }
@@ -192,9 +213,9 @@ export function getSettledState(): SettledState {
   @returns {boolean} `true` if settled, `false` otherwise
 */
 export function isSettled(): boolean {
-  let { hasPendingTimers, hasRunLoop, hasPendingRequests, hasPendingWaiters } = getSettledState();
+  let { hasPendingTimers, hasRunLoop, hasPendingRequests, hasPendingWaiters, hasPendingTransitions } = getSettledState();
 
-  if (hasPendingTimers || hasRunLoop || hasPendingRequests || hasPendingWaiters) {
+  if (hasPendingTimers || hasRunLoop || hasPendingRequests || hasPendingWaiters || hasPendingTransitions) {
     return false;
   }
 
--- a/addon-test-support/@ember/test-helpers/settled.ts
+++ b/addon-test-support/@ember/test-helpers/settled.ts
@@ -167,6 +167,9 @@ export interface SettledState {
   - `hasPendingRequests` - Checks if there are pending AJAX requests (based on
     `ajaxSend` / `ajaxComplete` events triggered by `jQuery.ajax`). If there
     are pending requests, this will be `true`, otherwise `false`.
+  - `hasPendingTransitions` - Checks if there are pending route transitions. If the
+    router has not been instantiated / setup for the test yet this will return `null`,
+    if there are pending transitions, this will be `true`, otherwise `false`.
   - `pendingRequestCount` - The count of pending AJAX requests.
 
   @public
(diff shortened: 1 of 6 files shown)


rwjblue:
Oh, no! I had started to work on fixing this bug over the holiday weekend (US Thanksgiving). You can see my WIP over in https://github.com/emberjs/ember-test-helpers/compare/master...rwjblue:settled-router-transitions if you are curious.

Shall we compare notes?

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -41,6 +43,23 @@ function pendingRequests() {
   return localRequestsPending + internalRequestsPending;
 }
 
+/**
+  @private
+  @returns {boolean} if there are any pending router transitions
+*/
+function hasPendingTransitions() {
+  const context = getContext();
```
rwjblue: We are generally in the "liberally-let" camp (phrasing [from this blog post](https://madhatted.com/2016/1/25/let-it-be)). This should be `let`...

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -41,6 +43,23 @@ function pendingRequests() {
...
+  const owner = context && context.owner;
+
+  if (!owner) {
+    return false;
+  }
+
+  const router = owner.lookup('router:main');
+  const isLoading = router._routerMicrolib && !!router._routerMicrolib.activeTransition;
```
rwjblue: In order to support older versions we will need to also check `router.router`.

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -41,6 +43,23 @@ function pendingRequests() {
...
+*/
+function hasPendingTransitions() {
+  const context = getContext();
+  const owner = context && context.owner;
+
+  if (!owner) {
+    return false;
+  }
+
+  const router = owner.lookup('router:main');
```
rwjblue: I would like to avoid doing this if we don't need it. Specifically, when there hasn't been a `visit` call, we don't need to check for transitions...

tests/unit/settled-test.js
```diff
@@ -144,6 +145,32 @@ module('settled', function(hooks) {
...
       assert.strictEqual(isSettled(), true, 'post cond');
     });
+
+    test('when router is transitioning', async function(assert) {
+      assert.expect(3);
+
+      setContext(this);
+      assert.strictEqual(isSettled(), true);
+
+      this.owner = {
```
rwjblue: I'd rather not use mocking style unit tests like this. In past versions of ember-test-helpers these style tests led to brittleness across Ember versions (and false confidence about what works in reality).

tests/unit/settled-test.js
```diff
@@ -281,6 +316,38 @@ module('settled', function(hooks) {
       assert.strictEqual(isSettled(), true, 'post cond');
     });
 
+    test('when router is transitioning', function(assert) {
+      assert.expect(3);
+      assert.strictEqual(isSettled(), true);
+
+      setContext(this);
+      this.owner = {
```
rwjblue: Same as above here. I'd really like tests that reflect a users reality (instead of mocking).

tests/unit/setup-application-context-test.js
```diff
@@ -29,8 +33,21 @@ module('setupApplicationContext', function(hooks) {
   if (!hasEmberVersion(2, 4)) {
     return;
   }
-
+  let isolatedApp;
   hooks.beforeEach(async function() {
+    const AppConfig = assign({ autoboot: false }, config.APP);
+    // .extend() because initializers are stored in the constructor, and we
```
rwjblue: This seems somewhat unrelated to the rest of the changes, which failures caused you to need to change this?

rwjblue:
OK, cool, I'll get some of my branches changes landed into this PR in the next hour or so...

tests/unit/setup-application-context-test.js
dexturr: @AlexTraher and I found the `setup-application-context-test.js` tests to fail intermittently before introducing this.

After some digging we found it was order dependent, for instance when we ran these two tests:
>setupApplicationContext: can perform a basic template rendering
>DOM Helper: click > non-focusable element types: clicking a div via element with context set

If the "DOM Helper" test was run first then we found that the "setupApplicationContext" test failed, however if you reran the tests then the "setupApplicationContext" would go first and both tests would pass.

We found commenting out [this line](https://github.com/emberjs/ember-test-helpers/pull/482/files/a7db2583040cb566d0601bb95450a7d25f39ceac#diff-2556457e0cd69598faa308feee8b8ffaR51) in the `hasPendingTransitions` method. Stopped the error, but is needed for the logic to work.

We attempted to find the bug, thought we might have some state that was not being cleaned up correctly, but I am sad to say we were unsuccessful.

So the next best idea we had was to construct an isolated `Ember.App()` to run the failing tests against [as is done here](https://github.com/emberjs/ember-test-helpers/blob/50c24e181f03b511cb9cb94b357251160fc1496e/tests/unit/setup-context-test.js#L294) and thats how we came to this solution. Neither of us were particularly happy with this and we hope there is a better one if you have any ideas?

tests/unit/setup-application-context-test.js
```diff
@@ -51,10 +55,20 @@ module('setupApplicationContext', function(hooks) {
...
+              resolve();
+            }, 50);
+          });
+        },
+      }),
     });
 
-    setApplication(application);
-
+    setContext(this);
```
AlexTraher: @rwjblue This didn't seem to fix the test issues @dexturr and I were seeing (when we added the shared state workaround discussed [here](https://github.com/emberjs/ember-test-helpers/pull/482#discussion_r236343138)). We weren't quite sure why it didn't resolve the issue?

dexturr: @rwjblue ping

We were wondering if order dependent tests were failing [due to sharing an Ember app across all tests](https://github.com/emberjs/ember-test-helpers/blob/master/tests/helpers/resolver.js#L11)? But were unsure of how to proceed to discover to fix this bug.

Would love to get in there and get it sorted! Any ideas?

Cheers.

rwjblue:
Rebased with master to fixup conflicts.

tests/unit/setup-application-context-test.js
rwjblue: Ya, you were totally correct. We had some test leakage issues, but not in the most obvious way 😩.

I documented my findings in d115c82bce552de9d2647fab6125efa786e85e0d along with  some better/alternative fixes, though I think the fix in that commit is fine for now.

rwjblue:
Thank you again @dexturr and @AlexTraher! Your work (and debugging) was very very helpful in getting this over the line...

PR #497, opened 2018-12-07
Only customize RSVP's async for Ember < 1.7.

rwjblue:
As of Ember 1.7 and higher, this customization is done automatically.  Originally we intended to just leave this alone, and not update (since these are already legacy testing systems) but apparently this is actively breaking other fixes in the newer system if *any* older style tests still exist in the test suite.

The main issue with this code on newer Ember versions is that it does not properly call `adapter.asyncStart()` and `adapter.asyncEnd()` which means that we no longer wait for the promise resolution in the test harness itself. :sob:

### DIFF
--- a/addon-test-support/ember-test-helpers/legacy-0-6-x/ext/rsvp.js
+++ b/addon-test-support/ember-test-helpers/legacy-0-6-x/ext/rsvp.js
@@ -1,5 +1,6 @@
 import RSVP from 'rsvp';
 import { run } from '@ember/runloop';
+import hasEmberVersion from '../../has-ember-version';
 
 let originalAsync;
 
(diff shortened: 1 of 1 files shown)


addon-test-support/ember-test-helpers/legacy-0-6-x/ext/rsvp.js
```diff
@@ -11,13 +12,15 @@ let originalAsync;
   @private
 */
 export function _setupPromiseListeners() {
```
scalvert: Do we actually need to still call these as a noop (effectively) if you're < 1.7? Can't we just simply not call these at all?

rwjblue: Ya, we could, but this seemed to localize the logic to the specific area that needed to care about the versioning.

PR #498, opened 2018-12-07
Add ability to opt-out of automatic settledness waiting in teardown.

rwjblue:
In order to allow things like `ember-qunit`'s upcoming async leak detection, we need to avoid waiting for `settled()`. By default we should continue to effectively `await settled()` but when ember-qunit (and soon ember-mocha) is configured to use async leak detection they can appropriately pass `waitForSettled` when they call `teardownContext` / `teardownRenderingContext` / `teardownApplicationContext`.

### DIFF
--- a/addon-test-support/@ember/test-helpers/teardown-context.ts
+++ b/addon-test-support/@ember/test-helpers/teardown-context.ts
@@ -17,9 +17,18 @@ import Ember from 'ember';
 
   @public
   @param {Object} context the context to setup
+  @param {Object} [options] options used to override defaults
+  @param {boolean} [options.waitForSettled=true] should the teardown wait for `settled()`ness
   @returns {Promise<void>} resolves when settled
 */
-export default function teardownContext(context: TestContext): Promise<void> {
+export default function teardownContext(
+  context: TestContext,
+  options?: { waitForSettled?: boolean }
+): Promise<void> {
+  let waitForSettled = true;
+  if (options !== undefined && 'waitForSettled' in options) {
+    waitForSettled = options.waitForSettled!;
+  }
   return nextTickPromise()
     .then(() => {
       let { owner } = context;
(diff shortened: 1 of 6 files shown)


addon-test-support/@ember/test-helpers/teardown-context.ts
```diff
@@ -31,13 +40,21 @@ export default function teardownContext(context: TestContext): Promise<void> {
...
+      if (waitForSettled) {
+        return settled();
+      }
+
+      return nextTickPromise();
     })
     .finally(() => {
       let contextGuid = guidFor(context);
 
       runDestroyablesFor(CLEANUP, contextGuid);
```
Turbo87: can we run this in a potentially unsettled state?

tests/unit/teardown-context-test.js
```diff
@@ -69,4 +76,16 @@ module('teardownContext', function(hooks) {
       assert.equal(state.pendingRequestCount, 0, 'pendingRequestCount is 0');
     });
   }
+
+  test('can opt out of waiting for settledness', async function(assert) {
+    later(() => assert.step('later'), 200);
```
Turbo87: looks like a potentially flaky test to me due to the fixed timeout. Maybe RSVP.defer() + custom waiter would be better?

tests/unit/teardown-rendering-context-test.js
```diff
@@ -54,4 +56,17 @@ module('teardownRenderingContext', function(hooks) {
       'previous ember-testing element is no longer in DOM'
     );
   });
+
+  test('can opt out of waiting for settledness', async function(assert) {
+    later(() => assert.step('later'), 200);
```
Turbo87: see above

tests/unit/teardown-context-test.js
rwjblue: Ya, I'll swap to a waiter (RSVP.defer also has issues).

addon-test-support/@ember/test-helpers/teardown-context.ts
rwjblue: Yes, the cleanup itself is unrelated to settledness (we call `owner.destroy()` synchronously just above).

PR #529, opened 2019-01-18
Fix lying doc block on findAll test helper

yandavid:
The `findAll` helper states that it is equivalent to running `querySelectorAll` but that is a lie. Tests were failing when I was doing `querySelectorAll` to `findAll` conversions as I was relying on the `item` method on `NodeList` instances.


rwjblue:
Thank you!

PR #540, opened 2019-01-24
Feature - adds setupOnerror function to allow patching Ember.onerror in tests

scalvert:
It's fairly common to want to temporarily set an `Ember.onerror` handler to test specific aspects of error handling within your application. Newer versions of sinon do not make this easy, as the `Ember.onerror` object is a defined property, and sinon has a codepath that excludes stubbing this particular method (it's a bit weird to _stub_ it anyway, isn't it?). 

This PR introduces a utility to setup `Ember.onerror`, and ensures that it's reset to its original state when the test context is torn down.

TODO:

- [x] Write tests

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-context.ts
+++ b/addon-test-support/@ember/test-helpers/setup-context.ts
@@ -132,6 +132,29 @@ export function resumeTest(): void {
   context.resumeTest();
 }
 
+const ORIGINAL_EMBER_ONERROR: (error: Error) => void | undefined = Ember.onerror;
+
+/**
+  Sets the Ember.onerror function for the duration of a single test. This
+  value is reset after each test to ensure correct test isolation.
+
+  @public
+  @param {Function} onError the onError function to be set on Ember.onerror
+*/
+export function setupOnerror(onError: (error: Error) => void): void {
+  let context = getContext();
+  let contextGuid = guidFor(context);
+  let contextCleanup = CLEANUP[contextGuid];
+
+  if (!Array.isArray(contextCleanup)) {
+    throw new Error(
+      'You must use `setupContext` / `setupRenderingContext` / `setupApplicationContext` in order to use `setupOnerror`'
+    );
+  }
+
+  Ember.onerror = onError;
+}
+
 export const CLEANUP = Object.create(null);
 
 /**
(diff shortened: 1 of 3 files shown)


addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -132,6 +132,19 @@ export function resumeTest(): void {
...
+const ORIGINAL_EMBER_ONERROR: (error: Error) => void | undefined = Ember.onerror;
+
+/**
+  Sets the Ember.onerror function for the duration of a single test. This
+  value is reset after each test to ensure correct test isolation.
+
+  @public
+  @param {Function} onError the onError function to be set on Ember.onerror
+*/
+export function setupEmberOnerror(onError: (error: Error) => void): void {
```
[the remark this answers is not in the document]
scalvert: Yep.

addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -132,6 +133,25 @@ export function resumeTest(): void {
...
+
+/**
+  Sets the Ember.onerror function for the duration of a single test. This
+  value is reset after each test to ensure correct test isolation.
+
+  @public
+  @param {Function} onError the onError function to be set on Ember.onerror
+*/
+export function setupOnerror(onError: (error: Error) => void): void {
+  if (!hasEmberVersion(2, 4)) {
```
rwjblue: I think what we want is:

```js
let context = getContext();
let contextCleanup = CLEANUP[contextGuid];

if (!Array.isArray(contextCleanup)) {
  throw new Error(
    'You must use `setupContext` / `setupRenderingContext` / `setupApplicationContext` in order to use `setupOnerror`'
  );
}
```

What do you think?

scalvert: You tell me. Is it less important the Ember version vs. the fact that this GUID is present in the CLEANUP?

rwjblue: For this check, the Ember version is unrelated (Ember.onerror = function() {} has worked since Ember 1.0). The thing we care about is that you will get the ability to have it cleaned up for you...

scalvert: Yep. I understand. That seems fine. I'll change it.

scalvert: What should the corresponding conditional be around the test for throws? https://github.com/emberjs/ember-test-helpers/pull/540/files#diff-51eaa617037ec59cacc57f7ccd70ca8aR7

scalvert: I just left the test as is for now.

PR #543, opened 2019-01-27
Add deprecation for this.$ in rendering tests

simonihmig:
As per RFC386

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
@@ -9,6 +9,7 @@ import settled from './settled';
 import hbs, { TemplateFactory } from 'htmlbars-inline-precompile';
 import getRootElement from './dom/get-root-element';
 import { Owner } from './build-owner';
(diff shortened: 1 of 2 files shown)


addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -53,6 +54,17 @@ function lookupOutletTemplate(owner: Owner): any {
   @returns {jQuery} a jQuery object representing the selector (or element itself if no selector)
 */
 function jQuerySelector(selector: string): any {
+  deprecate(
+    'Using this.$() in a rendering test has been deprecated, consider using this.element instead.',
+    false,
+    {
+      id: 'ember-test-helpers.rendering-context.jquery-element',
+      until: '2.0.0',
+      // @ts-ignore
```
simonihmig: `@types/ember` does not have `url` in its options interface! 🤔Seems like a bug?

simonihmig: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32538

PR #548, opened 2019-01-29
Refactors setupOnerror to be a pure setup/teardown approach

scalvert:
The first version of `setupOnerror` was wired up to the new hooks API `setupContext`/`teardownContext` functions, ensuring that the correct teardown occurred for each test. 

The old API, however, could leak the patched `Ember.onerror` as the teardown would never be called.

This PR makes the utilities simple, in that it simply provides the patching capabilities. It's expected that `ember-qunit` will wire up the teardown in `QUnit.testDone`, or some equivalent. This will also be the path forward for `ember-mocha`, using a global `afterEach`.


tests/unit/setup-ember-onerror-test.js
```diff
@@ -1,58 +1,28 @@
...
       assert.equal(Ember.onerror, undefined);
 
-      await setupContext(context);
-
       setupOnerror(onerror);
 
       assert.equal(Ember.onerror, onerror);
 
-      await teardownContext(context);
+      setupOnerror();
```
rwjblue: Can you move this to `afterEach` (to ensure we don't leak)?

scalvert: Surely

PR #554, opened 2019-02-02
Moving test isolation validation primatives into @ember/test-helpers

scalvert:
Moving test isolation validation primatives into @ember/test-helpers to enable similar logic to be shared by ember-qunit and ember-mocha.

Currently, the features exists in its entirety in [ember-qunit](https://github.com/emberjs/ember-qunit/blob/master/addon-test-support/ember-qunit/test-isolation-validation.js). In order to share this functionality we need to merge the primitives there into this library.

### DIFF
--- a/addon-test-support/@ember/test-helpers/-internal/debug-info.ts
+++ b/addon-test-support/@ember/test-helpers/-internal/debug-info.ts
@@ -36,6 +36,12 @@ export default interface DebugInfo {
   toConsole: () => void;
 }
 
+/**
+ * Determins if the `getDebugInfo` method is available in the
+ * running verison of backburner.
+ *
+ * @returns {boolean} True if `getDebugInfo` is present in backburner, otherwise false.
+ */
 export function backburnerDebugInfoAvailable() {
   return typeof run.backburner.getDebugInfo === 'function';
 }
--- a/addon-test-support/@ember/test-helpers/-internal/debug-info.ts
+++ b/addon-test-support/@ember/test-helpers/-internal/debug-info.ts
@@ -13,14 +13,14 @@ const SCHEDULED_AUTORUN = 'Scheduled autorun';
 
 type MaybeDebugInfo = BackburnerDebugInfo | null;
 
-interface ISettledState {
+interface SettledState {
   hasPendingTimers: boolean;
   hasRunLoop: boolean;
   hasPendingWaiters: boolean;
   hasPendingRequests: boolean;
 }
 
-interface ISummaryInfo {
+interface SummaryInfo {
   hasPendingRequests: boolean;
   hasPendingWaiters: boolean;
   autorunStackTrace: string | undefined | null;
@@ -32,6 +32,14 @@ interface ISummaryInfo {
   hasRunLoop: boolean;
 }
 
+export default interface DebugInfo {
+  toConsole: () => void;
+}
+
+export function backburnerDebugInfoAvailable() {
+  return typeof run.backburner.getDebugInfo === 'function';
+}
+
 /**
  * Retrieves debug information from backburner's current deferred actions queue (runloop instance).
  * If the `getDebugInfo` method isn't available, it returns `null`.
@@ -40,10 +48,7 @@ interface ISummaryInfo {
  * @returns {MaybeDebugInfo | null} Backburner debugInfo or, if the getDebugInfo method is not present, null
  */
 export function getDebugInfo(): MaybeDebugInfo {
-  let debugEnabled = run.backburner.DEBUG === true;
-  let getDebugInfoAvailable = typeof run.backburner.getDebugInfo === 'function';
-
-  return debugEnabled && getDebugInfoAvailable
+  return run.backburner.DEBUG === true && backburnerDebugInfoAvailable()
     ? <BackburnerDebugInfo>run.backburner.getDebugInfo()
     : null;
 }
@@ -59,10 +64,10 @@ export function getDebugInfo(): MaybeDebugInfo {
  * - info provided by backburner's getDebugInfo method (timers, schedules, and stack trace info)
  *
  */
-export default class DebugInfo {
-  private _settledState: ISettledState;
+export class TestDebugInfo implements DebugInfo {
+  private _settledState: SettledState;
   private _debugInfo: MaybeDebugInfo;
-  private _summaryInfo: ISummaryInfo | undefined = undefined;
+  private _summaryInfo: SummaryInfo | undefined = undefined;
 
   constructor(
     hasPendingTimers: boolean,
@@ -81,9 +86,9 @@ export default class DebugInfo {
     this._debugInfo = debugInfo;
   }
 
-  get summary(): ISummaryInfo {
+  get summary(): SummaryInfo {
     if (!this._summaryInfo) {
-      this._summaryInfo = assign(<ISummaryInfo>{}, this._settledState);
+      this._summaryInfo = assign(<SummaryInfo>{}, this._settledState);
 
       if (this._debugInfo) {
         this._summaryInfo.autorunStackTrace =
@@ -118,11 +123,6 @@ export default class DebugInfo {
     }
 
     return this._summaryInfo;
-  }
-
-  get message(): string {
-    return `Test is not isolated (async execution is extending beyond the duration of the test).\n
-    More information has been printed to the console. Please use that information to help in debugging.\n\n`;
   }
 
   toConsole(_console = console): void {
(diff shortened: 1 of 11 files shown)


scalvert:
The current failures are related to types not being correctly merged together in the `@ember/runloop` module. To address this, I've opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32970, which will hopefully allow us to use `run`'s `backburner` property without having to locally define it.

scalvert:
Waiting until upstream, unrelated TS issues are fixed in typescript@next. Once those are done, I'll rework the types in this PR to use `@types/ember__runloop` types.

rwjblue:
@scalvert and I chatted a bit about this yesterday. For now he is going to just inline the types we need directly (with a comment) from @types/ember to unblock general progress forward here. Once the upstream TS issue (and types/ember PR) is fixed, we can just remove those extraneous types.

scalvert:
I've inlined the types as discussed. Waiting on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32970.

addon-test-support/@ember/test-helpers/index.ts
```diff
@@ -22,6 +22,7 @@ export { default as settled, isSettled, getSettledState } from './settled';
 export { default as waitUntil } from './wait-until';
 export { default as validateErrorHandler } from './validate-error-handler';
 export { default as setupOnerror, resetOnerror } from './setup-onerror';
+export { default as DebugInfo, getDebugInfo } from './-internal/debug-info';
```
rwjblue: Do we need to expose `DebugInfo` here? I would think that constructing the class directly is “private”.

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -174,15 +176,20 @@ export interface SettledState {
   @returns {Object} object with properties for each of the metrics used to determine settledness
 */
 export function getSettledState(): SettledState {
```
rwjblue: The documentation for this method needs to be updated to include the new property, and probably cross link to the DebugInfo docs.

addon-test-support/@ember/test-helpers/-internal/debug-info.ts
```diff
@@ -0,0 +1,169 @@
...
+ * from:
+ * - info provided by getSettledState
+ *    - hasPendingTimers
+ *    - hasRunLoop
+ *    - hasPendingWaiters
+ *    - hasPendingRequests
+ * - info provided by backburner's getDebugInfo method (timers, schedules, and stack trace info)
+ *
+ */
+export default class DebugInfo {
```
rwjblue: Let’s make an IDebugInfo interface, which is public, then make this class implement the interface. Specifically, I want to make it clear that _using_ the public members of IDebugInfo is good, but relying on internal class implementation details is not. Also, it makes it clearer that the user can’t instantiate per-se (that you’d get one by way of getSettledState).

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -149,6 +150,7 @@ export interface SettledState {
   hasPendingRequests: boolean;
   hasPendingTransitions: boolean | null;
   pendingRequestCount: number;
+  debugInfo?: DebugInfo;
```
rwjblue: I’d like this to reference `IDebugInfo` instead of the class itself...

addon-test-support/@ember/test-helpers/-internal/debug-info.ts
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * Retrieves debug information from backburner's current deferred actions queue (runloop instance).
+ * If the `getDebugInfo` method isn't available, it returns `null`.
+ *
+ * @public
+ * @returns {MaybeDebugInfo | null} Backburner debugInfo or, if the getDebugInfo method is not present, null
+ */
+export function getDebugInfo(): MaybeDebugInfo {
+  let debugEnabled = run.backburner.DEBUG === true;
+  let getDebugInfoAvailable = typeof run.backburner.getDebugInfo === 'function';
```
rwjblue: Can we extract this into a separate helper function? Specifically in the tests we need to know if the runtimes backburner is capable of providing getDebugInfo independent of if the DEBUG flag is set. 

Maybe `backburnerSupportsDebugInfo()`?

tests/unit/test-debug-info-test.js
```diff
@@ -0,0 +1,208 @@
+import { module, test } from 'qunit';
+import { run } from '@ember/runloop';
+import DebugInfo, { getDebugInfo } from '@ember/test-helpers/-internal/debug-info';
+import MockStableError, { overrideError, resetError } from './utils/mock-stable-error';
+import { MockConsole, getRandomBoolean, getMockDebugInfo } from './utils/test-isolation-helpers';
+
+module('TestDebugInfo', function() {
+  test('summary returns minimal information when debugInfo is not present', function(assert) {
```
rwjblue: This module should have a beforeEach that sets run.backburner.DEBUG = false (since the tests rely on it being off).

tests/unit/test-debug-info-test.js
```diff
@@ -0,0 +1,208 @@
...
+      hasPendingWaiters: false,
+      hasRunLoop: false,
+      pendingScheduledQueueItemCount: 2,
+      pendingScheduledQueueItemStackTraces: ['STACK', 'STACK'],
+      pendingTimersCount: 2,
+      pendingTimersStackTraces: ['STACK', 'STACK'],
+    });
+  });
+
+  if (getDebugInfo()) {
```
rwjblue: This should use that extracted helper that I mentioned backburnerSupportsDebugInfo() instead, then in its beforeEach set the DEBUG flag. Otherwise this test may or may not be actually ran depending on the ambient state of the debug macro.

types/@ember/runloop.d.ts
```diff
@@ -0,0 +1,339 @@
+/**
+ * Temporarily copying these types into @ember/test-helpers until
```
rwjblue: Would you mind adding something like BEGIN / END comments so we can tell _which_ parts can be deleted? Or is it that we delete everything?

addon-test-support/@ember/test-helpers/index.ts
scalvert: Oh you're right. We don't need to instantiate it in `ember-qunit` anymore. I'll remove this.

addon-test-support/@ember/test-helpers/settled.ts
scalvert: Roger.

addon-test-support/@ember/test-helpers/-internal/debug-info.ts
scalvert: Agree. Doing it.

scalvert: I'm doing some renaming here for consistency and clarity.

addon-test-support/@ember/test-helpers/settled.ts
scalvert: Done.

addon-test-support/@ember/test-helpers/-internal/debug-info.ts
scalvert: I can, sure.

tests/unit/test-debug-info-test.js
scalvert: Done. Also added `run.backburner.DEBUG = true` for the nested module's `beforeEach`, since it relies on it.

tests/unit/test-debug-info-test.js
scalvert: Yerp.

types/@ember/runloop.d.ts
scalvert: Ya, we delete the whole shebang. Specifically, it will be replaced with an extension of the `RunNamespace` from `@types/ember__runloop` (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32970/files#diff-b87fda3978af44033e6da75ab82d1bc9R9).

rwjblue: 👏

scalvert:
Updated. Should be good now I think.

rwjblue:
Changes look good, I think we still need to do the two bullet points in https://github.com/emberjs/ember-test-helpers/pull/554#pullrequestreview-204530948 though.

scalvert:
I've done #1, but I'm not sure what you mean by #2.

rwjblue:
I just pushed a commit addressing both, should be good to go once CI is finished.

PR #565, opened 2019-02-14
Add pauseTestFor method.

rwjblue:
Ensures that tests wait for the provided promise before being considered "settled".

This function is expected to be used in application/addon code (aka non-test code).  When used during a test run, it ensures that the next `await settled()` check does not result until after the promise has settled, when used outside of a test it is a simple "pass through" function.

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.ts
+++ b/addon-test-support/@ember/test-helpers/settled.ts
@@ -1,4 +1,5 @@
 /* globals jQuery */
+
 import { run } from '@ember/runloop';
 import Ember from 'ember';
 
@@ -142,12 +143,52 @@ function checkWaiters() {
   return false;
 }
 
+interface IPauseTestPromiseDetails {
+  label: undefined | string;
+  error: Error;
+}
+
+let currentPauseTestPromises = new Map<Promise<unknown>, IPauseTestPromiseDetails>();
+
+/**
+  Ensures that tests wait for the provided promise before being considered "settled".
+
+  This function is expected to be used in application/addon code (aka non-test code).
+  When used during a test run, it ensures that the next `await settled()` check does
+  not result until after the promise has settled, when used outside of a test it is a
+  simple "pass through" function.
+
+  @public
+  @param {Promise<T>} promise the promise to wait for
+  @param {string} [label] an optional label for the promise
+  @returns {Promise<T>} the passed in promise
+*/
+export function pauseTestFor<T>(promise: Promise<T>, label?: string): Promise<T> {
+  let error = new Error();
+
+  currentPauseTestPromises.set(promise, { label, error });
+
+  return promise.then(
+    result => {
+      currentPauseTestPromises.delete(promise);
+
+      return result;
+    },
+    error => {
+      currentPauseTestPromises.delete(promise);
+
+      throw error;
+    }
+  );
+}
+
 export interface SettledState {
   hasRunLoop: boolean;
   hasPendingTimers: boolean;
   hasPendingWaiters: boolean;
   hasPendingRequests: boolean;
   hasPendingTransitions: boolean | null;
+  hasPendingPauseTestForPromises: boolean;
   pendingRequestCount: number;
 }
 
@@ -182,6 +223,7 @@ export function getSettledState(): SettledState {
     hasPendingWaiters: checkWaiters(),
     hasPendingRequests: pendingRequestCount > 0,
     hasPendingTransitions: hasPendingTransitions(),
+    hasPendingPauseTestForPromises: currentPauseTestPromises.size !== 0,
     pendingRequestCount,
   };
 }
@@ -203,6 +245,7 @@ export function isSettled(): boolean {
     hasPendingRequests,
     hasPendingWaiters,
     hasPendingTransitions,
+    hasPendingPauseTestForPromises,
   } = getSettledState();
 
   if (
@@ -210,7 +253,8 @@ export function isSettled(): boolean {
     hasRunLoop ||
     hasPendingRequests ||
     hasPendingWaiters ||
-    hasPendingTransitions
+    hasPendingTransitions ||
+    hasPendingPauseTestForPromises
   ) {
     return false;
   }
--- a/addon-test-support/@ember/test-helpers/settled.ts
+++ b/addon-test-support/@ember/test-helpers/settled.ts
@@ -155,7 +155,7 @@ let currentPauseTestPromises = new Map<Promise<unknown>, IPauseTestPromiseDetail
 
   This function is expected to be used in application/addon code (aka non-test code).
   When used during a test run, it ensures that the next `await settled()` check does
-  not result until after the promise has settled, when used outside of a test it is a
+  not resolve until after the promise has settled, when used outside of a test it is a
   simple "pass through" function.
 
   @public
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,7 +16,7 @@ module.exports = {
   overrides: [
     {
       files: ['.eslintrc.js', '.prettierrc.js', 'index.js', 'config/ember-try.js', 'scripts/**'],
-      excludedFiles: ['addon-test-support/**', 'tests/**'],
+      excludedFiles: ['addon-test-support/**', 'addon/**', 'tests/**'],
       parserOptions: {
         ecmaVersion: 2015,
         sourceType: 'script',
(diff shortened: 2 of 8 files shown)


scalvert:
To be clear, this is a more async-aware, lightweight test waiter, right? It basically provides you a quick and easy way to setup pauses until promise resolution.

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -142,12 +143,43 @@ function checkWaiters() {
...
+*/
+export function pauseTestFor<T>(promise: Promise<T>): Promise<T> {
+  currentPauseTestPromiseSemaphore++;
+  return promise.then(
+    result => {
+      currentPauseTestPromiseSemaphore--;
+
+      return result;
+    },
+    error => {
```
scalvert: Why this vs. `catch`?

scalvert:
In fact, looking at the implementation, this is exactly what we did in [ember-app-scheduler](https://github.com/ember-app-scheduler/ember-app-scheduler/blob/master/addon/scheduler.ts#L138), albeit a genericized version.

Would we recommend, in docs, for people to surround with the `DEBUG` macro?

👍

rwjblue:
> The one question I would ask is if we see a fairly clear path from here to something else we've discussed in the past, which is the ability to provide some more detail so users can better figure out _why_ things aren't settled
>
> For example, a simple (but non-trivial overhead) solution would be to replace the semaphor with an array that stores the tags of un-resolved promises.

Yes, this was exactly my plan. I was going to store the stack trace which called `pauseTestFor` along with the promise so that you can directly see the _source_ (hopefully this points at app code at that point) causing the settledness to be paused. 

I wanted to wait to incorporate that info until @scalvert's https://github.com/emberjs/ember-test-helpers/pull/554 lands which brings _tons_ of information (by way of stack traces) to `getSettledState().debugInfo` and a really convenient `getSettledState().debugInfo.toConsole()` method to print things out nicely.

rwjblue:
@scalvert 

> Would we recommend, in docs, for people to surround with the `DEBUG` macro?

No, the idea is that you would just leave it in. The "prod" version of `pauseTestFor` is _literally_ `promise => promise` (pass through) and **basically** zero cost.

addon-test-support/@ember/test-helpers/settled.ts
rwjblue:
```js
promise
  .then(...)
  .catch(...)
```

Creates 2 chained promises, whereas `promise.then(good,bad)` makes one.

rwjblue:
@scalvert 

> To be clear, this is a more async-aware, lightweight test waiter, right? It basically provides you a quick and easy way to setup pauses until promise resolution.

Yes, but I also envision plugging it into the work you are doing in #554 (in the `getSettledState().debugInfo` object), so at that point it will be _much_ more that just a simple test waiter (because its much more debuggable).

bendemboski:
> Yes, this was exactly my plan. I was going to store the stack trace which called `pauseTestFor` along with the promise so that you can directly see the _source_ (hopefully this points at app code at that point) causing the settledness to be paused.

I'm not totally clear -- are you thinking for storing just this, or this plus a string/tag? I don't think the stack trace captures everything because `pauseTestFor` could easily be used in some async helper function that truncates the (async) stack trace somewhere un-helpful. I can't recall -- is there an easy way to "name" promises or attach metadata to them? If so, that could be enough, but if not, then I think we'd still want a string/tag.

Anyway, if the plan is "store an array of info about each pending promise" then whether or not we end up adding a string/tag argument, I think this PR is sufficiently "future-compatible".

rwjblue:
> I can't recall -- is there an easy way to "name" promises or attach metadata to them?

Yep, RSVP promises have a label property (second argument to constructor, third argument to `.then`).

I'm happy to tweak this API to add a userland label too (e.g. `pauseTestFor(promise, "some optional label")`), in most cases I would think that the stack seems like it would be enough but having a user provided label seems totally fine too.

kategengler:
I think the name is confusing (that it isn't a "pair" with `pauseTest` but something completely different).

rwjblue:
@kategengler - Ya, I can definitely see that point. The rough idea is to ensure that the test systems concept of "settledness" incorporates the promise provided. Any ideas for alternate names?

rwjblue:
@bendemboski - Added an optional label argument to the signature. Also, I changed things around to store in a map with both the provided label and error (for stack) but intentionally did not introduce `getSettledState().debugInfo` yet (I'd like to do that all at once in #554).

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -142,12 +143,52 @@ function checkWaiters() {
...
+
+  @public
+  @param {Promise<T>} promise the promise to wait for
+  @param {string} [label] an optional label for the promise
+  @returns {Promise<T>} the passed in promise
+*/
+export function pauseTestFor<T>(promise: Promise<T>, label?: string): Promise<T> {
+  let error = new Error();
+
+  currentPauseTestPromises.set(promise, { label, error });
```
bendemboski: I guess we have to worry about the case where the same promise is passed in twice, e.g.

```javascript
// my-addon/addon/foo.js
myAddonFunction() {
  let promise = doSomethingAsync();
  pauseTestFor(promise);
  return promise;
}

// my-app/app/bar.js
function doThing() {
  return pauseTestFor(myAddonFunction());
}
```

I know the API is set up to encourage the addon function to `return pauseTestFor(doSomethingAsync())` which would then return a different promise, but it will always be possible to pass the same promise to two different `pauseTestFor()` calls.

I guess doing so wouldn't break the settled behavior -- things would still settle at the same time. The second call would clobber the first call's stack and label, which could potentially be confusing. Alternatively, we could run a `.has()` check and early out and/or issue a warning. It seems like overkill to implement the fully tolerant solution of having the `Set` store an array of `{ label, error }`...thoughts?

bendemboski:
Oh, good point @kategengler. I think it would be really nice to get the word `settled` into the name. Maybe `thisIsUnsettling()`? :trollface:

bendemboski:
Maybe something like `deferSettled()` or `blockSettled()`?

scalvert:
`delaySettledUntil(promise)`?

FML, naming yada yada.

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -142,12 +143,52 @@ function checkWaiters() {
...
+}
+
+let currentPauseTestPromises = new Map<Promise<unknown>, IPauseTestPromiseDetails>();
+
+/**
+  Ensures that tests wait for the provided promise before being considered "settled".
+
+  This function is expected to be used in application/addon code (aka non-test code).
+  When used during a test run, it ensures that the next `await settled()` check does
+  not result until after the promise has settled, when used outside of a test it is a
```
Turbo87:
```suggestion
  not resolve until after the promise has settled, when used outside of a test it is a
```

addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -142,12 +143,52 @@ function checkWaiters() {
...
+  label: undefined | string;
+  error: Error;
+}
+
+let currentPauseTestPromises = new Map<Promise<unknown>, IPauseTestPromiseDetails>();
+
+/**
+  Ensures that tests wait for the provided promise before being considered "settled".
+
+  This function is expected to be used in application/addon code (aka non-test code).
```
Turbo87: I am a little worried about this for several reasons:
- apps and addons don't have an explicit dependency on `@ember/test-helpers` currently
- would addons that use this need to declare it in `dependencies`?
- should it be a peer dependency instead?
- how would that work if apps only have it as a transitive dependency via e.g. `ember-qunit`?
- what if an addon declares it as a dependency, another addon declares it as a dependency with different version and those conflict?

I'm wondering if it would be better to have this in a separate addon until these bundling/dependency issues are resolved 🤔

rwjblue: Yes, I agree that some of this becomes easier in a separate addon, however the ability to debug/track becomes harder. Specifically, the only way a standalone addon could make the pausing behavior work would be via a test waiter. With a classic test waiter situation we would have absolutely no information about _why_ waiters are not settled.

addon-test-support/@ember/test-helpers/settled.ts
rwjblue: Yeah, good point. I guess I just find it unlikely to actually pause on the same promise multiple times. I think the simplest thing would be to add a `.has` check before `.set` that throws an error if the promise is already tracked.

rwjblue:
Replaced by the usage of ember-test-waiters.

PR #597, opened 2019-03-12
Update File Upload Example

jherdman:
The example didn't match the documentation, and now it does.

### DIFF
--- a/API.md
+++ b/API.md
@@ -356,7 +356,7 @@ When using triggerEvent to upload a file the `eventType` must be `change` and yo
 triggerEvent(
   'input.fileUpload',
   'change',
-  [new Blob(['Ember Rules!'])]
+  { files: [new Blob(['Ember Rules!'])] }
 );
 ```
 


API.md
```diff
@@ -356,7 +356,7 @@ When using triggerEvent to upload a file the `eventType` must be `change` and yo
 triggerEvent(
   'input.fileUpload',
   'change',
-  [new Blob(['Ember Rules!'])]
+  { files: [new Blob(['Ember Rules!'])] }
```
rwjblue: Oh, I just realized this changes the `API.md` directly. That file is autogenerated and shouldn't be manually modified.

Its built from the inline API docs, which for this example I think means here:

https://github.com/emberjs/ember-test-helpers/blob/f1927fb4fb396d7e96db4596f650f8aba523cb31/addon-test-support/%40ember/test-helpers/dom/trigger-event.ts#L22-L25

jherdman:
Closing in favour of #598

PR #604, opened 2019-03-17
Move ember-cli-typescript to dependencies

davewasmer:
Right now, ember-cli-typescript is listed as a devDependency.

However, `index.js` [references it directly](https://github.com/emberjs/ember-test-helpers/blob/master/index.js#L38).

Because it is currently listed in devDependencies and not dependencies, npm/yarn won't ensure that `ember-test-helpers` gets it's own copy, leading to the above linked line failing in certain cases.

I seems rather difficult to expose this - I suspect it had something to do with multiple versions of ember-cli-typescript being present in a monorepo I was working on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "broccoli-funnel": "^2.0.2",
     "ember-assign-polyfill": "^2.6.0",
     "ember-cli-babel": "^7.6.0",
-    "ember-cli-htmlbars-inline-precompile": "^2.1.0"
+    "ember-cli-htmlbars-inline-precompile": "^2.1.0",
+    "ember-cli-typescript": "^1.5.0"
   },
   "devDependencies": {
     "@ember/optional-features": "^0.7.0",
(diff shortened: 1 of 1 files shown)


rwjblue:
We should not be directly depending on ember-cli-typescript (devDep is correct). Instead we should fix the conditional to be something like:

```js
    if (this.isDevelopingAddon()) {
      // eslint-disable-next-line node/no-unpublished-require
      let TypescriptOutput = require('ember-cli-typescript/js/lib/incremental-typescript-compiler/typescript-output-plugin');
      // eslint-disable-next-line node/no-unpublished-require
      let MergeTrees = require('broccoli-merge-trees');

      let tsTree = debugTree(
        new TypescriptOutput(compiler, {
          'addon-test-support/@ember/test-helpers': '@ember/test-helpers',
        }),
        'addon-test-support:ts'
      );

      input = debugTree(new MergeTrees([input, tsTree]), 'addon-test-support:merged');
    }
```

davewasmer:
Closing in favor of #610

PR #610, opened 2019-03-25
Only require ember-cli-typescript while developing

davewasmer:
See #604 for context.

Wasn't sure what the `_incrementalTsCompiler` flag was doing - it's not referenced anywhere else in the codebase, so I just removed the reference, but I can add it back if it needs to be something like:

```
if (this.isDevelopingAddon() && this.project._incrementalTsCompiler) {
```


rwjblue:
Thank you @Turbo87!

PR #636, opened 2019-05-03
refactor(DebugInfo): Changes DebugInfo constructor signature to POJO

scalvert:
The signature of the `DebugInfo` constructor used to take a series of booleans. This type of signature becomes very hard to use and read, as a succession of `true`/`false` parameters doesn't help the reader understand what's required.

This PR addresses this by converting the signature to take a POJO, in this case of type `SettledState`.

### DIFF
--- a/addon-test-support/@ember/test-helpers/-internal/debug-info.ts
+++ b/addon-test-support/@ember/test-helpers/-internal/debug-info.ts
@@ -85,22 +85,8 @@ export class TestDebugInfo implements DebugInfo {
   private _debugInfo: MaybeDebugInfo;
   private _summaryInfo: SummaryInfo | undefined = undefined;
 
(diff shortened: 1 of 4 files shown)


scalvert:
This is technically a breaking change, but I think the only external consumer is `ember-qunit`. I'll leave it up to folks here to figure out the version number.

rwjblue:
> This is technically a breaking change

I don't think I agree, the constructor was not exposed as public API (AFAICT only this lib itself was creating instances of DebugInfo)

rwjblue:
Sorry, didn't mean to seem so argumentative there! Indeed changing the constructor args is a breaking change, but since we don't (AFAIK) leak the constructor we should be safe.

scalvert:
👍

PR #663, opened 2019-06-07
Update 'triggerEvent' examples

qpowell:
- Fixes missing line break in first `triggerEvent` example, and adds the `files` key to the `options` object.
- Includes additional example for working with the `drop` event.

### DIFF
--- a/API.md
+++ b/API.md
@@ -308,9 +308,8 @@ Triggers an event on the specified target.
 
 #### Examples
 
-Using `triggerEvent` to upload a file
-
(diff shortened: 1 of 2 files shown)


qpowell:
> Thanks for working on this!
> 
> The `API.md` file is auto-generated from the inline API documentation (in the JS / TS files), would you mind tweaking this to update this file instead?
> 
> https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/@ember/test-helpers/dom/trigger-event.ts

Yeah, no problem!

PR #674, opened 2019-06-28
Ensure that `triggerEvent` can simulate a file input change event with an empty `FileList`

### LINKED ISSUE #673: triggerEvent cannot simulate a file input change event with an empty FileList
triggerEvent cannot simulate the following case:
1. User open the file input selector and selects a file(s).
2. User opens the file input and hits cancel

Step 1 causes the input to fire a 'change' event with `event.target.files` to be a FileList with the selected file.
Step 2 causes the input to fire a 'change' event with `event.target.files` being an empty FileList.

In `fire-event.ts` the method `buildFileEvent` checks that the options.files is an array with at least one member. If options.files is an empty array, it does not update the files property on the event target. Because of this check cannot simulate the situation done as before.

Will try to get a fix & test in today.

ggayowsky:
Fixes #673

### DIFF
--- a/tests/unit/dom/select-files-test.js
+++ b/tests/unit/dom/select-files-test.js
@@ -45,8 +45,8 @@ module('DOM Helper: selectFiles', function(hooks) {
     });
 
     await setupContext(context);
-    await triggerEvent(element, 'change', [textFile]);
-    await triggerEvent(element, 'change', [imageFile]);
+    await triggerEvent(element, 'change', { files: [textFile] });
+    await triggerEvent(element, 'change', { files: [imageFile] });
 
     assert.verifySteps(['change', 'text-file.txt', 'change', 'image-file.png']);
   });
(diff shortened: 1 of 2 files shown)


tests/unit/dom/select-files-test.js
```diff
@@ -45,8 +45,8 @@ module('DOM Helper: selectFiles', function(hooks) {
     });
 
     await setupContext(context);
-    await triggerEvent(element, 'change', [textFile]);
-    await triggerEvent(element, 'change', [imageFile]);
```
rwjblue: I think its 👍 	to change this test to use the non-deprecated path, but would you mind adding a test specifically about the deprecation (passing options as an array)? I just want to make sure we keep that under test until we actually remove support.

ggayowsky: On line 68 there is already a test asserting the deprecation. So I figured it was extra safe to change this one.

rwjblue: Aha, thank you!

ggayowsky:
What is up with the CI failures? I do not think that they were caused by my changes looking through the logs.

rwjblue:
The failures on canary were due to an upstream issue that is fixed (https://github.com/emberjs/ember.js/pull/18168). I've restarted the failed tests...

ggayowsky:
And looks like the CI failed with seg faults. This looks like a problem with the CI and not my code?

rwjblue:
Yeah, I do not believe they are related.

ggayowsky:
Can I get the CI kicked off again?

rwjblue:
I'm trying to dig into the CI issues with canary Ember (important now since we will be rolling those to beta next week), the failures are definitely not related to your changes.

ggayowsky:
Oh great! Thank you for the information!

rwjblue:
OK, I think CI is fixed now just rebased and updated this branch.

Issue #680, opened 2019-07-09
`fillIn` ignores the `disabled` property on inputs


jenweber:
Even if an input is disabled, fillIn still lets you assign a value. I think this is a bug.

[Reproduction](https://github.com/jenweber/fillin-bug-repro) - see the [template with a disabled input](https://github.com/jenweber/fillin-bug-repro/blob/master/app/templates/application.hbs) and a [test that should not pass](https://github.com/jenweber/fillin-bug-repro/blob/master/tests/acceptance/inputs-test.js)

ro0gr:
same issue exists for readonly inputs, which may lead to false positive tests

rwjblue:
Ya, definitely would love to fix.

@ro0gr - Would you mind reviewing that PR to see if it also fixes that case?

ro0gr:
oh, I haven't noticed the PR. Sure, I'll check it tomorrow!

PR #681, opened 2019-07-09
bugfix - fillIn now reflects disabled attribute #680

jenweber:
See #680 for a minimal reproduction and explanation of the bug.

Changed
- if you try to fillIn a disabled input, it throws an error "Selected element is disabled"

Added:
- a test for the new behavior

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fill-in.ts
+++ b/addon-test-support/@ember/test-helpers/dom/fill-in.ts
@@ -42,6 +42,10 @@ export default function fillIn(target: Target, text: string): Promise<void> {
       throw new Error('Must provide `text` when calling `fillIn`.');
     }
 
+    if (element.disabled) {
+      throw new Error('Selected element is disabled')
+    }
+
     __focus__(element);
 
     if (isControl) {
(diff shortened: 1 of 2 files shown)


jenweber:
I'll fix my missing `;` and add some similar checks for typeIn and triggerKeyEvent. I think we should do the same for click and focus events, but will put them in a separate PR.

rwjblue:
Ya, my concern RE: click/focus is that its _probably_ a breaking bugfix.

rwjblue:
@jenweber - Mind rebasing (I think that will fix CI)?

ro0gr:
> I think we need the same treatment for typeIn and triggerKeyEvent. We should also consider if we should do similar in click/doubleClick/focus/etc...

Seems like most of the mentioned helpers(except of `triggerKeyEvent`) go through the `__focus__`. I I think it looks like a good single point to throw in case if non-interactable element is passed as target.

rwjblue:
Ya, that makes sense to me.

ro0gr:
I'll check if it's feasible

btecu:
Same issue exists for `readOnly`.

rwjblue:
Closing in favor of @ro0gr's updated PR (which incorporates this work): #741 and #779.

PR #685, opened 2019-07-17
Default `click` and `doubleClick` to be left button clicks.

evanfarina:
Prior to this change, neither click nor double-click helpers were passing the `button` and `buttons` mouseEvent options necessary to emit a left click. I recently ran into an issue as a library I'm using was specifically looking for 'left' clicks and it dawned on me that this should likely be the default behavior for these helpers.

Question for public / repo owners: It's possible, albeit rather unlikely, that this is a breaking change for some people. I'm open to feedback regarding the best way to handle this change so that it does not negatively impact users of this addon.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/click.ts
+++ b/addon-test-support/@ember/test-helpers/dom/click.ts
@@ -1,3 +1,4 @@
+import { assign } from '@ember/polyfills';
 import getElement from './-get-element';
 import fireEvent from './fire-event';
 import { __focus__ } from './focus';
@@ -8,10 +9,22 @@ import isFormControl from './-is-form-control';
 import Target from './-target';
 import { log } from '@ember/test-helpers/dom/-logging';
 
+const PRIMARY_BUTTON = 1;
+const MAIN_BUTTON_PRESSED = 0;
+
+/**
+ * Represent a particular mouse button being clicked.
+ * See https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons for available options.
+ */
+export const DEFAULT_CLICK_OPTIONS = {
+  buttons: PRIMARY_BUTTON,
+  button: MAIN_BUTTON_PRESSED,
+};
+
 /**
   @private
   @param {Element} element the element to click on
-  @param {Object} options the options to be merged into the mouse events
+  @param {MouseEventInit} options the options to be merged into the mouse events
 */
 export function __click__(element: Element | Document, options: MouseEventInit): void {
   fireEvent(element, 'mousedown', options);
@@ -53,7 +66,7 @@ export function __click__(element: Element | Document, options: MouseEventInit):
 
   @public
   @param {string|Element} target the element or selector to click on
-  @param {Object} options the options to be merged into the mouse events
+  @param {MouseEventInit} _options the options to be merged into the mouse events.
   @return {Promise<void>} resolves when settled
 
   @example
@@ -69,8 +82,10 @@ export function __click__(element: Element | Document, options: MouseEventInit):
 
   click('button', { shiftKey: true });
 */
-export default function click(target: Target, options: MouseEventInit = {}): Promise<void> {
+export default function click(target: Target, _options: MouseEventInit = {}): Promise<void> {
   log('click', target);
+
+  let options = assign({}, DEFAULT_CLICK_OPTIONS, _options);
 
   return nextTickPromise().then(() => {
     if (!target) {
--- a/addon-test-support/@ember/test-helpers/dom/double-click.ts
+++ b/addon-test-support/@ember/test-helpers/dom/double-click.ts
@@ -1,9 +1,11 @@
+import { assign } from '@ember/polyfills';
 import getElement from './-get-element';
 import fireEvent from './fire-event';
 import { __focus__ } from './focus';
 import settled from '../settled';
 import isFocusable from './-is-focusable';
 import { nextTickPromise } from '../-utils';
+import { DEFAULT_CLICK_OPTIONS } from './click';
 import Target from './-target';
 import { log } from '@ember/test-helpers/dom/-logging';
 import isFormControl from './-is-form-control';
@@ -11,7 +13,7 @@ import isFormControl from './-is-form-control';
 /**
   @private
   @param {Element} element the element to double-click on
-  @param {Object} options the options to be merged into the mouse events
+  @param {MouseEventInit} options the options to be merged into the mouse events
 */
 export function __doubleClick__(element: Element | Document, options: MouseEventInit): void {
   fireEvent(element, 'mousedown', options);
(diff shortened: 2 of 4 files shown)


addon-test-support/@ember/test-helpers/dom/click.ts
```diff
@@ -69,6 +70,8 @@ export function __click__(element: Element | Document, options: MouseEventInit):
   click('button', { shiftKey: true });
 */
 export default function click(target: Target, options: MouseEventInit = {}): Promise<void> {
+  const mergedOptions = assign({ buttons: 1, button: 0 }, options);
```
scalvert: Can the `buttons` and `button` values have variable representations to improve readability?

```js
const PRIMARY_BUTTON = 1;
const MAIN_BUTTON_PRESSED = 0;

//...

const mergedOptions = assign({ buttons: PRIMARY_BUTTON, button: MAIN_BUTTON_PRESSED }, options);
```

addon-test-support/@ember/test-helpers/dom/double-click.ts
```diff
@@ -80,6 +81,8 @@ export function __doubleClick__(element: Element | Document, options: MouseEvent
   doubleClick('button', { shiftKey: true });
 */
 export default function doubleClick(target: Target, options: MouseEventInit = {}): Promise<void> {
+  const mergedOptions = assign({ buttons: 1, button: 0 }, options);
```
scalvert: Same as my previous comment above re: variables to represent the numbers.

addon-test-support/@ember/test-helpers/dom/click.ts
```diff
@@ -52,7 +53,7 @@ export function __click__(element: Element | Document, options: MouseEventInit):
 
   @public
   @param {string|Element} target the element or selector to click on
-  @param {Object} options the options to be merged into the mouse events
+  @param {Object} options the options to be merged into the mouse events.
```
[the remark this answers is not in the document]
evanfarina: Updated here, and also in the private __click__ method. Will do the same for doubleClick.

addon-test-support/@ember/test-helpers/dom/double-click.ts
evanfarina: Added a `-constants` file

addon-test-support/@ember/test-helpers/dom/click.ts
evanfarina: Added a constants file to handle this.

addon-test-support/@ember/test-helpers/dom/click.ts
```diff
@@ -69,6 +71,8 @@ export function __click__(element: Element | Document, options: MouseEventInit):
   click('button', { shiftKey: true });
 */
 export default function click(target: Target, options: MouseEventInit = {}): Promise<void> {
+  const mergedOptions = assign({ buttons: MOUSE_EVENT_BUTTONS.PRIMARY }, options);
```
rwjblue: Should we explicitly default `button` also? The tests for it that existed previously passed because `button` is `0` when it is not initialized, but adding it to the `DEFAULT_CLICK_OPTIONS` explicitly seems much easier to reason about to me.

https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button

addon-test-support/@ember/test-helpers/-constants.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
rwjblue: Lets move this into `click.ts` and export from there (still allows us to share, but doesn't require this extra mental step of digging into a separate file)

addon-test-support/@ember/test-helpers/dom/double-click.ts
```diff
@@ -80,6 +82,9 @@ export function __doubleClick__(element: Element | Document, options: MouseEvent
   doubleClick('button', { shiftKey: true });
 */
 export default function doubleClick(target: Target, options: MouseEventInit = {}): Promise<void> {
```
rwjblue: Lets call this `_options` and not initialize it

```suggestion
export default function doubleClick(target: Target, _options: MouseEventInit): Promise<void> {
```

addon-test-support/@ember/test-helpers/dom/double-click.ts
```diff
@@ -80,6 +82,9 @@ export function __doubleClick__(element: Element | Document, options: MouseEvent
...
+
+  const mergedOptions = assign({ buttons: MOUSE_EVENT_BUTTONS.PRIMARY }, options);
```
rwjblue: Can we make a constant above for `DEFAULT_CLICK_OPTIONS` and tweak this to:

```suggestion
  const options = assign({}, DEFAULT_CLICK_OPTIONS, _options);
```

addon-test-support/@ember/test-helpers/dom/double-click.ts
```diff
@@ -90,7 +95,7 @@ export default function doubleClick(target: Target, options: MouseEventInit = {}
       throw new Error(`Element not found when calling \`doubleClick('${target}')\`.`);
     }
 
-    __doubleClick__(element, options);
+    __doubleClick__(element, mergedOptions);
```
rwjblue:
```suggestion
    __doubleClick__(element, options);
```

addon-test-support/@ember/test-helpers/dom/click.ts
```diff
@@ -69,6 +71,8 @@ export function __click__(element: Element | Document, options: MouseEventInit):
   click('button', { shiftKey: true });
 */
 export default function click(target: Target, options: MouseEventInit = {}): Promise<void> {
```
rwjblue:
```suggestion
export default function click(target: Target, _options: MouseEventInit): Promise<void> {
```

addon-test-support/@ember/test-helpers/dom/click.ts
rwjblue: Using the suggestion I made above, this would look like:

```suggestion
  const mergedOptions = assign({}, DEFAULT_CLICK_OPTIONS, _options);
```

addon-test-support/@ember/test-helpers/dom/click.ts
```diff
@@ -82,7 +86,7 @@ export default function click(target: Target, options: MouseEventInit = {}): Pro
     let isDisabledFormControl = isFormControl(element) && element.disabled;
 
     if (!isDisabledFormControl) {
-      __click__(element, options);
+      __click__(element, mergedOptions);
```
rwjblue:
```suggestion
      __click__(element, options);
```

rwjblue:
Thanks for working on this!

addon-test-support/@ember/test-helpers/dom/click.ts
evanfarina: I had `button` originally (see first commit) but felt it was unnecessary. I will add it back in for clarity.

rwjblue:
FYI - Looks like CI failures are due to changes in canary builds of Ember. I'll try to confirm and we can file a separate issue to track fixing those failures.

rwjblue:
Rebased and fixed conflicts, should be ready to go now.

PR #702, opened 2019-08-31
Fix issues due to ember-test-waiters setting up a legacy test waiter.

rwjblue:


tests/unit/settled-test.js
```diff
@@ -395,7 +395,14 @@ module('settled', function(hooks) {
         debugInfo: new TestDebugInfo({
           hasPendingTimers: false,
           hasRunLoop: false,
-          hasPendingLegacyWaiters: false,
+
+          // this is true due to
+          // https://github.com/rwjblue/ember-test-waiters/pull/13 (and
+          // https://github.com/rwjblue/ember-test-waiters/pull/17), but should
+          // actually be false once ember-test-waiters drops suppose for legacy
```
scalvert: “drops support”

rwjblue: eeeck, good catch thank you

PR #707, opened 2019-09-13
The typeIn helper should not wait for settled promises between inputs

jakebixbyavalara:
The current behavior of the `typeIn` helper waits for settled promises between text inputs/key events. This is not ideal because it prevents the helper from being used for testing things like debounced functions (like in a typeahead component for instance).

[Here is a repro repo demonstrating the issue.](https://github.com/jakebixbyavalara/concurrency-debounce-repro), I originally encountered the behavior using an `ember-concurrency` `timeout` to debounce a task, which worked fine in the app, but began failing in test due to the change to the `triggerKeyEvent` helper.

This PR fixes the issue by making the `triggerKeyEvent` helper optionally wait for settled promises via a boolean. It's a simple fix that may be overlooking something, so please let me know if there would be a more preferred solution.

_Initial PR text (when it was just a failing test PR):_
> Previous to [this PR](https://github.com/emberjs/ember-test-helpers/pull/465), the `typeIn` helper would not wait for settled promises, allowing the helper to work with things like `run.debounce` or ember-concurrency `timeout` in test. Now the helper waits for any unsettled promises to settle between input because it now uses the `triggerKeyEvent ` (which waits for settled promises).

> Added failing test for typeIn waiting for other promises, and will try and take a crack at fixing it as well. I mainly wanted to put this PR up for eyes and advice if I hit any walls.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/trigger-key-event.ts
+++ b/addon-test-support/@ember/test-helpers/dom/trigger-key-event.ts
@@ -2,6 +2,7 @@ import { assign } from '@ember/polyfills';
 import getElement from './-get-element';
 import fireEvent from './fire-event';
 import settled from '../settled';
+import { resolve } from 'rsvp';
 import { KEYBOARD_EVENT_TYPES, KeyboardEventType, isKeyboardEventType } from './fire-event';
 import { nextTickPromise, isNumeric } from '../-utils';
 import Target from './-target';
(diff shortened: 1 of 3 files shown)


jakebixbyavalara:
@rwjblue That works for me, I appreciate the refactor!

jakebixbyavalara:
@rwjblue just wanted to check in, did you need me to do any additional work on this PR?

rwjblue:
We had an issue with Node 6 that we just resolved. I just rebased on top of those fixes, and I think that CI _should_ be happy now...

Assuming it is, we can land and release.

Issue #721, opened 2019-10-18
Proposal: API for adding test metadata


scalvert:
Having a mechanism for adding metadata could be quite a useful addition. Information such as test name, test ID, type, and any additional information that can be added to metadata (such as what specific setups were run) that when composed tells a fuller picture about the composition of the test.

One of the key purposes of this addition would be to provide better annotations for the test characteristics in order to log and analyze them in a more parsable and queryable way.  

Details
-------
The potential API for this can be quite simple:

Proposed interface:

```ts
interface ITestMetadata {
  testName: string | null;
  types: string[];
  [key: string]: any;
}
```

Proposed API:

```ts
getTestMetadata(context: BaseContext): ITestMetadata; 
```

The `getTestMetadata` function would be backed by a `WeakMap`, which would associate the metadata with the test context. Accessing this for the first time, when a prior test context hadn't been added, would lazy initialize and add a test metadata object to the `WeakMap`.

Usage
-----

*Use case: adding test name and test ID in `ember-qunit`*

In [`ember-qunit`](https://github.com/emberjs/ember-qunit/blob/master/addon-test-support/ember-qunit/index.js#L138):

```js
QUnit.testStart(({ name, testId }) => {
  let testMetadata = getTestMetadata(QUnit.config.current);
  testMetadata.testName = name;
  testMetadata.testId = testId;

  Ember.testing = true;
});
```

*Use case: adding metadata for test type*

In `setupRenderingContext`:

```ts
let testMetadata: ITestMetadata = getTestMetadata(context);

testMetadata.types.push('rendering');
```

*Use case: adding metadata for additional setup functions that are invoked as part of the test setup*

In `setupMirage`:

```ts
let testMetadata: ITestMetadata = getTestMetadata(context);

testMetadata.types.push('mirage');
```

*Use case: adding metadata to attribute a test to a sub-team*

```ts
let testMetadata: ITestMetadata = getTestMetadata(context);

testMetadata.team = 'Sub team name';
```

Being able to provide arbitrary values to add to the metadata would allow for a multitude of scenarios.

rwjblue:
This sounds like a great addition, and would check off yet another feature that was originally mentioned in the "Grand Testing Unification" RFC (https://github.com/emberjs/rfcs/pull/119/files#diff-678318af322c33d1538f2e43a85ff5e8R325).

Turbo87:
what do we need this for? how will this help the user in the end?

scalvert:
[...]

PR #724, opened 2019-10-18
feat(test metadata): Adds test metadata primitives

scalvert:
Adds basic API for #721

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-context.ts
+++ b/addon-test-support/@ember/test-helpers/setup-context.ts
@@ -12,6 +12,7 @@ import global from './global';
 import { getResolver } from './resolver';
 import { getApplication } from './application';
 import { nextTickPromise } from './-utils';
+import getTestMetadata, { ITestMetadata } from './test-metadata';
 
 export interface BaseContext {
   [key: string]: any;
@@ -160,6 +161,9 @@ export default function setupContext(
 
   let contextGuid = guidFor(context);
   CLEANUP[contextGuid] = [];
+
+  let testMetadata: ITestMetadata = getTestMetadata(context);
+  testMetadata.setupTypes.push('setupContext');
 
   run.backburner.DEBUG = true;
 
--- a/addon-test-support/@ember/test-helpers/test-metadata.ts
+++ b/addon-test-support/@ember/test-helpers/test-metadata.ts
@@ -1,0 +1,34 @@
+import { BaseContext } from './setup-context';
+
+export interface ITestMetadata {
+  testName?: string;
+  types: string[];
+  [key: string]: any;
+}
+
+export class TestMetadata implements ITestMetadata {
+  [key: string]: any;
+  testName?: string;
+  types: string[];
+
+  constructor() {
+    this.types = [];
+  }
+}
+
+const TEST_METADATA = new WeakMap<BaseContext, ITestMetadata>();
+
+/**
+ * Gets the test metadata associated with the provided test context. Will create
+ * a new test metadata object if one does not exist.
+ *
+ * @param {BaseContext} context the context to use
+ * @returns {ITestMetadata} the test metadata for the provided context
+ */
+export default function getTestMetadata(context: BaseContext): ITestMetadata {
+  if (!TEST_METADATA.has(context)) {
+    TEST_METADATA.set(context, new TestMetadata());
+  }
+
+  return TEST_METADATA.get(context)!;
+}
--- a/addon-test-support/@ember/test-helpers/test-metadata.ts
+++ b/addon-test-support/@ember/test-helpers/test-metadata.ts
@@ -2,17 +2,34 @@ import { BaseContext } from './setup-context';
 
 export interface ITestMetadata {
   testName?: string;
-  types: string[];
+  setupTypes: string[];
+  usedHelpers: string[];
   [key: string]: any;
+
+  readonly isRendering: boolean;
+  readonly isApplication: boolean;
 }
 
 export class TestMetadata implements ITestMetadata {
   [key: string]: any;
   testName?: string;
-  types: string[];
+  setupTypes: string[];
+  usedHelpers: string[];
 
   constructor() {
-    this.types = [];
+    this.setupTypes = [];
+    this.usedHelpers = [];
+  }
+
+  get isRendering() {
+    return (
+      this.setupTypes.indexOf('setupRenderingContext') > -1 &&
+      this.usedHelpers.indexOf('render') > -1
+    );
+  }
+
+  get isApplication() {
+    return this.setupTypes.indexOf('setupApplicationContext') > -1;
   }
 }
 
--- a/tests/unit/test-metadata-test.js
+++ b/tests/unit/test-metadata-test.js
@@ -8,7 +8,7 @@ module('Test Metadata', function() {
     let testMetadata = getTestMetadata(test);
 
     assert.ok(testMetadata instanceof TestMetadata);
-    assert.deepEqual(testMetadata.types, []);
+    assert.deepEqual(testMetadata.setupTypes, []);
   });
 
   module('Annotated Test Metadata', function(hooks) {
--- a/tests/unit/test-metadata-test.js
+++ b/tests/unit/test-metadata-test.js
@@ -1,0 +1,31 @@
+import QUnit, { module, test } from 'qunit';
+import { getTestMetadata } from '@ember/test-helpers';
+import { TestMetadata } from '@ember/test-helpers/test-metadata';
+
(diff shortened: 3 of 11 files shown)


addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseContext } from './setup-context';
+
+export interface ITestMetadata {
```
scalvert: It mildly disturbs me that all other interfaces in this project don't use the 'I' prefix, but I also find it very irritating that you can't define an interface and class of the same name (for obvious reasons), which forces you to come up with some awful name just to accommodate the mismatch.

tests/unit/test-metadata-test.js
```diff
@@ -0,0 +1,31 @@
...
+    let test = QUnit.config.current;
+    let testMetadata = getTestMetadata(test);
+
+    assert.ok(testMetadata instanceof TestMetadata);
+    assert.deepEqual(testMetadata.types, []);
+  });
+
+  module('Annotated Test Metadata', function(hooks) {
+    hooks.beforeEach(function() {
+      let test = QUnit.config.current;
```
rwjblue: Same as above, we should probably use `this` instead of `QUnit.config.current` (since `current` is the `Test` instance, not the `this` context)

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,27 @@
...
+  testName: string | null;
+  types: string[];
+  [key: string]: any;
+}
+
+export class TestMetadata implements ITestMetadata {
+  [key: string]: any;
+  testName!: string | null;
```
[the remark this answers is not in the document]
scalvert: Seems fine. I was expecting that `testName` would always be provided (though not at instantiation, obviously), but I'm not opinionated either way.

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,27 @@
...
+  types: string[];
+
+  constructor() {
+    this.types = [];
+  }
+}
+
+const testMetadata = new WeakMap<BaseContext, ITestMetadata>();
```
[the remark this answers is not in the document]
scalvert: I usually only capitalize classes in this manner. If we want a const I'd use `TEST_METADATA`. You ok with the latter?

tests/unit/test-metadata-test.js
```diff
@@ -0,0 +1,31 @@
+import QUnit, { module, test } from 'qunit';
+import { getTestMetadata } from '@ember/test-helpers';
+import { TestMetadata } from '@ember/test-helpers/test-metadata';
+
+module('Test Metadata', function() {
+  test('getTestMetadata returns default test metadata', function(assert) {
+    let test = QUnit.config.current;
```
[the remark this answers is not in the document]
scalvert: Oof you're right. Changing.

tests/unit/test-metadata-test.js
scalvert: Same as above.

addon-test-support/@ember/test-helpers/test-metadata.ts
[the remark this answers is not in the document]
scalvert: I was following this convention: https://github.com/ember-lifeline/ember-lifeline/blob/master/addon/run-task.ts#L23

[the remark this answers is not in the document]
scalvert: I think if we feel this is truly const, it should probably be `ALL_CAPS`

rwjblue:
Should be able to rebase now that #725

addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -161,6 +162,9 @@ export default function setupContext(
   let contextGuid = guidFor(context);
   CLEANUP[contextGuid] = [];
 
+  let testMetadata: ITestMetadata = getTestMetadata(context);
+  testMetadata.types.push('container');
```
rwjblue: :sob: I don’t like this word for these type of tests :sad:

scalvert: Open to suggestions, but that's what the docs call them :)

https://guides.emberjs.com/release/testing/#toc_container-tests

scalvert: @rwjblue and I spoke offline. We decided that we'd rename `types` to `setupTypes` in order to capture the functions run during setup. We're also adding a `usedHelpers` to indicate whether rendering or visits had occurred.

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+  usedHelpers: string[];
+
+  constructor() {
+    this.setupTypes = [];
+    this.usedHelpers = [];
+  }
+
+  get isRendering() {
+    return (
+      this.setupTypes.length === 2 &&
```
rwjblue: I don't think this is quite right, we shouldn't _require_ that the length is 2 (IOW, why do we care of `this.setupTypes` is `['setupContext', 'setupRenderingContext', 'setupMirage', 'setupEngineContext']`)?
```suggestion
```

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+      this.setupTypes.includes('setupRenderingContext') &&
```
rwjblue: Sadly, we can't rely on `.includes`:

```suggestion
      this.setupTypes.indexOf('setupRenderingContext') > -1 &&
```

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+      this.usedHelpers.includes('render')
```
rwjblue:
```suggestion
      this.usedHelpers.indexOf('render') > -1
```

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+    );
+  }
+
+  get isApplication() {
+    return this.setupTypes.length === 2 && this.setupTypes.includes('setupApplicationContext');
```
rwjblue:
```suggestion
    return this.setupTypes.indexOf('setupApplicationContext') > -1;
```

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param {BaseContext} context the context to use
+ * @returns {ITestMetadata} the test metadata for the provided context
+ */
+export default function getTestMetadata(context: BaseContext): ITestMetadata {
+  if (!TEST_METADATA.has(context)) {
+    TEST_METADATA.set(context, new TestMetadata());
+  }
+
+  return TEST_METADATA.get(context)!;
```
rwjblue: Should we tweak this to something like:

```js
let testMetadata = TEST_METADATA.get(context);

if (testMetadata === undefined) {
  testMetadata = new TestMetadata();
  TEST_METADATA.set(context, testMetadata);
}

return testMetadata;
```

scalvert: Six of one, half dozen yada yada...

addon-test-support/@ember/test-helpers/test-metadata.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export class TestMetadata implements ITestMetadata {
+  [key: string]: any;
+  testName?: string;
+  setupTypes: string[];
+  usedHelpers: string[];
+
+  constructor() {
+    this.setupTypes = [];
```
choheekim: Other than application and rendering tests do we consider others as unit tests or container tests?

scalvert: So here are the classifications:
- Plain QUnit tests that don't use any features from `@ember/test-helpers`, including what we'd consider `unit` right now (`setupTypes` and `usedHelpers` are both completely empty)
- What is termed 'container' tests, which is effectively all tests built using `setupContext` within `@ember/test-helpers` (`setupTypes` does not contain `setupRenderingContext` and `setupApplicationContext`)
- rendering (covered already)
- application (covered already)

We only want to be explicit about `rendering` and `application`, as we have a plan to simplify the terminology of these tests types down the road.

Issue #692, opened 2019-08-06
Focusout event not triggered when test window is not focused


schorke:
So, I have a simple Ember component input which has the built-in `focusOut` Ember event. In an integration test, using `ember-test-helpers` `triggerEvent(input, 'blur')` on this input or `fillIn(other-input, 'blbl')` in another input does not trigger the `focusOut` event if the window test is not focused while running.

I created a new Ember project to demonstrate this bug (see `tests/integration/my-component-test.js`), you can check it out here => [test-focusout](https://github.com/schorke/test-focusout)

### How to test
1. `ember t -s --filter='Integration | Component | my-component'`
2. Open dev tools in test window.
3. Click in console and refresh page (click inside test window if you want to re-focusin).
4. Same behaviour when running the test while focusing another window.
Reproduced on Chrome 76.0.3809.87 and Firefox 68.0.1

Is that the expected behaviour ?

rwjblue:
> Is that the expected behaviour ?

Sadly, I believe it is. Specifically, we can't always trigger focus events when the window itself doesn't have focus.  The best way to confirm this would be to make a simpler reproduction with plain HTML / JS (no Ember, no jQuery, etc) and see if we can dispatch the events when the window isn't focused. If we **can** then we can likely make it work here too.

trek:
Not that it's a fix, but we worked around this by adding a `testWithBrowserFocus` wrapper for `test` that we use in these cases.

Here we preferred false positives.

```
import { test } from 'qunit';
/*
  Tests that need to verify HTMLElement.focus() behavior will
  fail when run in a window that isn't focused, but pass correctly
  when the window is focused or running headless (e.g. in CI).

  This helper checks if the window is focused. If it's not,
  the test will pass.
*/
export function testWithBrowserFocus(description, callback) {
  test(description, async function(assert) {
    if (!document.hasFocus()) {
      assert.ok(true);
      return;
    }
    await callback.call(this, assert);
  });
}
```

And it works just like `test`:

```
  testWithBrowserFocus('opening a chat marks it as read', async function(assert) {});

```

ohcibi:
I noticed that the current `fillIn` implementation does in fact fire `focus` and `focusin` but no `focusout` event. Is there a reason why the first two are fired but not the out event?

rwjblue:
@ohcibi - I thats a different issue than the one reported here, would you mind reporting that separately? I think we should fix...

rwjblue:
[...]

PR #728, opened 2019-10-26
Add `select()` helper

### LINKED ISSUE #708: How to select multiple?
Say that I have a `<select multiple>...</select>` element. The `fillIn` helper doesn't seem to work as it sets the `value` property. `click` just opens the dropdown.

Mikek2252:
Resolves #708 

Added a test helper to select a single value or multiple values for a HTMLSelectElement.


Mikek2252:
@rwjblue Hey sorry to bother you, i think this PR should pass because i did a travis build on my fork and it passed fine. Would it be possible to delete the cache and re run the tests

addon-test-support/@ember/test-helpers/dom/-is-select-element.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+  @private
+  @param {Element} element the element to check
+  @returns {boolean} `true` when the element is a select element, `false` otherwise
+*/
+export default function isSelectElement(element: Element | Document): element is HTMLSelectElement {
+  return (
+    !isDocument(element) &&
+    element.tagName === 'SELECT' &&
+    (element as HTMLInputElement).type !== 'hidden'
```
Turbo87: why does this fail for hidden selects?

and shouldn't this be a `HTMLSelectElement` instead?

addon-test-support/@ember/test-helpers/dom/select.ts
```diff
@@ -0,0 +1,74 @@
...
+  return nextTickPromise().then(() => {
+    if (!target) {
+      throw new Error('Must pass an element or selector to `select`.');
+    }
+
+    if (typeof options === 'undefined' || options === null) {
+      throw new Error('Must provide an `option` or `options` to select when calling `select`.');
+    }
+
+    const element = getElement(target) as any;
```
Turbo87: why do we cast this to `any` here?

addon-test-support/@ember/test-helpers/dom/select.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    if (element.disabled) {
+      throw new Error('Element is disabled');
+    }
+
+    options = Array.isArray(options) ? options : [options];
+
+    if (!element.multiple && options.length > 1) {
+      throw new Error(
+        'HTMLSelectElement multiple attribute is set to false but multiple options have been passed'
```
Turbo87:
```suggestion
        'HTMLSelectElement `multiple` attribute is set to `false` but multiple options have been passed'
```

addon-test-support/@ember/test-helpers/dom/select.ts
```diff
@@ -0,0 +1,74 @@
...
+      );
+    }
+
+    __focus__(element);
+
+    for (let i = 0; i < element.options.length; i++) {
+      let elementOption = element.options.item(i);
+      if (options.indexOf(elementOption.value) > -1) {
+        elementOption.selected = true;
+      }
```
Turbo87: shouldn't we deselect the ones that are not matching the `options`?

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,146 @@
+import { module, test } from 'qunit';
+import { select, setupContext, teardownContext } from '@ember/test-helpers';
+import { buildInstrumentedElement } from '../../helpers/events';
+import hasEmberVersion from '@ember/test-helpers/has-ember-version';
+import { isIE11 } from '../../helpers/browser-detect';
+
+let clickSteps = ['focus', 'focusin', 'input', 'change'];
+
+if (isIE11) {
+  clickSteps = ['focusin', 'input', 'change', 'focus'];
```
Turbo87: did you test this?

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+module('DOM Helper: select', function(hooks) {
+  if (!hasEmberVersion(2, 4)) {
```
Turbo87: why are we skipping this for older Ember versions?

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,146 @@
...
+
+  let context, element;
+
+  hooks.beforeEach(function() {
+    context = {};
+  });
+
+  hooks.afterEach(async function() {
+    element.setAttribute('data-skip-steps', true);
+    if (element) {
```
Turbo87: how could this condition fail if we call a method on the `element` above?

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+    if (context.owner) {
+      await teardownContext(context);
+    }
+
+    document.getElementById('ember-testing').innerHTML = '';
+  });
+
+  test('select without target', async function(assert) {
+    element = buildInstrumentedElement('div');
```
Turbo87: do we need the `element` for this test?

addon-test-support/@ember/test-helpers/dom/-is-select-element.ts
Mikek2252: Good spot on the Element type, as for the hidden I used the existing fill in component for the logic to keep it consistent. If you feel it isnt needed then i am happy to remove. But i wanted to keep it as close to how the other helper worked

addon-test-support/@ember/test-helpers/dom/select.ts
Mikek2252: I used the cast as any because it can return `Element | Document | null` also this is also how it is handled in fillIn so thought it would be best to keep the same for clarity

addon-test-support/@ember/test-helpers/dom/select.ts
Mikek2252: Hmm perhaps, i thought about it but i feels like it is a little bit of a side effect that isnt necessarily expected. But considering that is no reset function maybe a flag to clear would be a good idea? What do you think?

tests/unit/dom/select-test.js
Mikek2252: 😅 Im guilty of stealing this from another test, I'll remove this

tests/unit/dom/select-test.js
Mikek2252: Fair point, this seems to be in all the unit test, happy to remove, but perhaps we should create a issue to remove this.

tests/unit/dom/select-test.js
Mikek2252: Oops, will remove that

tests/unit/dom/select-test.js
Mikek2252: I haven't personally, but again this was on the other tests so felt that it was appropriate to add to this.

addon-test-support/@ember/test-helpers/dom/select.ts
Turbo87: `null` will be caught by the following condition, and `isSelectElement()` should automatically cast it down to a `HTMLSelectElement`

addon-test-support/@ember/test-helpers/dom/select.ts
Turbo87: I'm not sure if we need a flag. IMHO `clearPreviouslySelected` should be the default.

tests/unit/dom/select-test.js
Turbo87: we absolutely have to ensure that this is the right event order in the browsers. please do make sure that this is correct, otherwise the test helper will behave incorrectly.

addon-test-support/@ember/test-helpers/dom/-is-select-element.ts
```diff
@@ -0,0 +1,14 @@
+import { isDocument } from './-target';
+
+/**
+  @private
+  @param {Element} element the element to check
```
Turbo87:
```suggestion
  @param {Element|Document} element the element to check
```

addon-test-support/@ember/test-helpers/dom/select.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (typeof options === 'undefined' || options === null) {
+      throw new Error('Must provide an `option` or `options` to select when calling `select`.');
+    }
+
+    const element = getElement(target) as any;
+    if (!element) {
+      throw new Error(`Element not found when calling \`select('${target}')\`.`);
+    }
+    const isSelect = isSelectElement(element);
+    if (!isSelect) {
```
rwjblue: I think we can collapse this into a single conditional (e.g. no variable)

addon-test-support/@ember/test-helpers/dom/select.ts
rwjblue: > I'm not sure if we need a flag. IMHO clearPreviouslySelected should be the default.

Agree

tests/unit/dom/select-test.js
rwjblue: @Turbo87 - Because `setupContext` / `teardownContext` only work when >= Ember 2.4, and these tests (really all of these dom related tests) use `setupContext`.

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,209 @@
+import { module, test } from 'qunit';
+import { select, setupContext, teardownContext } from '@ember/test-helpers';
+import { buildInstrumentedElement } from '../../helpers/events';
+import hasEmberVersion from '@ember/test-helpers/has-ember-version';
+import { isIE11 } from '../../helpers/browser-detect';
+
+let clickSteps = ['focus', 'focusin', 'input', 'change'];
```
rwjblue: Can you rename this variable? 
```suggestion
let selectSteps = ['focus', 'focusin', 'input', 'change'];
```

addon-test-support/@ember/test-helpers/dom/-is-select-element.ts
```diff
@@ -0,0 +1,14 @@
...
+  @returns {boolean} `true` when the element is a select element, `false` otherwise
+*/
+export default function isSelectElement(element: Element | Document): element is HTMLSelectElement {
+  return (
+    !isDocument(element) &&
+    element.tagName === 'SELECT' &&
+    (element as HTMLSelectElement).type !== 'hidden'
```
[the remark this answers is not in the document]
Mikek2252: Hey sorry for the late reply, I added this because I wanted to keep it as close to how the fillIn in currently works and i noticed on the form-control helper also used hidden so i thought i might be expected behaviour that interfacing with hidden elements is to be discouraged. But happy to remove

Mikek2252:
@rwjblue @Turbo87 A little late but i have updated based off your comments, will run the tests on windows machine to double check that all tests pass on IE

Mikek2252:
@rwjblue @Turbo87 I check this on windows and it is passing so think I have covered any comments, would it be possible to get this checked?

addon-test-support/@ember/test-helpers/dom/select.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (element.disabled) {
+      throw new Error('Element is disabled');
+    }
+
+    options = Array.isArray(options) ? options : [options];
+
+    if (!element.multiple && options.length > 1) {
+      throw new Error(
+        'HTMLSelectElement `multiple` attribute is set to `false` but multiple options have been passed'
```
rwjblue, line 63: Can you add the `target` here to this error message?

addon-test-support/@ember/test-helpers/dom/select.ts
```diff
@@ -0,0 +1,85 @@
...
+      throw new Error('Must provide an `option` or `options` to select when calling `select`.');
+    }
+
+    const element = getElement(target) as any;
+    if (!element) {
+      throw new Error(`Element not found when calling \`select('${target}')\`.`);
+    }
+
+    if (!isSelectElement(element)) {
+      throw new Error('`select` is only usable on a HTMLSelectElement');
```
rwjblue, line 52: Can you add the `target` to this error message?

rwjblue:
Just tested in Chrome with [this demo](https://20xbc.csb.app/) (which can be [tweaked here](https://codesandbox.io/s/instrumented-element-20xbc)) and these are the events fired when clicking the select, selecting an option, and clicking out of the field:

```
focus
focusin
click
input
change
blur
focusout
```

I think we need to ensure that we emit the same events (the only ones I'm not 100% sure about is the `blur`/`focusout` as I can't recall if we do that for each helper or on the _next_ helper that runs).

Mikek2252:
Hmm looking at the fillin tests there doesn't seem to be any checks for blur or focus out so i guess it just happens when another helpers is used. Do you think it would be helpful to add another test case? perhaps with two selects that check for blur and focusout events when the second select is used.

rwjblue:
> Hmm looking at the fillin tests there doesn't seem to be any checks for blur or focus out so i guess it just happens when another helpers is used.

Sorry about the run around here, I was thinking of the conversation over in https://github.com/emberjs/ember-test-helpers/issues/692. We had decided to make the `blur`/`focusout` part of the internal `__focus__` helper function but hadn't made that change just yet.

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,210 @@
+import { module, test } from 'qunit';
+import { select, setupContext, teardownContext } from '@ember/test-helpers';
+import { buildInstrumentedElement } from '../../helpers/events';
+import { isIE11 } from '../../helpers/browser-detect';
+
+let selectSteps = ['focus', 'focusin', 'input', 'change'];
+let additionalSteps = ['input', 'change'];
+if (isIE11) {
+  selectSteps = ['focusin', 'input', 'change', 'focus'];
```
rwjblue, line 9: These are a bit off, under IE11 we should have the following (the only difference is the ordering between `focus` and `focusin`):

```suggestion
  selectSteps = ['focusin', 'focus', 'input', 'change'];
```

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,210 @@
...
+
+  test('select | 4 options - multiple: true - optionsToSelect.length : 2', async function (assert) {
+    const optionValues = ['apple', 'orange', 'pineapple', 'pear'];
+    element = buildInstrumentedElement('select');
+    element.multiple = true;
+
+    optionValues.forEach(optionValue => {
+      const optionElement = buildInstrumentedElement('option');
+      optionElement.value = optionValue;
+      element.append(optionElement);
```
rwjblue, line 110: Can you tweak this to `.appendChild` (`.append` doesn't work in IE11)?

```suggestion
      element.appendChild(optionElement);
```

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,210 @@
...
+
+  test('select | 4 options - multiple: false - optionsToSelect.length : 1', async function (assert) {
+    const optionValues = ['apple', 'orange', 'pineapple', 'pear'];
+    element = buildInstrumentedElement('select');
+    element.multiple = false;
+
+    optionValues.forEach(optionValue => {
+      const optionElement = buildInstrumentedElement('option');
+      optionElement.value = optionValue;
+      element.append(optionElement);
```
rwjblue, line 133:
```suggestion
      element.appendChild(optionElement);
```

tests/unit/dom/select-test.js
```diff
@@ -0,0 +1,210 @@
...
+    });
+
+    await setupContext(context);
+
+    await select(element, 'apple');
+
+    await select(element, 'orange');
+
+    assert.verifySteps([...selectSteps, ...additionalSteps]);
```
rwjblue, line 163: I didn't fully understand why we don't have to have `[..selectSteps, ...selectSteps]`. Specifically, what is avoiding firing `focus`/`focusin`.

Seems like this takes care of that:

https://github.com/emberjs/ember-test-helpers/blob/982320d5899af46f741bab4aea94468632b69cdc/addon-test-support/%40ember/test-helpers/dom/focus.ts#L18-L21

Figured I'd comment in case others were confused...

addon-test-support/@ember/test-helpers/dom/-is-select-element.ts
rwjblue: I think differing here is good, we don't _really_ want folks to pass us `Document`.

tests/unit/dom/select-test.js
Mikek2252: Hmm my guess is because selecting doesnt actually remove focus so if it is focused still the event will run. You can see it in your demo if you dont click out and select something new.

tests/unit/dom/select-test.js
rwjblue: I validated against FF, Chrome, Safari, and IE11. See my findings in https://github.com/emberjs/ember-test-helpers/pull/728#pullrequestreview-414433140.

tests/unit/dom/select-test.js
rwjblue: Yep, exactly. I was just surprised so I went digging to figure out why. 😃

rwjblue:
Thank you @Mikek2252!

Mikek2252:
Np thanks for reviewing, feels nice to contribute to something i use a lot

Issue #729, opened 2019-10-27
Add reseting `ember-testing-container` scroll position during render()


edeis53:
Expectation: When you use `render()` the DOM should be in a clean state and there shouldn't be any leftover DOM state of scrollbar positions.

Result: if you are scrolling elements into view during tests, and then doing subsequent `clearRender()` and `render()` the scroll position of `ember-testing-container` is not reset. 

Would you be open to a pull request for adding the following into the `render()` or `clearRender()` helpers?
`getRootElement().parentElement.scrollTo(0, 0);`

Since `clearRender()` is mostly concerned with triggering teardown behaviour, my vote would be to add a call to reset scroll in `render()` on line 107.

https://github.com/emberjs/ember-test-helpers/blob/5c4f7025cf530273b7fc240adec04a5a24113c57/addon-test-support/%40ember/test-helpers/setup-rendering-context.ts#L107

rwjblue:
Hmm. Doesn't it naturally scroll back to the top when all the content is removed?

edeis53:
That is correct. It does not scroll `<div id="ember-testing-container">` back to the top when content is removed.

I have made a reproduction with a test showing scroll position fails to reset.
https://github.com/edeis53/ember-testing-container-scroll-reproduction

Let me know if the test is clear. I am clearing the render and doing a second render within the same test. It's cleaner to illustrate that way.

The issue is also exists if the second render was a separate test. So if you have TestA which does some scrolling, and then TestB that requires an element to be in the viewport, TestB will fail if TestA runs before it because the scrollposition of the container is not guaranteed.

rwjblue:
The example test was very helpful thank you! I think a PR to add a reset mechanism seems totally good.

edeis53:
I am preparing the tests for the PR, but there is an issue with CSS used in the tests via`testing-support.css` . 

As of at least 3.7 Ember apps/addons are using `"ember-qunit": "^3.4.1",` but ember-test-helpers which is on 3.8.2  is using `"qunit": "^2.9.2",`

Ember-qunit adds some extra CSS, in particular, constrains the testing container.
```
#ember-testing-container {
...
  width: 640px;
  height: 384px;
  overflow: auto;
...
}
```

So currently, it's not possible test the scroll issue properly because there's no constraints on the testing container and without the constraints from ember-qunit it's `document.documentElement` that is scrolling, which is always going to scroll if you have lots of tests and should keep the scroll position between tests.

[...]

PR #732, opened 2019-11-06
Reset scroll state on render

edeis53:
Ensures`div#ember-testing-container` scrollTop and scrollLeft are zero on render. 

Fixes https://github.com/emberjs/ember-test-helpers/issues/729, where scrolling during a testA will persist to testB and thus make the tests brittle.

+Adds ember-qunit CSS so we can properly test the container.

Feedback/Changes welcome!

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
@@ -144,6 +144,8 @@ export function render(template: TemplateFactory): Promise<void> {
     };
     toplevelView.setOutletState(outletState);
 
+    resetScrollState();
+
     // returning settled here because the actual rendering does not happen until
     // the renderer detects it is dirty (which happens on backburner's end
     // hook), see the following implementation details:
@@ -152,6 +154,17 @@ export function render(template: TemplateFactory): Promise<void> {
     // * [backburner's custom end hook](https://github.com/emberjs/ember.js/blob/f94a4b6aef5b41b96ef2e481f35e07608df01440/packages/ember-glimmer/lib/renderer.js#L145-L159) detects that the current revision of the root is no longer the latest, and triggers a new rendering transaction
     return settled();
   });
+}
+
+/**
+  Resets the scroll position of testing container.
+
+  @private
+  @returns void
+*/
+function resetScrollState(): void {
+  const testingContainer = getRootElement().parentElement as Element;
+  testingContainer.scrollTo(0, 0);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


edeis53:
is it possible to rerun the test? yarn is throwing Segmentation fault for some reason.

rwjblue:
restarted

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -144,6 +144,8 @@ export function render(template: TemplateFactory): Promise<void> {
     };
     toplevelView.setOutletState(outletState);
 
+    resetScrollState();
```
Turbo87: I'm not sure if this is the correct place to call this. It would still have the same issue if a rendering test scrolls somewhere and then an application test follows that relies on the scroll position. Should we call this in "`afterEach()`" instead? 🤔

edeis53: I think I understand that issue. Please help me understand your thoughts on the recommendation for `afterEach()`. 

Are you suggestion that perhaps `resetScrollState()` should be a helper and then it'd be at the discretion of the user to call it in the afterEach hook whenever scroll state needs to be guaranteed? That way it will work regardless if it is an acceptance test or an integration test.

```
  setupRenderingTest(hooks);
 //OR
 //setupApplicationTest(hooks);

  hooks.afterEach(async function() {
    await resetScrollState()
  });

  test('it renders', async function(assert) {
    await render(hbs`<MyComponent />`);
});

```

edeis53: hmmm. But even that wouldn't solve the issue is there were two renders in a single component test.

```
    await render(hbs`
      await render(hbs`<MyComponent />`);
    `);

    clearRender();

    await render(hbs`
      <div>SomethingElse</div>
    `);

```

So maybe it has to be in either `render()` or `clearRender()` for integration tests as well as used in `afterEach` for acceptance tests?

rwjblue: > But even that wouldn't solve the issue is there were two renders in a single component test.

FWIW, I think this is pretty rare.

---

We can add logic to `teardownContext` and `clearRender` to reset the scroll state, that would work for both rendering and application tests as well as using `render` twice...

amk221:
Is this likely to get in?

It's just started to affect me...

Scenario: Testing dragging and dropping something. The test that is run before the drag and drop test (depending on test order with Ember Exam) may have scrolled `#ember-testing-container`, subsequently some third party drag/drop code that  uses `elementFromPoint` fails.

NullVoxPopuli:
hello! sorry for not looking at this PR sooner -- would you be willing to rebase?

PR #741, opened 2019-11-21
Respect "disabled" in "fillIn" and "typeIn"

ro0gr:
fixes: https://github.com/emberjs/ember-test-helpers/issues/680
continues: https://github.com/emberjs/ember-test-helpers/pull/681

With this PR `fillIn` and `typeIn` helpers start to fail on attempt to interact with `disabled` form controls.

There is a follow up PR, doing a similar thing for readonly(#779). Extracted it to make review easier.

---

my original intention was also to cover `click` and `doubleclick` and the rest of interaction helpers with the same logic, but seems they intentionally just skip focus for non-focusable elements:

https://github.com/emberjs/ember-test-helpers/blob/f053fe0f6ac1242cb07303a4c69600e223ed5840/addon-test-support/%40ember/test-helpers/dom/click.ts#L18-L20

Assuming changing of this behavior, might be considered a breaking change, I've extracted this part as a separate follow-up PR(#778)

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/-is-focusable.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-is-focusable.ts
@@ -1,14 +1,12 @@
 import isFormControl from './-is-form-control';
-import { isDocument } from './-target';
+import { isDocument, isContentEditable } from './-target';
 
 const FOCUSABLE_TAGS = ['A'];
 
 type FocusableElement = HTMLAnchorElement;
 
 // eslint-disable-next-line require-jsdoc
-function isFocusableElement(
-  element: HTMLElement | SVGElement | Element
-): element is FocusableElement {
+function isFocusableElement(element: Element): element is FocusableElement {
   return FOCUSABLE_TAGS.indexOf(element.tagName) > -1;
 }
 
@@ -24,11 +22,11 @@ export default function isFocusable(
     return false;
   }
 
-  if (
-    isFormControl(element) ||
-    (element as HTMLElement).isContentEditable ||
-    isFocusableElement(element)
-  ) {
+  if (isFormControl(element)) {
+    return !element.disabled;
+  }
+
+  if (isContentEditable(element) || isFocusableElement(element)) {
     return true;
   }
 
--- a/tests/unit/dom/fill-in-test.js
+++ b/tests/unit/dom/fill-in-test.js
@@ -44,6 +44,26 @@ module('DOM Helper: fillIn', function (hooks) {
     );
   });
 
+  test('filling in a disabled element', async function (assert) {
+    element = buildInstrumentedElement('input');
+    element.dataset.testDisabled = '';
+    element.setAttribute('disabled', true);
+
+    await setupContext(context);
+
+    assert.rejects(
+      fillIn(`[data-test-disabled]`, 'foo'),
+      new Error("Can not `fillIn` disabled '[data-test-disabled]'."),
+      'renders user selector'
+    );
+
+    assert.rejects(
+      fillIn(element, 'foo'),
+      new Error("Can not `fillIn` disabled '[object HTMLInputElement]'."),
+      'renders Element instance'
+    );
+  });
+
   test('rejects if selector is not found', async function (assert) {
     element = buildInstrumentedElement('div');
 
--- a/addon-test-support/@ember/test-helpers/dom/-target.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-target.ts
@@ -11,3 +11,8 @@ export function isElement(target: any): target is Element {
 export function isDocument(target: any): target is Document {
   return target.nodeType === Node.DOCUMENT_NODE;
 }
+
+// eslint-disable-next-line require-jsdoc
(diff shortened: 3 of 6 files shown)


ro0gr:
not sure what does fail on CI, it just says:

> The command "yarn test" exited with 139.

w/o any context.

ro0gr:
hm.. turns out, in `click` we don't even try to do anything with disabled inputs:

https://github.com/emberjs/ember-test-helpers/blob/7d8a9a507e4728122e67a4c6a15282436fe86341/addon-test-support/%40ember/test-helpers/dom/click.ts#L82-L86

I think it should throw. Will update.

rwjblue:
> I think it should throw. Will update.

👍

ro0gr:
@rwjblue thanks for the feedback and sorry for letting it hang, going to come back to it this week.

rwjblue:
No problem, thanks for working on it!

ro0gr:
Just added an error for `click`, and made sure all the other interactions fail if disabled element. rebased and squashed.

But a slight doubt appeared on my mind. I suspect it might be considered a breaking change, since we didn't fail on clicking disableds before.

I'm pretty sure there are some tests in the wild, like:

```js
this.doSmth = () => {
  this.called = true
}

await render(hbs`<Something 
  @disabled={{true}}
  @onClick={{this.doSmth}}
/>`

await click('.Something');

assert.equal(called, false)
``` 

which I believe is totally valid with existing click implementation, but would fail with this proposed change. From the other hand, not throwing for disableds in `{fill,type}-in` should not be allowed, even it'd break some tests (imo).

@rwjblue please let me know, what do you think? Should I split the PR into 2, in order to merge `{fill,type}-in` stuff first, and let the the rest wait for the next major, or is it fine for you to go as is?

ro0gr:
> Should I split the PR into 2, in order to merge {fill,type}-in stuff first, and let the the rest wait for the next major, or is it fine for you to go as is?

extracted everything unrelated to `fillIn` and `typeIn` into another PR https://github.com/emberjs/ember-test-helpers/pull/778

addon-test-support/@ember/test-helpers/dom/-is-focusable.ts
```diff
@@ -1,14 +1,12 @@
...
 
 const FOCUSABLE_TAGS = ['A'];
 
 type FocusableElement = HTMLAnchorElement;
 
 // eslint-disable-next-line require-jsdoc
-function isFocusableElement(
-  element: HTMLElement | SVGElement | Element
-): element is FocusableElement {
+function isFocusableElement(element: Element): element is FocusableElement {
```
Turbo87, line 9: why did you change the types here?

tests/unit/dom/fill-in-test.js
```diff
@@ -44,6 +44,17 @@ module('DOM Helper: fillIn', function(hooks) {
...
   });
 
+  test('filling in a disabled element', async function(assert) {
+    element = buildInstrumentedElement('input');
+    element.setAttribute('disabled', true);
+
+    await setupContext(context);
+    assert.rejects(
+      fillIn(`#${element.id}`, 'foo'),
+      new Error('Can not `fillIn` disabled [object HTMLInputElement]')
```
Turbo87, line 54: why do we have `[object HTMLInputElement]` in here?

addon-test-support/@ember/test-helpers/dom/-is-focusable.ts
ro0gr: I believe this was my attempt to simlify a bit. Both `HTMLElement` are `SVGElement` derrive from `Element`, so it does seem redundant to specify each.
Also, implementation wise it just checks if it's not an `A` tag!

tests/unit/dom/fill-in-test.js
ro0gr: we are passing an element instace, which is not focusable, into throw statement. Should prbably change it to use passed specifier(selector or Element) instead. Will update, thanks!

tests/unit/dom/fill-in-test.js
```diff
@@ -44,6 +44,26 @@ module('DOM Helper: fillIn', function(hooks) {
...
+
+    assert.rejects(
+      fillIn(`[data-test-disabled]`, 'foo'),
+      new Error("Can not `fillIn` disabled '[data-test-disabled]'."),
+      'renders user selector'
+    );
+
+    assert.rejects(
+      fillIn(element, 'foo'),
+      new Error("Can not `fillIn` disabled '[object HTMLInputElement]'."),
```
ro0gr, line 62: @Turbo87 just improved error messages. See this one, and 1 assert above. Please let me know if it looks good for you.

ro0gr:
CI is failing due to some conflicts in ember-data/ember types. I suspect this was caused by this PR https://github.com/emberjs/ember-test-helpers/pull/784, which doesn't have travis results reported in the PR for some reason, so we've missed the breakage. 

I think it's a good time to upgrade that types and ec-ts at once. Will look into it.

rwjblue:
I rerolled the yarn.lock in #814 (since floating deps tests worked), mind trying again?

PR #745, opened 2019-11-22
Implement test helper logging

### LINKED ISSUE #740: Action logging mode
When debugging tests it can often be useful to `console.log()` before/after e.g. an `await click()` to know when things are breaking or what action might be causing a deprecation message.

I would like to propose an "action logging mode", where each of our built-in actions uses a `console.log()` if a certain query parameter is set on the testing page. This could be coupled with a custom QUnit option, that would appear as a checkbox in the QUnit UI.

/cc @rwjblue

Turbo87:
Resolves #740 

The implementation is roughly similar to https://github.com/miragejs/ember-cli-mirage/pull/1818. I've imported the `elementToString()` implementation that [qunit-dom](https://github.com/simplabs/qunit-dom/blob/v0.9.2/lib/helpers/element-to-string.ts) is using to describe target elements if they are not passed in as selector strings.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/-logging.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-logging.ts
@@ -1,0 +1,37 @@
+/**
+ * This generates a human-readable description to a DOM element.
+ *
+ * @private
+ * @param {*} el The element that should be described
+ * @returns {string} A human-readable description
+ */
+export function elementToString(el: unknown) {
+  let desc: string;
+  if (el instanceof NodeList) {
+    if (el.length === 0) {
+      return 'empty NodeList';
+    }
+    desc = Array.prototype.slice
+      .call(el, 0, 5)
+      .map(elementToString)
+      .join(', ');
+    return el.length > 5 ? `${desc}... (+${el.length - 5} more)` : desc;
+  }
+  if (!(el instanceof HTMLElement || el instanceof SVGElement)) {
+    return String(el);
+  }
+
+  desc = el.tagName.toLowerCase();
+  if (el.id) {
+    desc += `#${el.id}`;
+  }
+  if (el.className && !(el.className instanceof SVGAnimatedString)) {
+    desc += `.${String(el.className).replace(/\s+/g, '.')}`;
+  }
+  Array.prototype.forEach.call(el.attributes, function(attr: Attr) {
+    if (attr.name !== 'class' && attr.name !== 'id') {
+      desc += `[${attr.name}${attr.value ? `="${attr.value}"]` : ']'}`;
+    }
+  });
+  return desc;
+}
--- a/addon-test-support/@ember/test-helpers/dom/-logging.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-logging.ts
@@ -1,3 +1,31 @@
+import Target from './-target';
+
+/**
+ * Logs a debug message to the console if the `testHelperLogging` query
+ * parameter is set.
+ *
+ * @private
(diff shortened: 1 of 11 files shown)


vendor/qunit-test-helper-logging-option.js
```diff
@@ -0,0 +1,8 @@
+(function() {
+  if (typeof QUnit !== 'undefined') {
```
rwjblue: IMHO, any/all QUnit logic belongs in `ember-qunit`. Let's move this over there, the result will function just the same way but keeps the separation between "worlds" clearer...

package.json
```diff
@@ -104,7 +104,10 @@
     }
   },
   "ember-addon": {
-    "configPath": "tests/dummy/config"
+    "configPath": "tests/dummy/config",
+    "after": [
+      "ember-qunit"
+    ]
```
rwjblue: FWIW, I find this a bit odd. `ember-qunit` depends on _us_ not the other way around.

addon-test-support/@ember/test-helpers/dom/-logging.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * This generates a human-readable description to a DOM element.
+ *
+ * @private
+ * @param {*} el The element that should be described
+ * @returns {string} A human-readable description
+ */
+function elementToString(el: unknown) {
```
rwjblue: Can you add some tests for this function? Doesn't need to be 100% exaustive, just something that confirms the output is reasonable...

rwjblue: Should this short circuit if `el` is a string? I think it would eventually do that (due to the condition of `!HTMLElement && !SVGElement`), but IMHO it would be easier to parse if the most common case was out of the way first...

vendor/qunit-test-helper-logging-option.js
Turbo87: good point, I will remove that part from the PR

package.json
Turbo87: true. fortunately, this is no longer needed if we move the QUnit option thing to `ember-qunit` 🎉

addon-test-support/@ember/test-helpers/dom/-logging.ts
Turbo87: yeah, I can probably just copy the tests from qunit-dom over here :)

Turbo87: done ✅

vendor/qunit-test-helper-logging-option.js
Turbo87: done ✅

addon-test-support/@ember/test-helpers/dom/-logging.ts
```diff
@@ -0,0 +1,65 @@
+import Target from './-target';
+
+/**
+ * Logs a debug message to the console if the `testHelperLogging` query
+ * parameter is set.
+ *
+ * @private
```
rwjblue: FWIW, I wonder if this should actually be private. Seems like it might be nice for folks authoring custom helpers too...

rwjblue:
restarted ci

addon-test-support/@ember/test-helpers/dom/-logging.ts
Turbo87: I'd say let's try it out and see if it works properly before we commit to making it a public API

Issue #684, opened 2019-07-16
fillIn and typeIn ignore maxlength input attribute


ghost:
Both the `fillIn` and `typeIn` helpers appear to ignore the `maxlength` input attribute (although maybe I'm missing something). I am using Ember 3.8. Here's the code I'm testing:

`{{input type="text" class="name-txt" value=name required=true maxlength="10" title=(t "form.message.format.nonBlank")}}`

This works in my app as I would expect: `Earl` is saved without issue, but `Margueritta` is not - a user cannot even enter all characters in the input.

But my below test isn't working. Both names save without issue. I also tried using the `fillIn` helper to no avail. Am I doing something dumb?

```
    assert.expect(1);

    await visit('admin/1');
    await typeIn('.name-txt', 'Margueritta');

    await click('.save-btn');

    assert.equal(this.element.querySelector('input.name-txt').value, "Earl", "did not save; over 10 characters");
```

rwjblue:
Hmm, this definitely seems not great. Would you mind making a failing test case here to make it easier for us to track down?

jaydgruber:
I ran into this same problem.
Not sure it needs to be fixed here, though it is somewhat unexpected behavior.
In my case, I ended up being content with trusting the browser to fulfill its contract, and went with:
```javascript
assert.dom(selector).hasAttribute('maxlength', maxLength);
```
Went ahead and added test cases though, since it was a quick ask. #747

hjdivad:
fixed in #747

PR #747, opened 2019-12-03
Error when `fillIn`/ `typeIn` attempt to enter a value longer than `maxlength`

jaydgruber:
### What
- adds failing test cases where fillIn and typeIn ignore the maxlength attribute of an input
- adds a util function `isMaxLengthConstrained`
- if isMaxLengthConstrained = true, will 
- fill-in helper
  - replaces element.value w/ the new value, truncated to maxlength if needed
- for type-in helper:
  - on character entries that would exceed the maxlength, fires `['keydown', 'keypress', 'keyup']` events, but no `input` event, and element.value does not change
- adds unit test cases for same

### Why
- helpers should respect constraint validation
- requested failing test case in https://github.com/emberjs/ember-test-helpers/issues/684


rwjblue:
Thanks for working on failing test cases here! If you have the time, you can probably implement a check that truncates (or whatever the "right" behavior is) here in these spots:

https://github.com/emberjs/ember-test-helpers/blob/7d8a9a507e4728122e67a4c6a15282436fe86341/addon-test-support/%40ember/test-helpers/dom/type-in.ts#L89

https://github.com/emberjs/ember-test-helpers/blob/7d8a9a507e4728122e67a4c6a15282436fe86341/addon-test-support/%40ember/test-helpers/dom/fill-in.ts#L48

jaydgruber:
I think fillIn is pretty straightforward - just truncate at maxlength, as it replaces any previous value in the control.

For typeIn I'm less sure. I think typeIn is additive, so if I did
```javascript
await typeIn('input', 'a');
await typeIn('input', 'b');
```
I would end up w/ 'ab'. So I guess typeIn should stop modifying element.value when the maxlength of the combined (previous values + new values).length === maxlength.
But I think it should still fire all the key events? Like if a user tried to type past the maxlength of a field, they still hit the keys, just no more values get added.

rwjblue:
> But I think it should still fire all the key events? Like if a user tried to type past the maxlength of a field, they still hit the keys, just no more values get added.

Yep, that is what I think also but we should confirm that is the actual browser.

jaydgruber:
I went down a bit of a rabbit hole reading docs around https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation
[max, maxlength, min, minlength, pattern, size, step] all have validation that prevents form submission, but looks like only maxlength prevents users from entering values.
I tried to figure out a way to check the Constraint Validation API's https://webplatform.github.io/docs/dom/ValidityState/tooLong/, but was getting false negatives. Not sure what sort of event (dirty flag maybe?) needs to be fired to get checkValidity to report correctly when the input was not user-entered.
Anyways, abandoned that approach in favor of simpler one-off

addon-test-support/@ember/test-helpers/dom/fill-in.ts
```diff
@@ -43,13 +43,17 @@ export default function fillIn(target: Target, text: string): Promise<void> {
...
     __focus__(element);
-
     if (isControl) {
-      element.value = text;
+      // ref: https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
+      const maxlengthApplicableTypes = ['text', 'search', 'url', 'tel', 'email', 'password'];
+      const { type, tagName } = element;
+      const maxlength = element.getAttribute('maxlength');
+      const shouldValidateMaxlength =
+        maxlength && (maxlengthApplicableTypes.includes(type) || tagName === 'TEXTAREA');
```
rwjblue: We still support IE11, so unfortunately we’ll need to use `.indexOf(type) > -1`

rwjblue: Also, can you swap this around _slightly_ to something like:

```js
const shouldValidateMaxLength = element.hasAttribute(‘maxlength’) && ((tagName === ‘INPUT’ && maxlengthApplicableTypes.indexOf(type) > -1) || tagName === ‘TEXTAREA’);

element.value = shouldValidateMaxlength ? text.slice(0, maxlength) : text;
```

The main things I’m concerned with here are:

* base shouldValidateMaxlength on element.hasAttribute
* only validate `type` for `input`s

jaydgruber:
Still working on typeIn

addon-test-support/@ember/test-helpers/dom/fill-in.ts
jaydgruber: Makes sense. I guess this should probably move into a utils/ fn, as typeIn will need need same logic

addon-test-support/@ember/test-helpers/dom/-is-maxlength-constrained.ts
```diff
@@ -0,0 +1,15 @@
+/**
+  @private
+  @param {Element} element the element to check
+  @returns {boolean} `true` when the element should constrain input by the maxlength attribute, `false` otherwise
+*/
+export default function isMaxLengthConstrained(element: Element) {
```
jaydgruber: New to TS, had a couple types questions here.
I noticed `-is-form-control` had an export type FormControl. Does a type guard make sense here?
```typescript
import { FormControl } from './-isFormControl';

export default function isMaxLengthConstrained(element: Element) element is FormControl {
```

Also I noticed formControl had the parameter annotation `isFormControl(element: Element | Document)` and I wasn't sure what the Document bit was about

btecu:
@jaydgruber what's the status on this?

jaydgruber:
oh I guess I could have pinged someone for re-review. this one should be ready to go, pending the one above question about do I need to do more to make it 'TypeScript-y'

rwjblue:
@ro0gr - Would you mind reviewing? This touches a bunch of the same general areas as #741 / #779  / #778 so it will need to land after those.

addon-test-support/@ember/test-helpers/dom/-is-maxlength-constrained.ts
ro0gr: I don't think `-is-form-control` is applicable here, cause in addition to inputs and textareas it does also account for buttons and selects, which, I think, is un-desirable. However, you could add you own type guard here to specify `element` type on the invocation side, like:
```ts
export function isMaxLengthConstrained(element: Element): element is HTMLInputElement | HTMLTextAreaElement {
// ...
```

> isFormControl(element: Element | Document)

ya, I believe it should not accept `Document`. I suspect it's encouraged by `getElement` which is currently able to return `Document` https://github.com/emberjs/ember-test-helpers/blob/637fdbc8a45cfed013e47701c786d19c89cf06ad/addon-test-support/%40ember/test-helpers/dom/-get-element.ts#L16

and we use its result to pass it down to the other interaction helpers. I think `getElement` return types may be worth to reconsider. But this is probably unrelated to this PR.

ro0gr:
Seems like `input` event stops being triggered once we reach the `maxlength` constraint. I've checked it in https://codepen.io/ro0gr/pen/YzyGvrZ?editors=1010

As far as I can see, in this PR, we try to trim the input and keep emitting events. If my codepen test is correct, I think we should not emit any events. Maybe we should even throw, to make un-intended `fillIn` more clear to the test helper user. Some context with related discussion in the thread https://github.com/emberjs/ember-test-helpers/pull/778#discussion_r410071274

jaydgruber:
Yeah, I think throw similarly makes sense here

addon-test-support/@ember/test-helpers/dom/-is-maxlength-constrained.ts
```diff
@@ -0,0 +1,15 @@
...
+  // ref: https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
+  const constrainedInputTypes = ['text', 'search', 'url', 'tel', 'email', 'password'];
+  return (
+    element.hasAttribute('maxlength') &&
```
ro0gr, line 10: I'd prefer properties over attributes, like `element.maxLength`. Not sure if it makes the difference here though.

ro0gr:
@jaydgruber I just wanna check if you are going to keep working on this? in case you are busy with something else, I can take it over.

rwjblue:
OK #741, #779, and #778 have all landed. This is ready for a rebase, and the tweaks that @ro0gr mentioned above (thanks @ro0gr for reviewing!).

jaydgruber:
@ro0gr I won't be able to get to it until probably Friday. I've got no objections if you want to pick it up. Thanks

ro0gr:
@jaydgruber Great! sounds like you are in the game :) Take your time.

addon-test-support/@ember/test-helpers/dom/-is-maxlength-constrained.ts
jaydgruber: I left this as is. Using properties:
`!!Number(element.maxLength)` was giving me lint error: `error TS2339: Property 'maxLength' does not exist on type 'Element'.`, and I got in over my head on casting trying to fix it

jaydgruber:
@ro0gr this should be ready for another look

addon-test-support/@ember/test-helpers/dom/type-in.ts
```diff
@@ -94,7 +95,20 @@ function keyEntry(element: FormControl, character: string): () => void {
...
       .then(() => {
-        element.value = element.value + character;
+        const newValue = element.value + character;
+        const maxlength = element.getAttribute('maxlength');
+        const shouldTruncate =
+          isMaxLengthConstrained(element) &&
+          maxlength &&
+          newValue &&
+          newValue.length > Number(maxlength);
+        if (shouldTruncate) {
```
ro0gr, line 105: The variable name looks obsolete.

When looking at it and at the condition above, I'm wondering if we can hide this logic into a standalone util, like:

```ts
/**
 * @throws if `value` exceeds the `maxlength`(if specified)
 */
function guardForMaxlength(element: FormControl, value: string): void;
```

I believe it could reduce duplication a bit.

tests/unit/dom/type-in-test.js
```diff
@@ -208,4 +208,38 @@ module('DOM Helper: typeIn', function (hooks) {
...
+    const maxLengthString = 'f';
+    const tooLongString = maxLengthString.concat('o');
+    element.setAttribute('maxlength', maxLengthString.length);
+
+    await setupContext(context);
+
+    await assert.rejects(
+      typeIn(element, tooLongString).finally(() => {
+        // should throw before the second input event
+        assert.verifySteps(expectedEvents.slice(0, 8));
```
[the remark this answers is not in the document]
rwjblue: Ya, it is possible, we don't have IE11 under CI test at the moment. We should look into hooking that up... 🤔

addon-test-support/@ember/test-helpers/dom/type-in.ts
jaydgruber: yeah that made sense. `isMaxLengthConstrained` then became just an implementation detail of `guardForMaxlength`.

tests/unit/dom/type-in-test.js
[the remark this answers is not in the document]
jaydgruber: I believe I corrected the assertion here, but I don't have an IE launcher for testem locally, so not sure the best way to confirm

rwjblue:
Awesome, thank you!

PR #749, opened 2019-12-09
Add more tests for resetOnerror

Exelord:
I found a bug, where resetOnerror is not resetting to the original function. Instead, it is always assigning undefined.

Here is the test. The issue is still out there :/

### DIFF
--- a/tests/unit/setup-ember-onerror-test.js
+++ b/tests/unit/setup-ember-onerror-test.js
@@ -46,8 +46,8 @@ module('setupOnerror', function(hooks) {
     test('Ember.onerror is reset correctly to original function', async function(assert) {
       assert.expect(3);
       
-      let originalOnerror = (err) => {
-        console.log('original on error')
+      let originalOnerror = err => {
+        console.log('original on error');
         throw err;
       }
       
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm, I was just looking into this. I don't think this is an actual error in `resetOnerror` (or we have a larger issue in play). `resetOnerror` is meant to bring the `Ember.onerror` state back to its original value **not** the most recent prior value.

tests/unit/setup-ember-onerror-test.js
```diff
@@ -5,7 +5,7 @@ import { setupOnerror, resetOnerror } from '@ember/test-helpers';
 
 module('setupOnerror', function(hooks) {
   hooks.afterEach(function() {
-    resetOnerror();
+    Ember.onerror = undefined;
```
scalvert, line 8: Calling `resetOnerror` is intended to reset the error state to the very first state that was stashed when the module was required. This shouldn't be reset to `undefined` explicitly, but rather to the original error state as defined [here](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/@ember/test-helpers/setup-onerror.ts#L3). Other than this, added the test case seems fine.

scalvert:
I just noticed you said the same thing above, @rwjblue. 👍

PR #750, opened 2019-12-12
Add examples to `waitUntil` and `waitFor`

jenweber:
I personally need this reminder. I keep forgetting to await and return things.

### DIFF
--- a/API.md
+++ b/API.md
@@ -503,7 +503,7 @@ interim DOM states (e.g. loading states, pending promises, etc).
 
 #### Examples
 
-Waiting until a selector is rendered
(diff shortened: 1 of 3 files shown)


scalvert:
👍 this is great. I agree we needed examples.

jenweber:
Hi @Turbo87, this is ready for another review. Apologies for taking so long to update this. I added the examples into the source code and regenerated `API.md` using `yarn docs`. Some of the diff in the md is from preexisting changes in the source that look like they didn't get built in yet.

Let me know if anything else is needed here!

Issue #738, opened 2019-11-20
Test helper to simulate tabbing


mydea:
I don't even know if that would really be doable, but it would be awesome to have a test helper to simulate tabbing through a document.

I recently wrote some code for focus trapping, and couldn't really find a way to test if everything works as expected. I tried to do e.g. `triggerKeyEvent(myElement, 'keydown', 'Tab')` but that did not actually tab to the next element. Calling `focus()` does focus it, but is not a real substitute for this type of test.

Having something like `tab(myElement)` and `tab(myElement, { backwards: true })` (or similar) would be very helpful for some a11y-related tests!

dbollinger:
Finding myself looking for a similar solution today, specifically testing that a focus trap has been properly applied to my component.

mydea:
Yeah, I did something similar to this in the end, which is OK I guess. But things like testing if the default behavior still works (e.g. if I tab on a non-first/last item, it moves to the next item), is not possible like this, I guess!

rwjblue:
I'm _generally_ ok with something in this space, but I don't know what the best way to implement it would be...

mydea:
Yeah, I sadly also don't really have an idea there, otherwise I'd have tried to do a PR myself - but maybe somebody out there _has_ an idea ;)

PR #771, opened 2020-01-29
Tab

mogstad:
Took a stab at making a tab-button emulator. (#738)

I wasn’t completely sure the best way to test it, had to make a few modifications to the existing test helper to test the events it generates. I realized a bit late that it might be good enough to test if the active element was set currently, and not ensuring the user-agent fires the right events. Should I refactor to just check the active element?

Few notes:

I exported `buildKeyboardEvent` as I had to knew if the event was prevented, it seems like the easiest way to accomplish that. IE11 doesn't support "Event.defaultPrevented".

I changed `instrumentElement` to accept key paths for its `logOptionsProperties` argument, to get information about the target of the event.

I wasn’t able to get the tests running in IE11, and had to port a native for-of loop to a `forEach`. But I assume I’m just doing something incorrect, any tricks I need to know. I added it as a dedicated commit, and can easily drop it.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/tab.ts
+++ b/addon-test-support/@ember/test-helpers/dom/tab.ts
@@ -1,0 +1,184 @@
+import getRootElement from './get-root-element';
+import settled from '../settled';
+import { buildKeyboardEvent } from './fire-event';
+import { isDocument } from './-target';
+import { __blur__ } from './blur';
+import { nextTickPromise } from '../-utils';
+import { createTreeWalker } from './-create-tree-walker';
+
+const FORM_TAGS = ['INPUT', 'BUTTON', 'SELECT', 'TEXTAREA', 'FIELDSET'];
+const SUPPORTS_INHERT = Element.prototype.hasOwnProperty('inert');
+const FALLBACK_ELEMENTS = ['CANVAS', 'VIDEO', 'PICTURE'];
+
+function isVisible(element: HTMLElement): boolean {
+  let styles = window.getComputedStyle(element);
+  return styles.display !== 'none' && styles.visibility !== 'hidden';
+}
+
+function isDisabled(element: HTMLElement): boolean {
+  if (FORM_TAGS.indexOf(element.tagName) !== -1) {
+    return (element as HTMLInputElement).disabled;
+  }
+  return false;
+}
+
+interface InhertHTMLElement extends HTMLElement {
+  inhert: boolean;
+}
+
+// Compiles a list of nodes that can be focued. Walkes the tree, discardes hidden
+// elements and a few edge cases. To calculate the right
+function compileFocusAreas(root: Element = document.body) {
+  let activeElment = getActiveElement(root.ownerDocument!);
+  let treeWalker = createTreeWalker(
+    root,
+    NodeFilter.SHOW_ELEMENT,
+    {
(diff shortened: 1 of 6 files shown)


Turbo87:
@MelSumner do you have time to take a look at this? I assume you know best about tabbing orders and things like that 😅

mogstad:
After reading an old conversation in Discord about this, I realized that the first expectation we have about manually invoked events are that the user-agent handles this for us. E.g. manually firing a key event with a letter, will actually insert that letter into the input.

Maybe we should exploit this exception and make a `press(key)` API instead. If you invoke it with a `keyCode` we fire the necessary events and emulate the user-agent as good as we can:

- If the target is an input we basically fire `fillIn` with that letter. 
- If you fire the tab key, we emulate the tab key. 
- invoking it with "space" or "enter" on a button we fire a click event on the button
- invoking it with "enter" on a link, we do the appropriate for it.

This obviously opens a can of worms, the above is just a few of the default behaviours that might happen, there are many keyboard events for input fields, to name a few:

- arrow navigation in inputs
- arrow navigation in the browser frame
- contenteditable commands (bold, italic, underline)
- browser specific command (e.g. "⌘+L" focuses the address-bar, basically blurs the window)

I don’t suggest implementing these, but I suspect if we start acting like a UA, people might be confused that only some of default behaviour works.

addon-test-support/@ember/test-helpers/dom/tab.ts
```diff
@@ -0,0 +1,248 @@
+import getRootElement from './get-root-element';
+import settled from '../settled';
+import fireEvent, { buildKeyboardEvent } from './fire-event';
+import { isDocument } from './-target';
+import { __blur__ } from './blur';
+import { nextTickPromise } from '../-utils';
+import { createTreeWalker } from './-create-tree-walker';
+import { __focus__ } from './focus';
+
+const FORM_TAGS = ['INPUT', 'BUTTON', 'SELECT', 'TEXTAREA', 'FIELDSET'];
```
MelSumner:
```suggestion
const FORM_TAGS = ['INPUT', 'BUTTON', 'SELECT', 'TEXTAREA'];
```
A fieldset isn't one of the interactive elements for form tags (assuming that's what you're going for here).

MelSumner:
@mogstad @Turbo87 considering `TAB` key interactions: 
- ignore if `tabindex=-1`
- cycle through tabindex at 0 (either explicit or implicit)
- cycle through tabindex at 1 (2,3, etc) OR give an error message that says "tabindex greater than 0 should not be defined". Current standard practice is to not support a tabindex greater than 0, so I prefer that.

addon-test-support/@ember/test-helpers/dom/tab.ts
mogstad: That part is actually handled by UA, with the `tabIndex` property of the element. The `tabIndex` property returns default values, or the `tabindex` attribute, parsed as a number. It doesn’t take into account anything else though, so we’ll need to manually handle that. One of these cases are the disabled attribute, this variable is the list of tags that can be disabled. When walking the DOM, it rejects any disabled elements and its children. 

It means it won’t tab to this input element in this case:

```html
<fieldset disabled>
  <input />
</fieldset>
```

The name of the variable doesn’t really describe that, will come up with a better name.

PR #664, opened 2019-06-10
WIP - Allow render to use an engine.

rwjblue:
Working with @villander on https://github.com/ember-engines/ember-engines/pull/653

Needs:

- [ ] Tests
- [ ] Review
- [ ] 💸

PR #795, opened 2020-04-16
Add support to Engines

villander:


ef4:
What is the use case for this? 

My understanding is that the easiest way to test an engine is within its own dummy app, like any other addon. 

Trying to test the inside of an engine from the host app fights against what engines are supposed to be -- they're isolated so they can be developed independently.

villander:
@ef4 the use case turns **Test API** on Engines run the closest possible to a normal Ember app. It already done on Engine addon - https://github.com/ember-engines/ember-engines/pull/653 

```js
  test('Ember Engines', async function(assert) {
    // Register engine
    let BlogEngine = Engine.extend({
      router: null,

      init() {
        this._super(...arguments);

        this.owner.register(
          'component:x-foo',
          Component.extend({
            actions: {
              clicked() {
                assert.ok(true, 'click was fired');
              },
            },
          })
        );
        this.owner.register(
          'template:components/x-foo',
          hbs`<button {{action 'clicked'}}>Click me!</button>`
        );
      },
    });

    this.owner.register('engine:blog', BlogEngine);

    // Register engine route map
    let BlogMap = function() {
      this.route('post');
    };
    this.owner.register('route-map:blog', BlogMap);
    debugger;
    this.owner.router.map(function() { // broken here
      this.mount('blog');
    });

    await visit('/blog');

    await render(hbs`{{x-foo}}`);

    assert.equal(this.element.textContent, 'Click me!', 'precond - component was rendered');
    await click(this.element.querySelector('button'));
})
```

I'm trying to test like we did on [ember.js repo](https://github.com/emberjs/ember.js/blob/04e324b7930247024915bcb32fbc75e91e3501ae/packages/ember/tests/routing/decoupled_basic_test.js#L2163), but `this.owner.router.map` return undefined

rwjblue:
Why does this not mention https://github.com/emberjs/ember-test-helpers/pull/664 (or target that branch)?

villander:
Mention added @rwjblue

villander:
@ef4 @rwjblue how can I write tests like [this one](https://github.com/villander/ember-engines/blob/master/tests/integration/setup-engine-test.js) in this current architecture of tests on `ember-test-helpers`, this API is a bunch different.

ef4:
I don't really understand what you're trying to do.

villander:
@ef4 in a simple way I need to add test coverage here - on these changes - https://github.com/emberjs/ember-test-helpers/pull/795/commits/0c76b771041885c9ce67752eb44b53e9cee70ae6

ef4:
You don't need to make a new engine dynamically for the test. You can probably add an in-repo engine to the dummy app.

villander:
@ef4 That's what I did, take a look at my changes, please. 

I'm using `engineResolverFor`. but I'm not sure if it works with `setupRenderingContext`

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -105,19 +105,21 @@ export function render(template: TemplateFactory): Promise<void> {
       throw new Error('Cannot call `render` without having first called `setupRenderingContext`.');
     }
 
-    let { owner } = context;
+    let { owner, engine } = context;
```
ef4, line 108: Your new feature expects to see an `engine` argument to the test context...

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,49 @@
...
+
+import hbs from 'htmlbars-inline-precompile';
+
+const modulePrefix = 'eager-engine';
+const resolver = engineResolverFor(modulePrefix);
+
+module('setupRenderingContext "ember-engines"', function(hooks) {
+  hooks.beforeEach(async function() {
+    await setupContext(this);
+    await setupRenderingContext(this, { resolver });
```
ef4, line 19: ...but your test doesn't set `engine`?

villander: How can I set `engine` here? I have no idea

villander: Done @ef4

villander:
@ef4 @rwjblue tests added, but the `yarn install` is broken on CI and sometimes local.

rwjblue:
We've landed a bunch of fixes to make CI more consistent (dropping Node < 10, reroll yarn.lock, etc), probably needs a rebase to kick things off again.

villander:
I did rebase, now I have this error: `Error: Could not find module `eager-engine/resolver` imported from `eager-engine/engine``

I don't know what's going on, I need to figure out

villander:
@rwjblue my tests to engines are passing when I run it separately, but when I run with all, it breaks every test inside `/unit/setup-application-context-test.js`. 

I couldn't find the reason for this =/

villander:
@ef4 @rwjblue thanks a lot for the help on it. 

it's ready to merge ; ) can you review it, please?

.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   env: {
     browser: true,
   },
+  ignorePatterns: ['node_modules/'],
```
rwjblue, line 13: What is the difference between `ignorePatterns` and `.eslintignore`?

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,50 @@
...
+} from '@ember/test-helpers';
+import engineResolverFor from 'ember-engines/test-support/engine-resolver-for';
+import hbs from 'htmlbars-inline-precompile';
+
+const modulePrefix = 'eager-engine';
+const resolver = engineResolverFor(modulePrefix);
+
+module('setupRenderingContext for "ember-engines"', function (hooks) {
+  hooks.beforeEach(async function () {
+    await setupContext(this, { resolver });
```
rwjblue, line 17: engines should no longer need a custom resolver, please remove

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,50 @@
...
+    // this.owner now is an engine.
+    this.engine = this.owner;
```
rwjblue, line 19: Hmm, can you actually make it a different owner (just aliasing them doesn't allow us to confirm that things actually function)? You can use the `build-owner.ts` file to make a different owner instance.

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,50 @@
...
+    await setupRenderingContext(this);
+  });
+
+  hooks.afterEach(async function () {
+    await teardownRenderingContext(this);
+    this.engine = undefined;
```
rwjblue, line 25: No need to reset here, but when you update to make `this.engine` a different value from `this.owner` (see other comment) you will need to destroy `this.engine`.

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,50 @@
...
+    );
+
+    this.set('colorValue', 'blue');
+
+    assert.equal(
+      this.element.querySelector('div').getAttribute('style'),
+      'color: blue',
+      'updates to blue'
+    );
+  });
```
rwjblue, line 49: This test doesn't _really_ work (it is just using `this.owner`), we should have a test that registers a component into `this.engine` (when `this.engine !== this.owner`) that is not present on `this.owner` and render it.

package.json
```diff
@@ -72,6 +72,8 @@
     "eslint-plugin-disable-features": "^0.1.3",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.1.3",
+    "ember-engines": "^0.8.5",
+    "eager-engine": "link:./tests/dummy/lib/eager-engine",
```
rwjblue, line 76: I don't think we need to add `ember-engines` as a dependency here.

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,50 @@
+import { module, test } from 'qunit';
+import {
+  setupContext,
+  setupRenderingContext,
+  teardownContext,
+  teardownRenderingContext,
+  render,
+} from '@ember/test-helpers';
+import engineResolverFor from 'ember-engines/test-support/engine-resolver-for';
```
rwjblue, line 9: ✂️

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,50 @@
...
+import hbs from 'htmlbars-inline-precompile';
```
rwjblue, line 10:
```suggestion
import { hbs } from 'ember-cli-htmlbars';
```

tests/integration/ember-engines-test.js
```diff
@@ -0,0 +1,50 @@
...
+
+const modulePrefix = 'eager-engine';
+const resolver = engineResolverFor(modulePrefix);
```
rwjblue, line 13: 🔪

package.json
villander: why not @rwjblue? how can I create a engine-repo without ember-engines addon?

tests/integration/ember-engines-test.js
villander: In this case `this.owner` is an engine using `engineResolverFor`(that's what this helper does). And this component just exist in the `eager-engine`. then it's working.

tests/integration/ember-engines-test.js
villander: @rwjblue `buildOwner` require an application and resolver as params, how can I create a `engine` instance with this file? 

are you proposing to create a `buildEngine.ts` or something like this?

package.json
villander: I'm following the @ef4 tip to create an engine, instead make engine dynamically (see the comment https://github.com/emberjs/ember-test-helpers/pull/795#issuecomment-615465238)

tests/integration/ember-engines-test.js
villander: I think our `build-owner.ts` doesn't have the same API to support build engines as done in ember.js repo - https://github.com/emberjs/ember.js/blob/0554471ba64d9ade1dd1c111e53803fa5f7a97b6/packages/internal-test-helpers/lib/build-owner.js

tests/integration/ember-engines-test.js
villander: makes sense!

do you want something like this

https://github.com/emberjs/ember.js/blob/f73d8440d19cf86a10c61ddb89d45881acfcf974/packages/%40ember/-internals/routing/tests/ext/controller_test.js#L15-L23

villander:
@rwjblue any chance to have a look here to make a progress?

PR #823, opened 2020-04-21
cleanup: restrict Target to string or Element only

ro0gr:
for the majority of interaction helpers, having `Document` or `Window` as a target doesn't make sense.

The only helper which, looks like, may need `Document` or `Window` as a target, is `fireEvent` when handling mouse events https://github.com/emberjs/ember-test-helpers/blob/fa114a0287510ba9f87237d75644a1cc9fbc1c32/addon-test-support/%40ember/test-helpers/dom/fire-event.ts#L93-L96

if I'm not mistaken it can be only reached via `triggerEvent`. If such, then `Document` as a target should be only allowed in `triggerEvent` https://github.com/emberjs/ember-test-helpers/blob/fa114a0287510ba9f87237d75644a1cc9fbc1c32/addon-test-support/%40ember/test-helpers/dom/trigger-event.ts#L52-L53, which would allow to simplify code for the rest of helpers.

todo: 
 - [ ] figure out if we need `Document | Window` as a `triggerEvent` target
 - [ ] if "yes", implementation should be updated accordingly, and missing test case should be added

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/-target.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-target.ts
@@ -1,4 +1,4 @@
-type Target = string | Element | Document | Window;
+type Target = string | Element;
 
 export default Target;
 
--- a/addon-test-support/@ember/test-helpers/dom/scroll-to.ts
+++ b/addon-test-support/@ember/test-helpers/dom/scroll-to.ts
@@ -2,7 +2,6 @@ import getElement from './-get-element';
 import fireEvent from './fire-event';
 import settled from '../settled';
 import { nextTickPromise } from '../-utils';
-import { isElement } from './-target';
 
 /**
   Scrolls DOM element or selector to the given coordinates.
@@ -37,12 +36,6 @@ export default function scrollTo(
     let element = getElement(target);
     if (!element) {
       throw new Error(`Element not found when calling \`scrollTo('${target}')\`.`);
-    }
-
-    if (!isElement(element)) {
-      throw new Error(
-        `"target" must be an element, but was a ${element.nodeType} when calling \`scrollTo('${target}')\`.`
-      );
     }
 
     element.scrollTop = y;
--- a/addon-test-support/@ember/test-helpers/dom/-get-element.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-get-element.ts
@@ -1,11 +1,9 @@
 import getRootElement from './get-root-element';
-import Target, { isDocument, isElement } from './-target';
+import Target, { isElement } from './-target';
 
 function getElement(target: string): Element | null;
 function getElement(target: Element): Element;
-function getElement(target: Document): Document;
-function getElement(target: Window): Document;
-function getElement(target: Target): Element | Document | null;
+function getElement(target: Target): Element | null;
 /**
   Used internally by the DOM interaction helpers to find one element.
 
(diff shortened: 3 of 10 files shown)


ro0gr:
this cleanup was inspired by the question https://github.com/emberjs/ember-test-helpers/pull/747#discussion_r358958284

rwjblue:
> The only helper which, looks like, may need Document or Window as a target, is fireEvent when handling mouse events

Doesn't `scrollTo` also need that?

rwjblue:
Can you help me understand what we are solving?

ro0gr:
> Doesn't scrollTo also need that?

currently it does only support an `Element` as a scrollable target. In case of `Document|Window` it throws.

https://github.com/emberjs/ember-test-helpers/blob/fa114a0287510ba9f87237d75644a1cc9fbc1c32/addon-test-support/%40ember/test-helpers/dom/scroll-to.ts#L42-L46

I think it can support `Document|Window`, but it does not do it yet. So this cleanup should not affect `scrollTo`.

> Can you help me understand what we are solving?

This is mostly a cleanup. For now we leak `Element|null|Document`|`Window` as a `getElement` return type to all all helpers, while for the most of the helpers `Element|null`.
In my opinion the current state of things leads to unnecessary complexity. 

So in few places you can see guards against the `Document` to avoid dealing with it, cause it's allowed to be a [`Target` ](https://github.com/emberjs/ember-test-helpers/blob/fa114a0287510ba9f87237d75644a1cc9fbc1c32/addon-test-support/%40ember/test-helpers/dom/-target.ts#L1)for any helper.  

Like one I've shared for the `scroll-to`, or `focus`
https://github.com/emberjs/ember-test-helpers/blob/fa114a0287510ba9f87237d75644a1cc9fbc1c32/addon-test-support/%40ember/test-helpers/dom/-is-focusable.ts#L21-L23, or `is-form-control` https://github.com/emberjs/ember-test-helpers/blob/fa114a0287510ba9f87237d75644a1cc9fbc1c32/addon-test-support/%40ember/test-helpers/dom/-is-form-control.ts#L16-L20

Also most of the helpers accept the first argument which has a `Target` type.
So for instance `fillIn` https://github.com/emberjs/ember-test-helpers/blob/fa114a0287510ba9f87237d75644a1cc9fbc1c32/addon-test-support/%40ember/test-helpers/dom/fill-in.ts#L27
does allow a `document` to be passed in. The same behavior applies to other helpers.

 I find all of this a bit misleading, and that can lead to extra bits of code which does not seem to produce any value. I believe it'd be better to explicitly declare if the Document is supported in helpers where it's only needed, like:

```ts
export default function triggerEvent(
  target: Target | Document | Window,
  eventType: string,
  options?: object
```

and treat them accordingly in that helpers only. That should increase clarity of the flow.

ro0gr:
not sure if my explanation is clear enough, hope it makes sense for you.

addon-test-support/@ember/test-helpers/dom/scroll-to.ts
```diff
@@ -39,12 +38,6 @@ export default function scrollTo(
       throw new Error(`Element not found when calling \`scrollTo('${target}')\`.`);
     }
 
-    if (!isElement(element)) {
-      throw new Error(
```
ro0gr, line 43: this is removed, cause `getElement(`, which stopped returning `Document` would fail if the element is not found. So we won't ever reach this  line.

addon-test-support/@ember/test-helpers/dom/-target.ts
```diff
@@ -1,4 +1,4 @@
-type Target = string | Element | Document | Window;
```
ro0gr, line 1: this is used as type for the first argument in most of the interaction helpers.
  
So, if we were emitting public types, this would be a breaking change. However, under the hood we often just ignore `Document`, and try to do nothing or fail.

ro0gr:
Alright, I think I try to solve 2 issues here:
 - `getElement`'s return type, which was my original intention
 - make `Target` more constrained, which I concluded makes sense while worked on `getElement`

I'll try to solve these 2 problems with 2 differents PRs.

PR #824, opened 2020-04-21
Migrate CI to GitHub Actions.

rwjblue:


.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: volta-cli@action@v1
+        with:
+          node-version: 12.x
+      - name: install dependencies
+        run: yarn install --frozen-lockfile
+      - name: test
+        env:
+          EMBER_TRY_SCENARIO: ${{ matrix.ember-try-scenario }}
```
Turbo87, line 67: why the env var?

rwjblue: No clue why I did it that way, just removed

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,66 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: volta-cli@action@v1
+        with:
+          node-version: 12.x
+      - name: install dependencies
+        run: yarn install --frozen-lockfile
+      - name: test
+        run: node_modules/.bin/ember try:one ${{ matrix.ember-try-scenario }} --skip-cleanup
```
Turbo87, line 66: if we're using volta here, do we still need to reach into `.bin` or can we call `ember` directly?

rwjblue:
> If we're using volta here, do we still need to reach into .bin or can we call ember directly?

Ya, thought about that, but we'd have to `volta install ember-cli` and in this case we don't actually need / want the global install to run (since we know **for sure** that we'd always want to defer).

.github/workflows/ci-build.yml
rwjblue: > If we're using volta here, do we still need to reach into .bin or can we call ember directly?

Ya, thought about that, but we'd have to volta install ember-cli and in this case we don't actually need / want the global install to run (since we know for sure that we'd always want to defer).

Turbo87: hmm, okay, then I misunderstood volta apparently. we could still use `yarn ember` though.

PR #849, opened 2020-05-05
Migrate from `requirejs.entries` to `has`.

rwjblue:


types/require.d.ts
```diff
@@ -1 +1,2 @@
 export default function <T = any>(moduleName: string): T;
+export function has<T = any>(moduleName: string): T;
```
ro0gr, line 2: Seems this should return boolean

rwjblue: Doh! you are correct

rwjblue: https://github.com/emberjs/ember-test-helpers/pull/851

Issue #871, opened 2020-05-26
triggerKeyEvent is not IE11 compatible due to Array.find


raido:
`Array.find` usage in `triggerKeyEvent` method breaks it for IE11.

Reference: https://github.com/emberjs/ember-test-helpers/blob/a6af003242f04653fac31df8f93a55bfb57ac830/addon-test-support/%40ember/test-helpers/dom/trigger-key-event.ts#L109

Current workaround is to include `Array.find` polyfill for tests only if application itself does not depend on it.

`app.import("vendor/array-find-polyfill.js", { type: "test" });`

rwjblue:
Eeck! Sorry about that @raido!

We should definitely fix, would love a PR if someone has the time.

raido:
This should be closed, PR is merged.

PR #872, opened 2020-05-27
Issue 871 - Make triggerKeyEvent compatible with the IE11 

lyubarskiy:
https://github.com/emberjs/ember-test-helpers/issues/871

In order to use triggerKeyEvent in IE11 `Array.find` was replaced by `Array.filter`

_`filter` is a bit slower than `find` here, if it's an issue I'll come up with another solution_


rwjblue:
CI on 3.8 is failing for unrelated reasons.

rwjblue:
Re-running CI to pick up fixes for ember-cli-babel@7.20.

PR #870, opened 2020-05-21
Issue 794 - add deprecation messages to this.render and this.clearRender

Mikek2252:
#794 

Added deprecation warnings to both functions when added to this.

@rwjblue wasn't sure what number to put the until under so have a place holder at the moment.

Issue #794, opened 2020-04-14
Deprecate `this.render` and `this.clearRender`


rwjblue:
These APIs were a complete accident: during a very early iteration of the new system I added these APIs, then subsequently replaced them with the utility style import APIs for the final design.

I just realized that these had not been deprecated yet 😩:facepalm: 

https://github.com/emberjs/ember-test-helpers/blob/a5f59e41bce78a66dbcd0f4dd7e82ff874c75594/addon-test-support/%40ember/test-helpers/setup-rendering-context.ts#L210-L223

Mikek2252:
@rwjblue I can help with this, Could i get some info on what to put in the messages?

rwjblue:
Thanks! Ya, the message basically would need to say that you should refactor from `this.render` to `render` imported from `@ember/test-helpers`.

Mikek2252:
@rwjblue I guess this can be closed now?

rwjblue:
Yep, thank you!

PR #876, opened 2020-06-04
Add deprecation for `this.render` and `this.clearRender`

Mikek2252:
#794 Fixing issue to add deprecation warning to this.clearRender and this.render.
Clone of # #870 but with cleaned up hit history. Wasn't sure how best to fix my forked master.


rwjblue:
Thank you!

PR #878, opened 2020-06-04
Adding hooks to allow adding work start/end of test helpers

scalvert:


addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Registers a hook function to be run during the invocation of a test helper.
+ *
+ * @public
+ * @param {string} helperName The name of the test helper in which to run the hook.
+ * @param {Function} hook The hook function to run when the test helper is invoked.
+ * @returns {HookUnregister} An object containing an unregister function that will unregister
+ *                           the specific hook registered to the helper.
+ */
+export function registerHook(helperName: string, hook: Function): HookUnregister {
```
rwjblue, line 16: Can we make the hook function types to be a bit more specific? For example, I'd like to have this indicate that we can return a promise from these hooks.

scalvert: Yep, I think that makes sense.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Registers a hook function to be run during the invocation of a test helper.
+ *
+ * @public
+ * @param {string} helperName The name of the test helper in which to run the hook.
+ * @param {Function} hook The hook function to run when the test helper is invoked.
+ * @returns {HookUnregister} An object containing an unregister function that will unregister
+ *                           the specific hook registered to the helper.
+ */
+export function registerHook(helperName: string, hook: Hook): HookUnregister {
```
ro0gr, line 17: `helperName` doesn't seem acccurate. It's used like `$helperName:$start_or_end` currently.
I'm wondering if we should split it to 2 args, like:

```ts
export function registerHook(helperName: string, checkpoint: string, hook: Hook): HookUnregister
```

or maybe even separate utils for `start` and `end`, like:
```ts
export function beforeEach(helperName: string, hook: Hook): HookUnregister
export function afterEach(helperName: string, hook: Hook): HookUnregister
```

or just rename the arg?

scalvert: I think your first suggestion actually makes sense. This API evolved a bit along the way, it used to just be the raw helper name. I think splitting these into `helperName` and `when` (`checkpoint`, or whatever) sounds reasonable.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Registers a hook function to be run during the invocation of a test helper.
+ *
+ * @public
+ * @param {string} helperName The name of the test helper in which to run the hook.
+ * @param {string} order The order, either `start` or `end`, to run the hook.
+ * @param {Function} hook The hook function to run when the test helper is invoked.
+ * @returns {HookUnregister} An object containing an unregister function that will unregister
+ *                           the specific hook registered to the helper.
+ */
+export function registerHook(helperName: string, order: string, hook: Hook): HookUnregister {
```
rwjblue, line 28: I like the second arg, but I'm worried about the name here. Can we rename it from `order` to `label`? 

Specifically, I'd like to add something that isn't "before" / "after" related to `waitUntil` (e.g. `runHooks('waitUntil', 'check')` each time we iterate), so naming this "order" is a bit bad.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const registeredHooks = new Map<string, Set<Hook>>();
+
+/**
+ *
+ * @param {string} helperName The name of the test helper in which to run the hook.
+ * @param {string} order The order, either `start` or `end`, to run the hook.
+ * @returns {string} The compound key for the helper.
+ */
+function getHelperKey(helperName: string, order: string) {
```
rwjblue, line 14: Can you add `@private` to this func?

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+  return `${helperName}:${order}`;
+}
+
+/**
+ * Registers a hook function to be run during the invocation of a test helper.
+ *
+ * @public
```
rwjblue, line 21: Lets make this `@private` while we iterate here.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+    unregister() {
+      hooksForHelper!.delete(hook);
+    },
+  };
+}
+
+/**
+ * Runs all hooks registered for a specific test helper.
+ *
+ * @public
```
rwjblue, line 49: Lets make this `@private` for now

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
scalvert: I added an `order` argument (I think order makes the most sense). I'm also going to validate the string passed in, to ensure people don't pass invalid `order` options (though I believe it won't error).

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
scalvert: I can do that. Label seems fine, though a tad more vague, but it is a bit more flexible in its usage.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
scalvert: No problem.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
scalvert: No problem.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
scalvert: No problem.

addon-test-support/@ember/test-helpers/index.ts
```diff
@@ -25,6 +25,7 @@ export { default as setupOnerror, resetOnerror } from './setup-onerror';
 export { getDebugInfo } from './-internal/debug-info';
 export { default as registerDebugInfoHelper } from './-internal/debug-info-helpers';
 export { default as getTestMetadata } from './test-metadata';
+export { registerHook } from './-internal/helper-hooks';
```
rwjblue, line 28: This export should probably be underscored while we figure things out, also I think we need to expose `runHooks` as well (so authors of custom helpers can participate in the same system).

```suggestion
export { registerHook as _registerHook, runHooks as _runHooks } from './-internal/helper-hooks';
```

tests/unit/dom/focus-test.js
```diff
@@ -35,6 +35,27 @@ module('DOM Helper: focus', function (hooks) {
...
+      assert.step('focus:end');
+    });
+
+    await focus(element);
+
+    assert.verifySteps(['focus:start', 'focus:end']);
+
+    startHook.unregister();
+    endHook.unregister();
+  });
```
rwjblue, line 57: Commenting here, but all of these tests need the same treatment:

```suggestion
  test('it executes registered focus hooks', async function (assert) {
    assert.expect(3);

    element = document.createElement('input');
    insertElement(element);
    
    let startHook = registerHook('focus', 'start', () => {
      assert.step('focus:start');
    });
    let endHook = registerHook('focus', 'end', () => {
      assert.step('focus:end');
    });

    try {
      await focus(element);

      assert.verifySteps(['focus:start', 'focus:end']);
    } finally {
      startHook.unregister();
      endHook.unregister();
    }
  });
```

This ensures that should a failure / error occur the rest of the tests aren't being broken...

tests/unit/dom/helper-hooks-test.js
```diff
@@ -0,0 +1,48 @@
...
+    let registeredHook = registeredHooks.get('click:start');
+
+    assert.ok(registeredHooks.has('click:start'));
+    assert.equal(registeredHook.size, 2);
+    assert.ok(registeredHook.has(func));
+
+    hook.unregister();
+
+    assert.notOk(registeredHook.has(func));
+  });
```
rwjblue, line 32: Instead of testing the implementation details, lets test what we expect folks to do (we don't _really_ care **how** the `runHooks`/`registerHook` implementation works, we care that it works):

```suggestion
  test('it can register an unregister a hook for a helper', function (assert) {
    let func = () => assert.step('click:start hook');
    let hook = registerHook('click', 'start', func);

    // it runs the hook
    runHooks('click', 'start');    
    assert.verifySteps(['click:start hook']);

    // can run multiple times
    runHooks('click', 'start');    
    assert.verifySteps(['click:start hook']);

    // unregister works
    hook.unregister();
    runHooks('click', 'start');    
    assert.verifySteps([]);
  });
```

Can you update all the other tests in this file accordingly?

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,64 @@
+type Hook = (...args: any[]) => void | Promise<void>;
+type HookUnregister = {
+  unregister: () => void;
+};
+
+export const registeredHooks = new Map<string, Set<Hook>>();
```
rwjblue, line 6: Once the unit tests have been updated to avoid mucking with this, we can stop exporting it at all:

```suggestion
const registeredHooks = new Map<string, Set<Hook>>();
```

scalvert: Done. I updated this in my local branch, so will ignore this suggestion.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param {string} label A label to help identify the hook. Built-in labels are `start` and `end`,
+ *                       designating the start of the helper invocation and the end.
+ * @param {any[]} args Any arguments originally passed to the test helper.
+ * @returns {Promise<void>} A promise representing the serial invocation of the hooks.
+ */
+export function runHooks(helperName: string, label: string, ...args: any[]): Promise<void> {
+  let hooks = registeredHooks.get(getHelperKey(helperName, label)) || new Set<Hook>();
+  let promise = Promise.resolve();
+
+  hooks.forEach(hook => promise.then(() => hook(...args)));
```
rwjblue, line 61: This leaks the promise at the moment, I think we need to change to:

```suggestion
  hooks.forEach(hook => promise = promise.then(() => hook(...args)));
```

Specifically, the original code would not allow a hook to return a promise and have `runHooks` properly wait for it to resolve. Can you add a test for this?

tests/integration/dom/scroll-to-test.js
```diff
@@ -23,6 +24,39 @@ module('DOM Helper: scroll-to', function (hooks) {
...
+        </ul>
+      </div>
+      `);
+
+    await scrollTo('.container', 0, 50);
+
+    assert.verifySteps(['scrollTo:start', 'scrollTo:end']);
+
+    startHook.unregister();
+    endHook.unregister();
```
rwjblue, line 57: Need these in a `finally`

tests/unit/dom/blur-test.js
```diff
@@ -44,6 +44,29 @@ module('DOM Helper: blur', function (hooks) {
...
+    let endHook = _registerHook('blur', 'end', () => {
+      assert.step('blur:end');
+    });
+
+    await blur(elementWithFocus);
+
+    assert.verifySteps(['blur', 'focusout', 'blur:start', 'blur:end']);
+
+    startHook.unregister();
+    endHook.unregister();
```
rwjblue, line 67: finally

tests/unit/dom/click-test.js
```diff
@@ -28,6 +28,27 @@ module('DOM Helper: click', function (hooks) {
...
+    let endHook = _registerHook('click', 'end', () => {
+      assert.step('click:end');
+    });
+
+    await click(element);
+
+    assert.verifySteps(['click:start', 'click:end']);
+
+    startHook.unregister();
+    endHook.unregister();
```
rwjblue, line 49: 🏁

tests/unit/dom/double-click-test.js
```diff
@@ -29,6 +29,27 @@ module('DOM Helper: doubleClick', function (hooks) {
...
+    let endHook = _registerHook('doubleClick', 'end', () => {
+      assert.step('doubleClick:end');
+    });
+
+    await doubleClick(element);
+
+    assert.verifySteps(['doubleClick:start', 'doubleClick:end']);
+
+    startHook.unregister();
+    endHook.unregister();
```
rwjblue, line 50: 🧟

tests/unit/dom/fill-in-test.js
```diff
@@ -35,6 +35,27 @@ module('DOM Helper: fillIn', function (hooks) {
...
+    let endHook = _registerHook('fillIn', 'end', () => {
+      assert.step('fillIn:end');
+    });
+
+    await fillIn(element, 'foo');
+
+    assert.verifySteps(['fillIn:start', 'fillIn:end']);
+
+    startHook.unregister();
+    endHook.unregister();
```
rwjblue, line 56: 🇫🇮

tests/unit/dom/helper-hooks-test.js
```diff
@@ -0,0 +1,54 @@
...
+    // it runs the hook
+    await _runHooks('click', 'start');
+    assert.verifySteps(['click:start hook']);
+
+    // can run multiple times
+    await _runHooks('click', 'start');
+    assert.verifySteps(['click:start hook']);
+
+    // unregister works
+    hook.unregister();
```
rwjblue, line 18: Needs to be in a `finally`, or maybe move to `afterEach`

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
rwjblue: Something like:

```js
  test('it can register a hook that returns a promise / has a delay', async function (assert) {
    let func = () => {
      assert.step('starting hook');

      return new Promise((resolve) => {
        setTimeout(() => {
          assert.step('resolving hook promise');
          resolve();
        }, 100);
      });
    };
    let hook = _registerHook('click', 'start', func);

    try {
      // it runs the hook
      assert.step('running hooks for click:start');
      await _runHooks('click', 'start');
      assert.step('hooks finished for click:start');

      assert.verifySteps([
        'running hooks for click:start',
        'starting hook',
        'resolving hook promise',
        'hooks finished for click:start',
      ]);
    } finally {
      hook.unregister();
    }
  });
```

scalvert: Good catch! I've added this.

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param {any[]} args Any arguments originally passed to the test helper.
+ * @returns {Promise<void>} A promise representing the serial invocation of the hooks.
+ */
+export function runHooks(helperName: string, label: string, ...args: any[]): Promise<void> {
+  let hooks = registeredHooks.get(getHelperKey(helperName, label)) || new Set<Hook>();
+  let promise = Promise.resolve();
+
+  hooks.forEach(hook => (promise = promise.then(() => hook(...args))));
+
+  return promise;
```
rwjblue, line 63: Chatted with @scalvert offline about this, he is going to tweak to something like this instead:

```js
let promises = [...hooks].map(
  (hook) => hook(...args)
);

return Promise.all(promises).then(() => {});
```

tests/unit/dom/helper-hooks-test.js
```diff
@@ -0,0 +1,85 @@
...
```
rwjblue, line 35: Also needs `try`/`finally` I think

tests/unit/dom/helper-hooks-test.js
```diff
@@ -0,0 +1,85 @@
...
+    let fooHook2 = _registerHook('click', 'foo', () => {
+      assert.step('click:foo2');
+    });
+
+    await _runHooks('click', 'foo');
+
+    assert.verifySteps(['click:foo1', 'click:foo2']);
+
+    fooHook1.unregister();
+    fooHook2.unregister();
```
rwjblue, line 83: Also needs `try`/`finally`

fivetanley:
This would be a very handy thing to have for our accessibility testing also (we're hoping to contribute some back to ember-a11y-testing but still learning a lot). Is there a path to this become a stable exported api? (I noticed the words "internal" and also exports with underscores so I assume I shouldn't use them, but I would love to be wrong)

scalvert:
@fivetanley this is actually precisely one of the use cases we're looking to support! The plan is definitely to have this be a publicly, supported API. We typically iterate on this type of API actively before we publish it for general consumption, as that helps validate our assumptions about the functionality. 

Please let us know if there's anything you need from these APIs that we're missing.

PR #879, opened 2020-06-04
fix: Stabilize test suite isolation

scalvert:
There were a few issues this PR is attempting to fix:

1. Correct reset `Application` when it's set to `null`
1. Correctly return from `waitUntil` when we reject so we stop recursing
1. Correctly sets `element` to `null` in tests where element is used.

### DIFF
--- a/addon-test-support/@ember/test-helpers/wait-until.ts
+++ b/addon-test-support/@ember/test-helpers/wait-until.ts
@@ -59,6 +59,7 @@ export default function waitUntil<T>(
           value = callback();
         } catch (error) {
           reject(error);
+          return;
         }
 
         if (value) {
@@ -67,6 +68,7 @@ export default function waitUntil<T>(
           scheduleCheck(timeoutsIndex + 1);
         } else {
           reject(waitUntilTimedOut);
+          return;
         }
       }, interval);
     }
--- a/tests/unit/setup-rendering-context-test.js
+++ b/tests/unit/setup-rendering-context-test.js
@@ -28,7 +28,7 @@ module('setupRenderingContext', function (hooks) {
     return;
   }
 
-  hooks.after(function () {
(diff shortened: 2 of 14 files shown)


addon-test-support/@ember/test-helpers/wait-until.ts
```diff
@@ -59,6 +59,7 @@ export default function waitUntil<T>(
           value = callback();
         } catch (error) {
           reject(error);
+          return;
```
rwjblue, line 62: This was a doosey of a thing to debug 😭. Basically, without this change `waitUntil` would run **forever** if the callback ever threw an error. 😭

This probably needs to be backported to the 1.x branch.

tests/unit/setup-rendering-context-test.js
```diff
@@ -28,7 +28,7 @@ module('setupRenderingContext', function (hooks) {
     return;
   }
 
-  hooks.after(function () {
+  hooks.afterEach(function () {
```
rwjblue, line 31: Using `hooks.after` here meant that the tests _within_ this module were not able to assume `owner.visit` was available (because when `setApplication(null)` is ran, we use a "fake owner" which doesn't have the ability to `visit`).

tests/unit/wait-until-test.js
```diff
@@ -69,4 +70,19 @@ module('DOM helper: waitUntil', function () {
       assert.equal(reason.message, 'error goes here', 'valid error was thrown');
     });
   });
+
+  test('does not continue to wait if callback throws', async function (assert) {
+    assert.expect(2);
```
rwjblue, line 75: For context, prior to the `waitUntil` changes above, this test failed with a large number of additional assertions (> 10+) because `waitUntil` would happily continue looping.

PR #907, opened 2020-08-17
Backport: Ensure `waitUntil` stops recursing when the callback throws an error

rwjblue:
Brings the changes from https://github.com/emberjs/ember-test-helpers/pull/879 back to v1.x.


rwjblue:
The actual changes here are super minimal (adding  two `return` statements where they clearly belong), but CI is going to be red due to many down stream packages. 😭

PR #914, opened 2020-08-25
Migrate to using @ember/destroyable for teardown.

rwjblue:
Moves actual logic from `teardownContext` into `setupContext` by using `registerDestructor` and `associateDestroyableChild`.

### DIFF
--- a/addon-test-support/@ember/test-helpers/teardown-context.ts
+++ b/addon-test-support/@ember/test-helpers/teardown-context.ts
@@ -1,9 +1,7 @@
-import { run } from '@ember/runloop';
-import { _teardownAJAXHooks } from './settled';
-import { unsetContext, TestContext } from './setup-context';
+import { TestContext } from './setup-context';
 import { nextTickPromise } from './-utils';
 import settled from './settled';
-import Ember from 'ember';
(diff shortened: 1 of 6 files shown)


addon-test-support/@ember/test-helpers/teardown-context.ts
```diff
@@ -28,28 +29,14 @@ export default function teardownContext(
...
-      (Ember as any).testing = false;
-
-      unsetContext();
-
-      if (waitForSettled) {
-        return settled();
-      }
-
-      return nextTickPromise();
+      destroy(context);
```
scalvert, line 35: Were we only waiting for `settled` before because we were using runloop APIs?

scalvert:
You have a linting failure with an unused import.

addon-test-support/@ember/test-helpers/teardown-context.ts
rwjblue: If you look just below, we are still `return await settled()` in the finally. I honestly do not know why we were doing it twice.

rwjblue:
Currently blocked by https://github.com/emberjs/ember-test-helpers/pull/915

PR #916, opened 2020-08-27
fix typo in example that calles `new File(...)`

luxferresum:
The `File` constructor takes 2 arguments, first an array with data, second the name.


rwjblue (approved):
Thank you!

PR #918, opened 2020-08-28
Blur the previous active element on focus

ro0gr:
implements [suggestion](https://github.com/emberjs/ember-test-helpers/issues/692#issuecomment-617563815) by @Turbo87 

I don't feel strongly if we need to cover by tests each helper which uses `__focus__` internally. Please, let me know if we should.


rwjblue (approved):

rwjblue:
Thank you @ro0gr!

PR #922, opened 2020-09-20
Remove deprecated ember-cli-htmlbars-inline-precompile for v1-x

dmytro-krekota:
In this pull request, I removed usage of ember-cli-htmlbars-inline-precompile in the currently latest stable v1 version.
Inspired by https://github.com/emberjs/ember-test-helpers/pull/821

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "broccoli-funnel": "^2.0.2",
     "ember-assign-polyfill": "^2.6.0",
     "ember-cli-babel": "^7.7.3",
-    "ember-cli-htmlbars": "~5.2.0",
+    "ember-cli-htmlbars": "^4.4.0",
     "ember-test-waiters": "^1.1.1"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "broccoli-funnel": "^2.0.2",
     "ember-assign-polyfill": "^2.6.0",
     "ember-cli-babel": "^7.7.3",
-    "ember-cli-htmlbars-inline-precompile": "^2.1.0",
+    "ember-cli-htmlbars": "^5.3.1",
     "ember-test-waiters": "^1.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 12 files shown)


dmytro-krekota:
Tests were passed: https://travis-ci.org/github/emberjs/ember-test-helpers/jobs/728739086

package.json
```diff
@@ -38,7 +38,7 @@
     "broccoli-funnel": "^2.0.2",
     "ember-assign-polyfill": "^2.6.0",
     "ember-cli-babel": "^7.7.3",
-    "ember-cli-htmlbars-inline-precompile": "^2.1.0",
+    "ember-cli-htmlbars": "^4.4.0",
```
rwjblue, line 41: I think the problem with this (for the 1.x series) is with Node version support.

`ember-cli-htmlbars@4` supports Node 8, 10, >= 12, but the 1.x branch of `@ember/test-helpers` supports Node 6.

https://github.com/ember-cli/ember-cli-htmlbars/blob/v4.4.0/package.json#L95

https://github.com/emberjs/ember-test-helpers/blob/02bb484bb91f5cc7413ee38b0ad026e8d3010b22/package.json#L90

rwjblue (commented):
I don't think we can land this in the 1.x series, it is a breaking change.

package.json
dmytro-krekota: Ok, thanks for the explanation, I see the problem now.

Then just want to share my problem. In my project, I try to upgrade dependencies and receive the next error in tests: 
"Could not find module `htmlbars-inline-precompile` imported from `@ember/test-helpers/setup-rendering-context`".
I suspect that my PR can solve it or somehow need to use the not stable beta release of the library.

Maybe you or somebody can help me to fix it in another way.

Screenshot 2020-09-20 at 23 30 19

dmytro-krekota: Looks like it is an issue related to updates in ember-cli-mirage: 1.1.4 => 1.1.6, so this PR can be closed.

PR #923, opened 2020-10-07
Make it possible to call `triggerEvent(fileInput, 'change', { files })` twice with the same files

### LINKED ISSUE #921: `triggerEvent(fileInput, 'change', { files })` cannot be called twice with the same files
```js
const files = [ new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' }) ];
await triggerEvent(fileInput, 'change', { files });
await triggerEvent(fileInput, 'change', { files });

// -> Uncaught TypeError: Cannot redefine property: item
//    at Function.defineProperty (<anonymous>)
//    at buildFileEvent (fire-event.js:212)
```

The workaround is to clone the `files` array:
```
await triggerEvent(fileInput, 'change', { files: files.toArray() });
await triggerEvent(fileInput, 'change', { files: files.toArray() });
```

It seems like `files.item` property should be `configurable: true`:
https://github.com/emberjs/ember-test-helpers/blob/e599e64b203104566a452a28eff5f79b4c6f8896/addon-test-support/%40ember/test-helpers/dom/fire-event.ts#L261-L264

sdahlbac:
Fixes #921 

Not really sure what I did wrong the previous time, now it seems to work fine.

Also moved the testcase to a better location as suggested.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fire-event.ts
+++ b/addon-test-support/@ember/test-helpers/dom/fire-event.ts
@@ -262,7 +262,7 @@ function buildFileEvent(
       value(index: number) {
         return typeof index === 'number' ? this[index] : null;
       },
-      configurable: true
+      configurable: true,
     });
     Object.defineProperty(element, 'files', {
       value: files,
--- a/addon-test-support/@ember/test-helpers/dom/fire-event.ts
+++ b/addon-test-support/@ember/test-helpers/dom/fire-event.ts
@@ -262,6 +262,7 @@ function buildFileEvent(
       value(index: number) {
         return typeof index === 'number' ? this[index] : null;
       },
+      configurable: true
     });
     Object.defineProperty(element, 'files', {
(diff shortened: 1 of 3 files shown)


rwjblue:
I think if you have the time that you can update the defineProperty that we do to add `configurable` that would do the trick.

Mind giving that a shot?

sdahlbac:
I thought that was what I tried with the results in the PR comment, but I can give it another go

rwjblue:
Oh, I see! I'm sorry, I missed that those failures were with that change.

nag5000:
> I also tried to apply the suggested fix, but ended up with different test failures, unsure how to fix.

@sdahlbac Just use the `element` variable instead of local `fileInput`.

Also, it seems `/tests/unit/dom/select-files-test.js` is a better place for the new testcase.

rwjblue (approved):
Awesome, thank you!!

PR #927, opened 2020-10-29
Fixes interop issue between native promises and RSVP promises in test helper hooks

drewlee:
For reasons I'm not too sure of, the interoperability between the RSVP promise which is utilized by `nextTickPromise` and the native Promise used at: https://github.com/emberjs/ember-test-helpers/blob/657338372e8344dd2ea50f00d3a72a42ec5c54f3/addon-test-support/%40ember/test-helpers/-internal/helper-hooks.ts#L62 causes numerous component test failures throughout our test suites. Using RSVP's `Promise.all` instead, fixes the test failures.


drewlee:
@scalvert @rwjblue if you could take a look. Thanks!

scalvert (approved):

PR #928, opened 2020-10-29
Fixes missing RSVP promise usage in helper hooks

scalvert:
The helpers hooks were incorrectly using native promise vs. RSVP promise, which causes the new release of `@ember/test-helpers` to not be IE compatible. This change fixes that, where we're now using RSVP promises to chain.


scalvert:
Closing in favor of #927

PR #941, opened 2020-11-07
triggerKeyEvent() #=> keyCodeFromKey() optimization

izelnakri:
- In the first commit replaced addon-test-support/@ember/test-helpers/dom/trigger-key-event.ts#120 `.filter()[0]` to `find()` for short circuit optimization.
- In the second commit made the second check lazy with `||`.


rwjblue (commented):
Unfortunately, I think `Array.find` is not supported on IE11 (which we are still trying to support).

izelnakri:
@rwjblue  closed in favor for polyfilled implementation: https://github.com/emberjs/ember-test-helpers/pull/949

PR #943, opened 2020-11-10
Allow `click`, `doubleClick`, and `triggerEvent` to trigger events on `window`.

scalvert:
During [a refactor](https://github.com/emberjs/ember-test-helpers/commit/0233a7fcb3a663c9cbcd84c1385bb864fc68676e) that improved typings in this project, the `getElement` function was narrowed in what it would return - specifically it was correctly fixed to return elements, of which `window` is not. This had the side effect of removing support for triggering window events via `triggerEvent`, in addition to removing support for other helpers. 

This PR restores that capability by adding a new `getTarget` util, which wraps the `getElement` util but additionally checks for the presence of window, and subsequently returns that in favor of an element.

TODO:

Update `getElement` calls in other helpers that are valid for `window`:
- [x] `triggerEvent`
- [x] `click`
- [x] `doubleClick`


addon-test-support/@ember/test-helpers/dom/trigger-event.ts
```diff
@@ -72,13 +72,17 @@ export default function triggerEvent(
...
 
-      let element = getElement(target);
+      let element = getTarget(target);
       if (!element) {
         throw new Error(`Element not found when calling \`triggerEvent('${target}', ...)\`.`);
       }
 
-      if (isFormControl(element) && element.disabled) {
-        throw new Error(`Can not \`triggerEvent\` on disabled ${element}`);
+      if (!isWindow(element)) {
```
rwjblue, line 80: Let's drop the helper function, and do the check here directly (I don't think the helper function gives us anything useful).

```js
let element = target instanceof Window ? target : getElement(target);
if (!element) {
  throw new Error(`Element not found when calling \`triggerEvent('${target}', ...)\`.`)
}
```

rwjblue (changes requested):

rwjblue:
We should evaluate the other DOM helpers, see which should allow `window` (I think **most** should not).

addon-test-support/@ember/test-helpers/dom/-target.ts
```diff
@@ -20,3 +23,18 @@ export function isDocument(target: any): target is Document {
...
+  @param {string|Element} target the window, an element or selector to retrieve
+  @returns {Element|Window} the target or selector
+*/
+export function getTarget(target: Target): Element | Document | Window | null {
+  if (isWindow(target)) {
+    return target as Window;
+  }
+
+  return getElement(target);
+}
```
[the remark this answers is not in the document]
scalvert: I need the same logic for `click` and `doubleClick`, so I think that's it's use.

addon-test-support/@ember/test-helpers/dom/trigger-event.ts
scalvert: It's useful for using it in a few different helpers, such as `click` and `doubleClick`. I have to add the same check there, so this helper function was in preparation for that. The next commit should make it more clear (if the TODO didn't).

scalvert:
@rwjblue I did. It looks as though `click` and `doubleClick` are the only other ones that make sense with `window`. Updating the PR to include those.

addon-test-support/@ember/test-helpers/dom/-target.ts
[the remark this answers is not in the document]
scalvert: See recent updates.

addon-test-support/@ember/test-helpers/dom/trigger-event.ts
scalvert: See recent updates.

addon-test-support/@ember/test-helpers/dom/-target.ts
```diff
@@ -20,3 +27,18 @@ export function isDocument(target: any): target is Document {
...
 }
+
+/**
+  Used internally by the DOM interaction helpers to find either window or an element.
+
+  @private
+  @param {string|Element} target the window, an element or selector to retrieve
+  @returns {Element|Window} the target or selector
+*/
+export function getTarget(target: Target): Element | Document | Window | null {
```
rwjblue, line 38: Hmm, `getTarget` seems odd to me (as a name here). Maybe `getWindowOrElement` (definitely makes more sense to what it is)?

addon-test-support/@ember/test-helpers/dom/click.ts
```diff
@@ -31,11 +30,15 @@ export const DEFAULT_CLICK_OPTIONS = {
   @param {Element} element the element to click on
   @param {MouseEventInit} options the options to be merged into the mouse events
 */
-export function __click__(element: Element | Document, options: MouseEventInit): void {
+export function __click__(element: Element | Document | Window, options: MouseEventInit): void {
   fireEvent(element, 'mousedown', options);
 
-  if (isFocusable(element)) {
-    __focus__(element);
+  if (!isWindow(element)) {
```
rwjblue, line 36: Why do we need this? If you widen `isFocusable` and `isFormControl` to allow passing `window` and just return `false` wouldn't that be fine? I think if we do that it may simplify the rest of the code?

rwjblue (commented):
You've labeled this as `breaking`, but how is it breaking? It seems that we are widening the types / allowed values (you can now do a thing you couldn't before), no?

addon-test-support/@ember/test-helpers/dom/-target.ts
scalvert: Agreed.

addon-test-support/@ember/test-helpers/dom/click.ts
scalvert: I thought about that too, but it pulls the thread all the way through those APIs, which also feel odd to widen to accept window.

scalvert: OK as we discussed offline, I've moved to using a type guard on the `isWindow` function, which has negated the need for this.

addon-test-support/@ember/test-helpers/dom/-target.ts
scalvert: Renamed.

rwjblue (approved):

PR #944, opened 2020-11-11
dom/-to-array.ts replaced with Array.from()

izelnakri:
Removed redundant code. We can now use native `Array.from()`.


rwjblue:
> We can now use native Array.from().

Hmm, I didn't think that `Array.from` was supported by IE11 🤔

izelnakri:
> Hmm, I didn't think that `Array.from` was supported by IE11

Yes its still not supported, however the current browser support is 94.59%: https://caniuse.com/?search=Array.from

Array.from is easy to shim, and native `Array.from` must be much faster since its native, thats why I removed it.

rwjblue:
Sure, but we still support IE11 (and do not require a polyfill).

izelnakri:
@rwjblue closed in favor for polyfilled implementation: https://github.com/emberjs/ember-test-helpers/pull/949

PR #945, opened 2020-11-18
Use RSVP promises instead of `_Promise` class extension in helper hooks

drewlee:
Still seeing a few strange async related test failures throughout a bunch of tests. For whatever reason, using RSVP promises directly resolves these failures. Applying this change has somewhat reduced the overall number of failing tests we've been seeing after upgrading to the new test dependencies.

### DIFF
--- a/addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
+++ b/addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
@@ -1,4 +1,4 @@
-import { _Promise as Promise } from '../-utils';
+import { Promise } from 'rsvp';
 
 type Hook = (...args: any[]) => void | Promise<void>;
 type HookLabel = 'start' | 'end' | string;


rwjblue:
Hmm. The internal promise is still an RSVP promise (just one that can resolve without run loop shenanigans).

drewlee:
Yeah, I realize this change is somewhat naive. It likely just masks another issue by forcing async coalescence via the run loop. Will continue to investigate.

rwjblue:
Ya, we are going to have to run these async interaction issues down I think.

scalvert:
Ya I'd be interested to see if we can narrow down repros for this. This'll really help us identify what's going on. I'd be concerned to make a change like this without understanding the why, since it _is_ just extending from an `RSVP.Promise`.

scalvert:
@drewlee did you determine this wasn't an issue?

drewlee:
@scalvert no, will reopen or create a new PR once the exact cause is determined

PR #946, opened 2020-11-21
Fixes maxlength conditional check for textarea & constrained input types

drewlee:
Currently, the `maxlength` guard for `typeIn` and `fillIn` helpers doesn't actually execute for `textarea` elements, nor does it discriminate between constrained and non-constrained input types. The issue is actually from an [inverted check](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/-guard-for-maxlength.ts#L16-L17) in the conditional. I've also added more tests to catch this sort of thing. The non-constrained test is slightly contrived, but it would definitely catch a problem if the input type isn't properly verified.


rwjblue (approved):
Thanks for fixing this @drewlee!

PR #949, opened 2020-11-29
Polyfilled test selector optimizations 

izelnakri:
- In the first commit we introduce a polyfill for ie11 to support `Array.from` and `Array.prototype.find`.
- In the second commit we optimize `findAll()` and `waitFor` to use `Array.from`
- In the third commit we optimize `triggerKeyEvent` to use `array.find()` instead of `array.filter(predicate)[0]`;

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/find-all.ts
+++ b/addon-test-support/@ember/test-helpers/dom/find-all.ts
@@ -1,5 +1,5 @@
 import getElements from './-get-elements';
-import toArray from './-to-array';
+import { toArray } from '../ie-11-polyfills';
 
 /**
   Find all elements matched by the given selector. Similar to calling
--- a/addon-test-support/@ember/test-helpers/dom/trigger-key-event.ts
+++ b/addon-test-support/@ember/test-helpers/dom/trigger-key-event.ts
@@ -8,6 +8,7 @@ import Target from './-target';
 import { log } from '@ember/test-helpers/dom/-logging';
 import isFormControl from './-is-form-control';
 import { runHooks, registerHook } from '../-internal/helper-hooks';
+import { find } from '../ie-11-polyfills';
 
 registerHook(
   'triggerKeyEvent',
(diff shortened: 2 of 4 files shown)


scalvert:
Thanks, @izelnakri. Would you mind rebasing/resolving conflicts?

izelnakri:
@scalvert rebased :+1:

scalvert:
Thanks! One thing I like better about this approach than the first PRs is that once we eventually drop support for IE11, we can just delete the polyfills.

scalvert (approved):

rwjblue (commented):
I worry a little bit about automatically polyfilling these array methods. If a consuming app that intends to support IE11 adds some usage in their app of `[].find` (for example) the tests will pass (even if they are testing against IE11) but when deployed the app will throw/fail.

🤔

izelnakri:
This is a good point i think but bit rare case. We could make them exported find() and/or toArray() from the same module name, thats an option. Solves both issues, what do you think @rwjblue?

‘’’js
export function find(array, func) {
  return array?.find?.(func) || findFunction(array, func);
}

export function toArray(array, func) {

}
‘’’

rwjblue:
Ya, that seems good to me @izelnakri. Thank you!

izelnakri:
@rwjblue updated the changes. 

In the CI tests are passing, however for other CI actions there is a weird broccoli specific error that I couldn't understand.  It doesn't seem to be related to the changes, would appreciate if you take a look: https://github.com/emberjs/ember-test-helpers/pull/949/checks?check_run_id=1542886060#step:5:121

scalvert:
@izelnakri this was a bug in ember-auto-import if I'm not mistaken. I think there's been [a fix that's shipped already](https://github.com/ef4/ember-auto-import/pull/330). It looks like it was released in 1.9.0 of that package.

scalvert:
@izelnakri if you rebase this branch onto master I think you should be g2g.

izelnakri:
@scalvert done, thanks!

rwjblue (approved):
Awesome, thanks for working through this @izelnakri!

Issue #947, opened 2020-11-24
Helpers no longer work properly with scheduled run-loop queues in v2


drewlee:
Test helpers, particularly `render`, seem to be out of sync with scheduled run-loop timers. This causes integration and acceptance tests to fail when a component schedules work on a queue (e.g., `afterRender`).

I've create a simple use case to demonstrate the issue. It consists of a basic component that schedules programmatic focus placement on an `afterRender` queue. It is very similar to some of the test failures I've noticed within our applications after upgrading to v2: https://github.com/drewlee/helpers-async.

[There are two tests for the component.](https://github.com/drewlee/helpers-async/blob/master/tests/integration/components/basic-search-test.js) The first demonstrates the regression. What's particularly interesting is that running `getSettledState()` after `render`/`click` shows that the app still consists of pending timers and run-loop. The second test shows that a test can be coerced into passing by adding a `settled` waiter immediately after calling `render`. However, the app is still not quite settled after `click` resolves, so this seems to be symptomatic of all test helpers.

<img width="1392" alt="Screen Shot 2020-11-24 at 12 41 12 AM" src="https://user-images.githubusercontent.com/945868/100069590-ca550280-2ded-11eb-9f14-d605152a7618.png">

I'm not exactly too sure whether the test helpers resolve prematurely before the run-loop stuff is complete, or whether the test helpers somehow trigger a new run loop. Especially given that the component's behavior is altered because the scheduled `focus` call ends up executing before the input is rendered (hence the `Cannot read property 'focus' of null` exception).

What I did find, is that removing this [noop then callback in helper-hooks.js](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/-internal/helper-hooks.ts#L64) resolves the issue. I'm fairly certain that it's there for a good reason, so I'm somewhat hesitant for suggesting its removal as this may be a symptom of an issue related to Ember source, Backburner, or even RSVP. It seems very enigmatic for the stuff we're trying to avoid in this [RSVP config](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/-utils.ts#L55) though.

drewlee:
@scalvert this is the bug that I've been trying to pin down for a while. I've finally traced it down to the fact that all of the test failures were from components that schedule stuff on a run-loop queue.

scalvert:
[...]

PR #958, opened 2020-12-04
Ensure `@ember/test-helpers` promise based helpers never create a run loop for resolving the promise.

rwjblue:
This removes the custom `RSVP.configure('async', fn)` that we have used for a while (introduced in https://github.com/emberjs/ember-test-helpers/commit/0da35f418a603eb060b1c89ae2b0b778fe2cef0f) in favor of using native `Promise` whenever possible and falling back to a separately loaded (via `ember-auto-import`) `es6-promise` polyfill when native Promise is not available.

This does increase the overall size of `assets/test-support.js` by ~ 27kb ([see the es6-promise README](https://github.com/stefanpenner/es6-promise)), but since that doesn't really affect any production application situations it should be fine. Once we drop support for browsers that do not have a native `Promise` we can remove `ember-auto-import` and our `es6-promise` polyfill.

In order to test that this implementation functions properly in IE11, I had to get the testing infrastructure back into a state where IE11 can run tests. That required a number of changes:

* https://github.com/emberjs/ember-test-waiters/pull/213
* https://github.com/machty/ember-concurrency-decorators/pull/85
* Fixup syntax using IE11 in this repo

Closes https://github.com/emberjs/ember-test-helpers/pull/956
Fixes https://github.com/emberjs/ember-test-helpers/issues/947


scalvert:
cc/ @drewlee

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
```diff
@@ -59,7 +59,13 @@ export function registerHook(helperName: string, label: HookLabel, hook: Hook):
  */
 export function runHooks(helperName: string, label: HookLabel, ...args: any[]): Promise<void> {
   let hooks = registeredHooks.get(getHelperKey(helperName, label)) || new Set<Hook>();
-  let promises = [...hooks].map(hook => hook(...args));
+  let promises = [];
+
+  hooks.forEach(hook => {
```
scalvert, line 64: 👎 but 🤷

scalvert: IE is the 👎

scalvert (approved):
This looks 👍 and is largely what was discussed. Thanks!

addon-test-support/@ember/test-helpers/-internal/helper-hooks.ts
rwjblue: 👎 for IE11 🎉

Issue #964, opened 2020-12-15
Dispatched `blur` event causes false test failures for components using `event.relatedTarget` checks


drewlee:
This regression was introduced in v2.0.0 as a result of PR https://github.com/emberjs/ember-test-helpers/pull/918.

It's fairly common for UI overlay-type widgets (e.g., dropdowns, tooltips, typeaheads, etc.), to utilize checking of the `event.relatedTarget` property to control the visibility and state of the associated overlay elements. For example, within the context of a `focusout` event handler, the `relatedTarget` property references the element which just gained focus. This is useful to determine whether the newly focused element is a descendent of a component, as demonstrated at https://github.com/drewlee/eth-rtarget/blob/5a083c65b323e9261ff69b04d217f1d3915cf993/app/components/basic-dd.js#L16. A non-descendant element signifies that the associated overlay can be dismissed. This is very practical for handling interactions where a user Tabs out of a dropdown menu, or clicks outside of a particular widget.

However, since introducing [`blur` to fire before each invocation of `focus`](https://github.com/emberjs/ember-test-helpers/blob/b4dd8c2dcfa840a885581b9f42deabddace90d39/addon-test-support/%40ember/test-helpers/dom/focus.ts#L26-L32), this behavior is no longer working properly in tests, causing overlays to prematurely dismiss. Since the `event.relatedTarget` property from the simulated event  has a `null` value, the component assumes that focus has moved out of the browser window and dismisses.

A reproduction of this case has been created at https://github.com/drewlee/eth-rtarget with an associated test case at https://github.com/drewlee/eth-rtarget/blob/5a083c65b323e9261ff69b04d217f1d3915cf993/tests/integration/components/basic-dd-test.js#L9. For comparison purposes, there's a branch using an older version of @ember/test-helpers, where the test passes as expected: https://github.com/drewlee/eth-rtarget/tree/old-deps.

rwjblue:
In order to fix this, we will have to significantly change the strategy from #918. Instead of calling the existing `__blur__` shared function, we will need to fire the appropriate events manually (via the `fireEvent` mechanism) so that we can set the proper options.

ro0gr:
[...]

PR #965, opened 2020-12-15
Fixes event.relatedTarget for dispatched blur events (#964)

drewlee:
This is to address issue #964. The proposed fix is to update the `__blur__` function so that the `event.relatedTarget` property can be set when called from `__focus__`.

As far as I can tell, `element.focus()` automatically triggers `blur` with the correct `event` parameters when the browser is active. This is why it is necessary to wrap the `element.blur();` invocation in an additional check to avoid duplicate triggers of `blur` and `focusout`.

When the browser is not active, we set the `relatedTarget` property for the simulated event (`fireEvent`).

While I did consider calling `fireEvent` directly from https://github.com/drewlee/ember-test-helpers/blob/d0edc5af972cc9fb31c22cd4f48cd9fc0947cac3/addon-test-support/%40ember/test-helpers/dom/focus.ts#L31, I found this undesirable as `blur` and `focusout` also got triggered duplicate times when the browser is active.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/blur.ts
+++ b/addon-test-support/@ember/test-helpers/dom/blur.ts
@@ -37,7 +37,7 @@ export function __blur__(
   // Chrome/Firefox does not trigger the `blur` event if the window
   // does not have focus. If the document does not have focus then
   // fire `blur` event via native event.
-  if (browserIsNotFocused) {
+  if (browserIsNotFocused || needsCustomEventOptions) {
     let options = { relatedTarget };
     fireEvent(element, 'blur', assign({ bubbles: false }, options));
     fireEvent(element, 'focusout', options);
--- a/addon-test-support/@ember/test-helpers/dom/blur.ts
+++ b/addon-test-support/@ember/test-helpers/dom/blur.ts
@@ -1,3 +1,4 @@
+import { assign } from '@ember/polyfills';
 import getElement from './-get-element';
 import fireEvent from './fire-event';
 import settled from '../settled';
@@ -14,24 +15,32 @@ registerHook('blur', 'start', (target: Target) => {
 /**
   @private
   @param {Element} element the element to trigger events on
+  @param {Element} relatedTarget the element that is focused after blur
 */
-export function __blur__(element: HTMLElement | Element | Document | SVGElement): void {
+export function __blur__(
+  element: HTMLElement | Element | Document | SVGElement,
+  relatedTarget: HTMLElement | Element | Document | SVGElement | null = null
+): void {
   if (!isFocusable(element)) {
     throw new Error(`${element} is not focusable`);
   }
 
   let browserIsNotFocused = document.hasFocus && !document.hasFocus();
+  let needsCustomEventOptions = relatedTarget !== null;
 
-  // makes `document.activeElement` be `body`.
-  // If the browser is focused, it also fires a blur event
-  element.blur();
+  if (!needsCustomEventOptions) {
+    // makes `document.activeElement` be `body`.
+    // If the browser is focused, it also fires a blur event
+    element.blur();
+  }
 
   // Chrome/Firefox does not trigger the `blur` event if the window
   // does not have focus. If the document does not have focus then
   // fire `blur` event via native event.
   if (browserIsNotFocused) {
-    fireEvent(element, 'blur', { bubbles: false });
-    fireEvent(element, 'focusout');
+    let options = { relatedTarget };
+    fireEvent(element, 'blur', assign({ bubbles: false }, options));
+    fireEvent(element, 'focusout', options);
   }
 }
 
--- a/addon-test-support/@ember/test-helpers/dom/focus.ts
+++ b/addon-test-support/@ember/test-helpers/dom/focus.ts
@@ -23,12 +23,15 @@ export function __focus__(element: HTMLElement | Element | Document | SVGElement
 
   let browserIsNotFocused = document.hasFocus && !document.hasFocus();
 
+  // fire __blur__ manually with the correct relatedTarget when the browser is not
+  // already in focus and there was a previously focused element
   if (
     document.activeElement &&
     document.activeElement !== element &&
-    isFocusable(document.activeElement)
+    isFocusable(document.activeElement) &&
+    browserIsNotFocused
   ) {
-    __blur__(document.activeElement);
+    __blur__(document.activeElement, element);
   }
 
   // makes `document.activeElement` be `element`. If the browser is focused, it also fires a focus event
--- a/tests/unit/dom/focus-test.js
+++ b/tests/unit/dom/focus-test.js
@@ -105,13 +105,13 @@ module('DOM Helper: focus', function (hooks) {
     const focusedElement = document.createElement('textarea');
     insertElement(focusedElement);
     focusedElement.focus();
-    instrumentElement(focusedElement, ['target']);
+    instrumentElement(focusedElement, ['target', 'relatedTarget']);
 
     await focus(element);
 
     assert.verifySteps([
       ...blurSteps.map(s => {
-        return `${s} [object HTMLTextAreaElement]`;
+        return `${s} [object HTMLTextAreaElement] [object HTMLInputElement]`;
       }),
       ...focusSteps.map(s => {
         return `${s} [object HTMLInputElement]`;


addon-test-support/@ember/test-helpers/dom/blur.ts
```diff
@@ -24,14 +29,17 @@ export function __blur__(element: HTMLElement | Element | Document | SVGElement)
 
   // makes `document.activeElement` be `body`.
   // If the browser is focused, it also fires a blur event
-  element.blur();
+  if (!relatedTarget) {
```
ro0gr, line 32: I might miss something, but it looks like the `__blur__(` would be no-op, in case if there is no `relatedTarget` and browser is focused?

drewlee: `relatedTarget` defaults to `null` if it is not provided, meaning that `element.blur();` will execute if it's not set by way of the `!` operator. I can make this a stricter comparison so that it's clearer, for example:
`if (relatedTarget === null) {`

ro0gr:
I think it'd be nice if we test the `relatedTarget`. We should probably just log it in https://github.com/emberjs/ember-test-helpers/blob/b4dd8c2dcfa840a885581b9f42deabddace90d39/tests/unit/dom/focus-test.js#L103, like:
```ts
element = buildInstrumentedElement('input', ['target', 'relatedTarget']);
```

drewlee:
Alternatively, we could just do:
```
if (
    document.activeElement &&
    document.activeElement !== element &&
    isFocusable(document.activeElement) &&
    browserIsNotFocused
  ) {
    let options = { relatedTarget: element };
    fireEvent(document.activeElement, 'blur', assign({ bubbles: false }, options));
    fireEvent(document.activeElement, 'focusout', options);
  }
```
which would simplify the implementation a bit.

drewlee:
Updated the test to verify `blur` steps for `relatedTarget`. Verified tests pass when browser is active, and not active.

rwjblue:
https://github.com/emberjs/ember-test-helpers/pull/966 should have fixed CI

@drewlee - Can you rebase?

addon-test-support/@ember/test-helpers/dom/blur.ts
rwjblue: Ya, I'd vaguely prefer to check directly against `null`. I _might_ even tweak to clarify with a variable 🤔. Something like:

```suggestion
  let relatedTargetProvided = relatedTarget !== null;
  if (!relatedTargetProvided) {
```

rwjblue: I'm not 100% sure that this case was what @ro0gr was mentioning, but I do think we still have a logic problem here @drewlee. When `relatedElement` is passed as a non-null value, **and** the browser is in focus we never fire `blur` event.

addon-test-support/@ember/test-helpers/dom/blur.ts
```diff
@@ -24,14 +29,17 @@ export function __blur__(element: HTMLElement | Element | Document | SVGElement)
...
   // Chrome/Firefox does not trigger the `blur` event if the window
   // does not have focus. If the document does not have focus then
   // fire `blur` event via native event.
   if (browserIsNotFocused) {
-    fireEvent(element, 'blur', { bubbles: false });
-    fireEvent(element, 'focusout');
+    let options = { relatedTarget };
+    fireEvent(element, 'blur', assign({ bubbles: false }, options));
+    fireEvent(element, 'focusout', options);
   }
```
rwjblue, line 43: I'm thinking we need something like:

```ts
  let browserIsNotFocused = document.hasFocus && !document.hasFocus();
  let needsCustomEventOptions = relatedTarget !== null;

  if (!needsCustomEventOptions) {
    // makes `document.activeElement` be `body`.
    // If the browser is focused, it also fires a blur event
    element.blur();
  }

  // Chrome/Firefox does not trigger the `blur` event if the window
  // does not have focus. If the document does not have focus then
  // fire `blur` event via native event.
  if (browserIsNotFocused || needsCustomEventOption) {
    let options = { relatedTarget };
    fireEvent(element, 'blur', assign({ bubbles: false }, options));
    fireEvent(element, 'focusout', options);
  }
```

Sorry, it wouldn't let me make a suggestion for this.

addon-test-support/@ember/test-helpers/dom/blur.ts
```diff
@@ -24,14 +29,17 @@ export function __blur__(element: HTMLElement | Element | Document | SVGElement)
 
   // makes `document.activeElement` be `body`.
   // If the browser is focused, it also fires a blur event
```
rwjblue, line 31: I'd like to keep this comment with the `element.blur()` call.

rwjblue (changes requested):

addon-test-support/@ember/test-helpers/dom/blur.ts
drewlee: @rwjblue this is intentional. The idea is that if the browser is in focus, and `__blur__` is triggered from `__focus__`, we shouldn't need to do anything. Based on my findings, the browser automatically fires native blur when `element.focus()` is called with the expected event properties (including `relatedTarget`). Otherwise, blur and focusout end up executing multiple times, which I believe is also one of the causes of the issue. I think the initial tests were masking this behavior as the condition never executed due to `activeElement` being the body. Let me set something up to better demonstrated what I mean, at least for my own sanity. It could be that the check to trigger `__blur__` needs to be modified, or I'm somehow getting trolled by the associated tests.

rwjblue: OK, that generally makes sense. We need to change this around so that the `focus` side deals with that nuanced detail because this code in `__blur__` should make the assumption that it is called from `focus`.

I think from the `__blur__` perspective, we should do the snippet I mentioned in a separate comment, then update `focus` to handle the "don't call __blur__ if you don't need to" case.

addon-test-support/@ember/test-helpers/dom/focus.ts
```diff
@@ -28,7 +28,7 @@ export function __focus__(element: HTMLElement | Element | Document | SVGElement
     document.activeElement !== element &&
     isFocusable(document.activeElement)
   ) {
-    __blur__(document.activeElement);
+    __blur__(document.activeElement, element);
   }
```
rwjblue, line 32: Based on https://github.com/emberjs/ember-test-helpers/pull/965/files#r544655601. This would become:

```js
if (
  document.activeElement &&
  document.activeElement !== element &&
  isFocusable(document.activeElement) &&
  browserIsNotFocused
) {
  __blur__(document.activeElement, element);
}
```

addon-test-support/@ember/test-helpers/dom/blur.ts
drewlee: Makes sense. I've incorporated this in my latest commit.

addon-test-support/@ember/test-helpers/dom/blur.ts
drewlee: Done.

addon-test-support/@ember/test-helpers/dom/focus.ts
drewlee: Updated.

addon-test-support/@ember/test-helpers/dom/blur.ts
drewlee: I've incorporated this snippet. Only minor change is that we don't need the extra ` || needsCustomEventOption` check, as `fireEvent` should only execute when browser is not focused.

drewlee:
I've put up a quick demo comparing the sequencing and `relatedTarget` property in terms of firing focus/blur and friends, which helps spotlight the crux of the issue.

Demo page https://codepen.io/drewlee/full/bGwqeap and the associated code https://codepen.io/drewlee/pen/bGwqeap.

As I've mentioned previously, calling `event.blur()` is unnecessary in sequence with `event.focus()`, as the browser automatically fires it with the expected event properties (scenario 1). Explicitly calling `event.blur()` disrupts the browser's default event sequence and causes the associated `relatedTarget` properties to default to `null` (scenario 2). In terms of using `element.dispatchEvent`, it requires explicitly setting the `relatedTarget` properties (scenarios 3 and 4). And obviously, we should use either one (blur/focus) or the other (dispatchEvent) depending on whether the browser is focused or not, otherwise events are duplicated. Given this information, the suggested fix is very much inline.

drewlee:
I ran the fix through several large apps and didn't encounter any issues or unexpected test failures.

addon-test-support/@ember/test-helpers/dom/blur.ts
```diff
@@ -14,24 +15,32 @@ registerHook('blur', 'start', (target: Target) => {
...
+  if (!needsCustomEventOptions) {
+    // makes `document.activeElement` be `body`.
+    // If the browser is focused, it also fires a blur event
+    element.blur();
+  }
 
   // Chrome/Firefox does not trigger the `blur` event if the window
   // does not have focus. If the document does not have focus then
   // fire `blur` event via native event.
   if (browserIsNotFocused) {
```
rwjblue, line 40: This needs to be:

```suggestion
  if (browserIsNotFocused || needsCustomEventOptions) {
```

ro0gr:
🚀

PR #987, opened 2021-02-11
Support getOwner

ef4:
The test context already has an owner, in the sense of `this.owner` working. But the standard way to get an owner in Ember is `getOwner`, which doesn't work.

This makes `getOwner` work.

A motivating example are [these tests](https://github.com/embroider-build/embroider/blob/0349de7a5386c30dff8d4aa859fe01138b026878/packages/util/tests/integration/helpers/ensure-safe-component-test.js#L18-L22) which currently use a workaround. They're testing an API that accepts any owned object, and it's nice to use the test context as that object.


rwjblue (approved):
Seems good to me, thank you!

PR #989, opened 2021-02-12
Add @embroider/test-setup to test against embroider

thoov:
Adds 2 test cases against embroider for ember-try.


stefanpenner:
I've rebased this, the failing tests are due to the ember-cli-typescript used in this project, which as implemented is not compatible with embroider. 

I intend to address next.

package.json
```diff
@@ -125,7 +128,7 @@
     }
   },
   "volta": {
-    "node": "10.20.1",
+    "node": "12.22.4",
```
stefanpenner, line 131: !!breaking change!!

stefanpenner (changes requested):

.github/workflows/ci-build.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: volta-cli/action@v1
         with:
-          node-version: 10.x
+          node-version: 12.x
```
stefanpenner, line 36: !!breasking change!!

package.json
```diff
@@ -41,24 +41,25 @@
     "broccoli-funnel": "^3.0.8",
     "ember-cli-babel": "^7.26.6",
     "ember-cli-htmlbars": "^5.7.1",
-    "ember-destroyable-polyfill": "^2.0.3"
+    "ember-destroyable-polyfill": "^2.0.3",
+    "ember-cli-typescript": "^4.2.1"
```
stefanpenner, line 45: Suboptimal, but at-least demonstrates embroider + ember-cli-typescript 4.2.3 work together

package.json
```diff
@@ -41,24 +41,25 @@
...
   },
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
+    "@embroider/test-setup": "^0.43.0",
```
stefanpenner, line 49: The primary reason for node 10.x drop

package.json
```diff
@@ -41,24 +41,25 @@
...
     "broccoli-merge-trees": "^4.2.0",
     "documentation": "^13.2.5",
-    "ember-auto-import": "^1.10.1",
+    "ember-auto-import": "^2.0.2",
     "ember-cli": "~3.27.0",
     "ember-cli-dependency-checker": "^3.2.0",
     "ember-cli-inject-live-reload": "^2.1.0",
     "ember-cli-shims": "^1.2.0",
     "ember-cli-test-loader": "^3.0.0",
-    "ember-cli-typescript": "^1.5.0",
```
stefanpenner, line 61: Moving to new versions of ember-cli-typescript continues the need to ship typescript, but provides embroider support.
There is another path, where we stop shipping typescript and rather do typescript as a pre-build step. Although I am not apposed to this, It's not clear its needed at this time.

package.json
```diff
@@ -81,10 +82,11 @@
     "qunit": "^2.16.0",
     "release-it": "^14.7.0",
     "release-it-lerna-changelog": "^3.1.0",
-    "typescript": "^4.3.5"
+    "typescript": "^4.3.5",
+    "webpack": "^5.49.0"
   },
   "engines": {
-    "node": "10.* || 12.* || >= 14.*"
+    "node": "12.* || 14.* || >= 16.*"
```
stefanpenner, line 89: !!breasking change!! due primarily to `@embroider/test-setup`

package.json
stefanpenner: Move to ember-try scenario

package.json
```diff
@@ -4,8 +4,7 @@
   "description": "Helpers for testing Ember.js applications",
   "keywords": [
     "ember-addon"
-  ],
-  "homepage": "https://github.com/emberjs/ember-test-helpers#readme",
+  ], "homepage": "https://github.com/emberjs/ember-test-helpers#readme",
```
rwjblue, line 7: this change should probably be reset

package.json
```diff
@@ -23,8 +22,10 @@
     "test": "tests"
   },
   "scripts": {
+    "prepare": "tsc -b ./tsconfig.json",
+    "clean": "git clean -x -f",
```
rwjblue, line 26: Just to confirm, this does **not** delete `node_modules` right?

package.json
```diff
@@ -81,10 +83,11 @@
     "qunit": "^2.16.0",
     "release-it": "~14.10.1",
     "release-it-lerna-changelog": "^3.1.0",
-    "typescript": "^4.3.5"
+    "typescript": "^4.3.5",
+    "webpack": "^5.0.0"
```
rwjblue, line 87: Why do we have webpack here?

package.json
```diff
@@ -58,15 +59,16 @@
     "ember-cli-inject-live-reload": "^2.1.0",
     "ember-cli-shims": "^1.2.0",
     "ember-cli-test-loader": "^3.0.0",
-    "ember-cli-typescript": "^1.5.0",
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-fetch": "^8.1.0",
     "ember-in-element-polyfill": "^1.0.1",
     "ember-load-initializers": "^2.1.2",
     "ember-maybe-import-regenerator-for-testing": "^1.0.0",
+    "ember-page-title": "^6.2.2",
```
rwjblue, line 67: Do we need this? Was it just brought in via an `ember init`?

package.json
```diff
@@ -58,15 +59,16 @@
...
     "ember-resolver": "^8.0.2",
     "ember-source": "~3.24.1",
     "ember-source-channel-url": "^3.0.0",
+    "ember-welcome-page": "^4.0.0",
```
rwjblue, line 71: ✂️

tsconfig.json
```diff
@@ -14,37 +14,35 @@
     "noUnusedParameters": false,
     "noImplicitReturns": true,
     "noEmitOnError": false,
-    "noEmit": true,
+    "declaration": true,
     "inlineSourceMap": true,
     "inlineSources": true,
-    "baseUrl": ".",
-    "module": "es6",
+    "module": "esnext",
```
rwjblue, line 20: why is this `esnext`? I would have thought we'd use either `ES2020` or `ES2015`... 🤔 

https://www.typescriptlang.org/tsconfig#module

rwjblue (dismissed):
Can you confirm:

* `rm API.md && yarn docs` generates the current `API.md` (e.g. has no diff)
* running a server works (file changes are detected, and cause a reload, tests re-run, etc)

package.json
stefanpenner: @rwjblue ember-page-title is used in the dummy app, embroider detected the missing dependency. Alternatively, I could remove the usage & the dependency now that we know it works.

package.json
stefanpenner: https://github.com/emberjs/ember-test-helpers/pull/989#discussion_r685551212

package.json
stefanpenner: Confirmed it does not

package.json
stefanpenner: i've removed this

stefanpenner:
A small bonus:

before:

```sh
> rm API.md && time yarn docs
Done in 22.35s.
________________________________________________________
Executed in   23.28 secs   fish           external
   usr time   23.74 secs   59.00 micros   23.74 secs
   sys time    2.56 secs  671.00 micros    2.56 secs
```

after:

```sh
> rm API.md && time yarn docs
Done in 1.37s.

________________________________________________________
Executed in    1.54 secs   fish           external
   usr time  1385.11 millis   68.00 micros  1385.04 millis
   sys time  209.11 millis  760.00 micros  208.35 millis
```

index.js
```diff
@@ -7,33 +7,16 @@ module.exports = {
...
-          'addon-test-support/@ember/test-helpers': '@ember/test-helpers',
-        }),
-        'addon-test-support:ts'
-      );
-
-      input = debugTree(
-        new MergeTrees([input, tsTree]),
-        'addon-test-support:merged'
-      );
+    if (this.isDevelopingAddon()) {
```
stefanpenner, line 12: @rwjblue this simplifies development, quickly removing the need to run a separate "TS build" process.
I wouldn't consider this a final product, but demonstrates the viability  "TS in dev + precompiled on publish approach"

When you have time, we should discuss

stefanpenner:
GH actions appears to be queued up... Will poke back later...

.github/workflows/ci-build.yml
scalvert: Is there a reason we're coupling this change with this PR? Do dependent packages drop support for Node 10? Can we not do this separately? As @stefanpenner mentioned, this is a breaking change, and our package.jsons' engines support still specifies Node 10.

scalvert: I found this further down the PR https://github.com/emberjs/ember-test-helpers/pull/989/files#diff-e835654135548f75db36136807171b4d12d3514e043f4f80a146b88cc668fba4R32.

I still think we need a better strategy for removing this. We could even define a new scenario for CI that uses node 10 only for embroider tests.

index.js
scalvert: I'd like to understand this more too. I'm not entirely sure what you mean when you refer to "TS in dev + precompiled on publish approach", because that describes what's done today - we use TS in dev, and precompile for releases.

package.json
```diff
@@ -84,7 +86,7 @@
     "typescript": "^4.3.5"
   },
   "engines": {
-    "node": "10.* || 12.* || >= 14.*"
+    "node": "10.* || 12.* || 14.* || 15.* || >= 16.*"
```
scalvert, line 89: From what I can tell, this isn't necessary. If you test these ranges on https://semver.npmjs.com/ they do exactly the same thing...

package.json
scalvert: Understood, but this should likely match a change to our engines support, and CI (reflected in my comment on the change to the ci-build.yml above).

index.js
stefanpenner: the comment was more based on an intermediate step + discussion with rob. Functionally this appears the same, but no longer requires ember-cli-typescript

package.json
scalvert: So we'd have to update the engines then anyway, right? Why not just leave it until then?

index.js
scalvert: As yes, I remember this. This was the removal of the `incremental-typescript-compiler` from 1.5.0 -> 2.0.0. Makes sense.

stefanpenner (approved):

PR #992, opened 2021-02-17
Inline promise polyfill fallback.

rwjblue:
The UMD wrapping that was used by es6-promise was causing issues in certain contexts (see https://github.com/embroider-build/embroider/issues/677).

This moves from es6-promise to the same promise polyfill that is used by qunit@2.14.0 and higher (promise-polyfill@8.2.0 with some tweaks).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -70,5 +70,12 @@ module.exports = {
         'require-jsdoc': 'error',
       },
     },
+    {
+      files: ['addon-test-support/@ember/test-helpers/-internal/promise-polyfill.js'],
+      rules: {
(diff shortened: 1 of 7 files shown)


scalvert:
Oh boy, this is the gift that keeps on giving...

scalvert:
I there a way we can add tests to ensure this works as expected?

rwjblue:
> I there a way we can add tests to ensure this works as expected?

I tested in a local IE11 VM. There are the same number of failures with this PR as there are on master (the failures are around event ordering with things like `typeIn`, and are not related to async/promises).

PR #994, opened 2021-02-23
More robust handling in setup-onerror

### LINKED ISSUE #993: `onerror` should not be cached on load
Instead, it should be cached the first time `setupOnerror` is called.

See https://github.com/emberjs/ember-qunit/issues/818.

wagenet:
Resolves an issue where Ember.onerror could get cached too soon.

Fixes #993


rwjblue (approved):
Thanks @wagenet!

Issue #1032, opened 2021-05-24
Blur event is not triggered on the current active element in the click helper


helgablazhkun:
**Scenario:**
Expected result: 
**GIVEN** There are two elements on the form: focusable input and label (simple div).  
**AND** Input element has focus
**WHEN** Click on the label occurs (via click test helper)
**THEN** Input gets blur event and loses its focus

Actual result: 
**Input doesn't get blur event and doesn't  lose its focus**

The root cause of the issue is that the blur event is not triggered on the current active element in the click helper in case if the clicked element is not focusable.
If the clicked element is focusable the previous active element gets blur event trough the __focus__ helper. 

**Proposed solution:** Add a call to __blur__  for the case when the clicked element is not focusable.

**Real world example:** The issue is noticed in the integration test for dropdown component with Ember Power Select in the real application. When click occurs on the dropdown option (non-focusable element) then input doesn't lose its focus in the test but it does in the real application. So Click test helper has different behaviour with browser click.

rwjblue:
Can you create a plain HTML demo showing the real world scenario you described (e.g. codepen with just a few lines of JS wiring up a blur event)?

I want to make sure that we properly emulate the events across all of our supported browsers.

helgablazhkun:
Please use this example: https://codepen.io/helgablazhkun/pen/qBrjbzd
Events are logged into the console.

helgablazhkun:
Closed 
https://github.com/emberjs/ember-test-helpers/pull/1033

PR #1033, opened 2021-05-24
Ensure `blur` event fires on prior focused element even if click target is not focusable

helgablazhkun:
The root cause of the issue is that the blur event is not triggered on
the current active element in the click helper in case if the clicked
element is not focusable.

If the clicked element is focusable the previous active element gets
blur event trough the __focus__ helper.

A call to __blur__ was added for the case when the clicked element
is not focusable.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/focus.ts
+++ b/addon-test-support/@ember/test-helpers/dom/focus.ts
@@ -17,21 +17,29 @@ registerHook('focus', 'start', (target: Target) => {
   @param {Element} element the element to trigger events on
 */
 export function __focus__(element: HTMLElement | Element | Document | SVGElement): void {
+  const previousFocusedElement =
+    document.activeElement &&
+    document.activeElement !== element &&
+    isFocusable(document.activeElement)
+      ? document.activeElement
+      : null;
+
+  // fire __blur__ manually with the null relatedTarget when the target is not focusable
+  // and there was a previously focused element
   if (!isFocusable(element)) {
-    throw new Error(`${element} is not focusable`);
+    if (previousFocusedElement) {
+      __blur__(previousFocusedElement, null);
+    }
+
+    return;
   }
 
   let browserIsNotFocused = document.hasFocus && !document.hasFocus();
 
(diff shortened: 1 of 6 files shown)


helgablazhkun:
Fixes #1032

rwjblue (commented):
If this is generally the correct behavior (hopefully the plain HTML demo will prove that to us), we should consider doing this inside __focus__ itself. I suspect other of our interaction helpers would need this same thing (e.g. `doubleClick`, `tap`, etc).

addon-test-support/@ember/test-helpers/dom/focus.ts
```diff
@@ -18,7 +18,15 @@ registerHook('focus', 'start', (target: Target) => {
...
-    throw new Error(`${element} is not focusable`);
+    if (
+      document.activeElement &&
+      document.activeElement !== element &&
+      isFocusable(document.activeElement)
+    ) {
+      __blur__(document.activeElement, null);
+    }
+
+    return;
```
rwjblue, line 29: This seems like we are basically duplicating the code just below (line 36), but after a closer inspection I can see that it is quite different (e.g. we are only doing it if the target is not focusable, and we do it regardless of `document.hasFocus()`).

Can you add an inline comment to explain why this exists?

rwjblue (commented):
Overall seems great, thanks for working on this!

addon-test-support/@ember/test-helpers/dom/focus.ts
helgablazhkun: Done

rwjblue (approved):
Great job, thank you!!

bendemboski:
Hey @rwjblue and @helgablazhkun, this is a great improvement, and definitely moves things in the right direction! However, it has turned out to be a pretty significantly breaking change for my test suite -- see #1126 and #1127. I guess it's already released, but maybe we can fast-track those PRs to minimize the time that the latest version of `@ember/test-helpers` has breaking changes relative to v2.2.5?

PR #1066, opened 2021-07-19
Correctly import deprecate from `@ember/debug`

mydea:
This was incorrectly imported from `@ember/application/deprecations`, which seemed to work, but does not work in current canary - I got this error in an addon testing on canary:

```js
Global error: Uncaught Error: Could not find module `@ember/application/deprecations` 
imported from `@ember/test-helpers/setup-rendering-context`
```

Anyway, even if this might/should be fixed in canary, the import is wrong and can be updated to the correct place.

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
+++ b/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
@@ -9,7 +9,7 @@ import { hbs, TemplateFactory } from 'ember-cli-htmlbars';
 import getRootElement from './dom/get-root-element';
 import { Owner } from './build-owner';
 import getTestMetadata, { ITestMetadata } from './test-metadata';
-import { deprecate } from '@ember/application/deprecations';
+import { deprecate } from '@ember/debug';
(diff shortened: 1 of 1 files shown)


mydea:
Canary still fails because of:
https://github.com/emberjs/ember-test-helpers/pull/1067

simonihmig:
Nice, thanks for the quick fix here @mydea :+1:

I am also seeing basically any addon test suite failing on canary. Strangely the GH Actions job is not failing immediately, but timing out after running >5h! 🙈. 
[Same here in this repo](https://github.com/emberjs/ember-test-helpers/runs/3100279355?check_suite_focus=true)

So for both addon maintainers and Github container utilization getting this landed would be very helpful! 😆

NullVoxPopuli:
I just experienced the same thing, but I have my timeout set much lower :)

sandydoo:
Thanks, @mydea. Linking #1068.

rwjblue:
Thanks!

mydea:
Any chance you could cut a release with this change? Would be great to get canary tests passing again for addons :)

PR #1067, opened 2021-07-19
Test `{{in-element}}` instead of deprecated `{{-in-element}}`

mydea:
This leads to failing tests on Canary, where `{{-in-element}}` is removed.

This updates the test to test the `{{in-element}}` helper on ember versions upwards of 3.20. I did not include an alternate branch to test `{{-in-element}}` on older versions because I figured that is kind of repetitive and we can assume they work the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "ember-data": "~3.20.0",
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-fetch": "^8.1.0",
+    "ember-in-element-polyfill": "^1.0.1",
     "ember-load-initializers": "^2.1.2",
     "ember-maybe-import-regenerator-for-testing": "^1.0.0",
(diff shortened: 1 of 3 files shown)


mydea:
Canary tests will fail because of: 
https://github.com/emberjs/ember-test-helpers/pull/1066

simonihmig:
> Canary tests will fail because of:

Actually 3.8 is failing. Probably because the error is thrown at build-time by the AST transform, at least that's what I would assume. Maybe adding https://github.com/ember-polyfills/ember-in-element-polyfill as a devDependency instead of the Ember version check would help?

mydea:
Hmm, interesting! I changed the test to just always run and use `{{in-element}}` and installed `ember-in-element-polyfill` as dev dependency. Should hopefully work now :sweat_smile:

rwjblue (approved):

PR #1095, opened 2021-08-05
Workaround classic failure

stefanpenner:
- workaround classic build failure so our tests aren't constantly red. I wasn't able (in the time alloted) track down the issue, but it __apears__ to be a bug/quirk in ember's classic mode. If i had more time I would totally attempt to resolve that, but unfortunately I must stay a little more focused.
- add a quick DX improvement by always redirecting the browser from `/` -> `/tests` when loading the dev harness

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "dependencies": {
     "@ember/test-waiters": "^2.4.5",
     "broccoli-debug": "^0.6.5",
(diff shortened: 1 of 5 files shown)


stefanpenner:
I'm planning to follow this up (when green) with something that removes the defunct `_APPLICATION_TEMPLATE_WRAPPER` code, which appears to be a remnant of another long forgotten era.

PR #1096, opened 2021-08-05
Partially application of `ember init` to “modernize” code-base

stefanpenner:
These should be very safe changes, I will follow up with the ones that require more testing/verification

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,64 @@
+---
+language: node_js
+node_js:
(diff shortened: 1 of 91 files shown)


.travis.yml
```diff
@@ -0,0 +1,64 @@
+---
```
rwjblue, line 1: @stefanpenner - Should be deleted

stefanpenner: https://github.com/emberjs/ember-test-helpers/pull/1098

PR #1104, opened 2021-08-16
Fixes tarball to include correct contents

stefanpenner:


package.json
```diff
@@ -23,7 +23,8 @@
     "test": "tests"
   },
   "scripts": {
-    "prepublishOnly": "yarn babel --extensions '.ts' --presets @babel/preset-typescript addon-test-support --out-dir addon-test-support/ --ignore '**/*.d.ts'",
```
stefanpenner, line 26: turns out this works fine, but sucks if you try to test via `npm pack`

package.json
```diff
@@ -83,6 +84,7 @@
     "qunit": "^2.16.0",
     "release-it": "~14.10.1",
     "release-it-lerna-changelog": "^3.1.0",
+    "rimraf": "^3.0.2",
```
stefanpenner, line 87: cross platform cleanup on postpack

.npmignore
```diff
@@ -35,6 +35,17 @@
 /bower.json.ember-try
 /package.json.ember-try
 
+# other
```
stefanpenner, line 38: cleanup tarball while I'm at it.

scalvert (approved):

PR #1108, opened 2021-08-17
[Idea] no longer vendor the promise-polyfill rather use auto-import

stefanpenner:
Assuming we enforce auto-import on users in some future version

@rwjblue / @scalvert WDYT?


chriskrycho:
@rwjblue given we're about to cut a 3.0.0, thoughts on landing this? The only reason I can think of not to—but/and it's a significant one—is that this package will still support Ember 3.28, which itself still supports 🔥 DIAF IE11 🔥.

PR #1109, opened 2021-08-18
Upgrade @ember/test-waiters to 3.0.0

stefanpenner:
- `@emberjs/test-waiters`
- add ember-source 3.24 to ember-try

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -53,6 +53,7 @@ jobs:
           - ember-lts-3.12
           - ember-lts-3.16
           - ember-lts-3.20
+          - ember-lts-3.24
           - ember-release
           - ember-beta
           - ember-canary
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -84,6 +84,14 @@ module.exports = async function () {
         },
       },
       {
+        name: 'ember-lts-3.24',
+        npm: {
+          devDependencies: {
+            'ember-source': '~3.24.0',
+          },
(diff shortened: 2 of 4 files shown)


scalvert (approved):

scalvert (approved):
Thanks for adding the 3.24 scenario!

Turbo87:
@stefanpenner this appears to be a breaking change (see https://github.com/rust-lang/crates.io/pull/3840). we are using ember-cli-dependency-lint to avoid having multiple versions of the same addon in the app, and now we have two different major versions of `@ember/test-waiters` that can't be deduplicated.

scalvert:
Hey @Turbo87, `@ember/test-waiters` has a mechanism that [dedupes that dependency at build time](https://github.com/emberjs/ember-test-waiters/blob/master/force-highlander-addon.js), which helps ensure you only have one version at any time in your project. Is this manifesting as an actual issue, or are you just anticipating this based on how current ember-cli build dedupes work?

stefanpenner:
@Turbo87 yes, it is a major version...

scalvert:
I think you can accommodate this dependency using:

```json
// config/dependency-lint.js
module.exports = {
  allowedVersions: {
    '@ember/test-waiters': '*'
  }
};
```

Turbo87:
@scalvert TIL, thank you. I've opened https://github.com/salsify/ember-cli-dependency-lint/pull/22 to adjust the default list of `allowedVersions` in that linter.

PR #1111, opened 2021-08-23
Fixes postpack script to cleanup correct file type

scalvert:
Fixes bug in `postpack` that incorrectly deletes `.ts` files rather than `.js`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "prepack": "yarn babel --extensions '.ts' --presets @babel/preset-typescript addon-test-support --out-dir addon-test-support/ --ignore '**/*.d.ts'",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "prepack": "yarn babel --extensions '.ts' --presets @babel/preset-typescript addon-test-support --out-dir addon-test-support/ --ignore '**/*.d.ts'",
-    "postpack": "rimraf addon-test-support/**/*.ts",
+    "postpack": "rimraf addon-test-support/**/*.js",
```
stefanpenner, line 27: good catch, i got this backwards, thank you!

stefanpenner (approved):

PR #1113, opened 2021-08-24
Tab support

NullVoxPopuli:
Continuation of: https://github.com/emberjs/ember-test-helpers/pull/771/files

I was going to try to "quick do this", but it seems that since the #771 PR was created, enough API has changed, where I need to take a deeper look. Standby.


Turbo87:
does this respect the fact that on macOS Chrome only links are tab enabled by default, and links won't receive focus?

Turbo87:
> I am not sure what you mean? can you describe that behavior?

on macOS Chrome links are not part of the tab order. in other words: when you press <kbd>Tab</kbd> it won't focus links.

> and/or how it's different from other browsers?

Firefox on macOS does focus links. Chrome on other platforms AFAIK also focusses links.

NullVoxPopuli:
I can focus links via tab
image


but anyway, running the tests in Chrome on Mac still results in a successful test suite.

Turbo87:
OMG they finally fixed it 😂 

forget everything I said... 🙈

MelSumner:
@rwjblue I love this, can we have this please? ^_^

MelSumner (commented):
I don't think I can officially approve this but 100% I think it's great!

rwjblue:
I'll try to review in detail this week.

@NullVoxPopuli - Can you use the `Co-authored-by:` commit header to add attribution for @mogstad's hard work over in #771?

NullVoxPopuli:
sure thing! did I do it right? I've never done a Co-authored by thing

mogstad:
Thanks for picking this up NullVoxPopuli, and bringing it over the finish line. 

Re `Co-authored` I usually just includes it at the bottom of the commit body and not in the title.

> does this respect the fact that on macOS Chrome only links are tab enabled by default, and links won't receive focus?

This is the default behaviour in Safari on macOS, and there is a setting to change it.. I haven’t found a way of detecting this setting. The code doesn’t try to accommodate for Safari’s behaviour. Detecting Safari and filter out links with a default `tabindex` should be rather straightforward.

NullVoxPopuli:
Ok, I've fixed the attribution (thanks again, @mogstad !!!)
And I've rebased. 

Lemme know if anything needs to be done for this batch of work :)

PR #1114, opened 2021-08-25
[task]: remove deprecated Ember.assign

snewcomer:
RFC https://deprecations.emberjs.com/v4.x#toc_ember-polyfills-deprecate-assign


addon-test-support/@ember/test-helpers/-internal/debug-info.ts
```diff
@@ -95,7 +94,7 @@ export class TestDebugInfo implements DebugInfo {
 
   get summary(): SummaryInfo {
     if (!this._summaryInfo) {
-      this._summaryInfo = assign(<SummaryInfo>{}, this._settledState);
+      this._summaryInfo = Object.assign(<SummaryInfo>{}, this._settledState);
```
rwjblue, line 97: I wonder, can we just use `...` style syntax now?

```js
      this._summaryInfo: SummaryInfo = { ...this._settledState };
```

I think we can?

PR #1125, opened 2021-09-08
Prevent leaking application instances when jQuery integration is enabled

bendemboski:
The AJAX listener cleanup code wasn't always calling this._super(), so with jQuery integration enabled, lots of application instances were leaking


rwjblue (commented):
Thank you!

PR #1126, opened 2021-09-08
Preserve focus when default is prevented on mousedown

bendemboski:
Preventing default on a mousedown event prevents the focus from changing, so implement that behavior in the click and dblclick DOM helpers

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/click.ts
+++ b/addon-test-support/@ember/test-helpers/dom/click.ts
@@ -34,9 +34,9 @@ export function __click__(
   element: Element | Document | Window,
   options: MouseEventInit
 ): void {
-  fireEvent(element, 'mousedown', options);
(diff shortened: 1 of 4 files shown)


bendemboski:
This fixes one of the cases that #1033 broke.

My use case is toolbar buttons that perform content operations on the focused contenteditable. The toolbar buttons might do things like enable bold formatting or insert an image, and when clicking them the selection needs to be preserved, so I call preventDefault() on the `mousedown` event on those buttons to preserve focus. Since the buttons themselves are un-focusable, before #1033, clicking them would do the right thing for the wrong reason (preserve focus because we weren't blurring the active element when clicking an un-focusable element). After #1033 it started doing the wrong thing, but with this change it will do the right thing for the right reason!

rwjblue (approved):

PR #1127, opened 2021-09-09
Fix focus behavior when ancestor is focusable

bendemboski:
When clicking/double-clicking on an un-focusable element with a focusable ancestor, the __focus__ helper was not noticing that there was a focusable ancestor, and blurring the active element without focusing the focusable ancestor (or leaving focus alone of the focusable ancestor was already focused).

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/focus.ts
+++ b/addon-test-support/@ember/test-helpers/dom/focus.ts
@@ -3,7 +3,7 @@ import fireEvent from './fire-event';
 import settled from '../settled';
 import isFocusable from './-is-focusable';
 import { Promise } from '../-utils';
-import Target from './-target';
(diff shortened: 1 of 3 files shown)


bendemboski:
This fixes one of the cases that #1033 broke.

My use case is clicking on an un-focusable element inside a focused contenteditable to perform actions using a different selection model (CKEditor's, which is different from the browser's focus model). Before #1033, doing so would do the right thing for the wrong reason (keep the contenteditable focused because we weren't blurring the active element when clicking an un-focusable element). After #1033 it started doing the wrong thing (blurring the contenteditable), but with this change it will do the right thing for the right reason, as well as supporting other focus scenarios such as clicking on spans inside of anchors, etc.

rwjblue (approved):
Thank you for working on this!

PR #1139, opened 2021-10-11
Add drag and drop support

JimSchofield:
I had to write this for one of my projects, and thought it would be helpful to include.

Tries to support the dataTransfer part of the drag, as that seems to be the most important thing with drag events.

I did not model drag events as a MouseEvent because keeping it a basic event is the simplest, however they are truly mouse events so it may be worth putting it under that umbrella in the code in the future.


JimSchofield:
@MelSumner wondering if there's time to review this?

MelSumner (approved):
This looks good to me but I'd like to have @rwjblue review it for consistency. Thank you for doing this work! 👍

locks:
@JimSchofield hi! do you have some time to resolve the conflicts?

JimSchofield:
I spent some time today trying to rebase of of master here.  I think I fixed the conflicts, but we're getting behavior changes and the tests are failing.  I'm noticing that options being passed into `fireEvent` are missing when the event is called. Specifically, when I mock the DataTransfer object to allow setting and getting values this object seems to disappear after the event is fired.  Perhaps this has to do with making fireEvent return a promise?  But I have to visit this another time.

addon/addon-test-support/@ember/test-helpers/dom/drag-move.ts
```diff
@@ -48,28 +48,32 @@ export default function dragMove(
     .then(() =>
       runHooks('dragMove', 'start', draggableStartElement, dropTargetElement)
     )
-    .then(() => {
+    .then(async () => {
```
JimSchofield, line 51: This is a naive try to get it to work as it was before the `fireEvent` change.

NullVoxPopuli:
hello! sorry for not looking at this PR sooner -- would you be willing to rebase?

PR #1140, opened 2021-10-12
Add a `triggerEvent` doc example

yannbertrand:
My team and I have been struggling with the `triggerEvent` options parameter.

Our usecase is fixing [a Firefox bug](https://stackoverflow.com/questions/46831247/select-triggers-mouseleave-event-on-parent-element-in-mozilla-firefox) where the mouseleave event is triggered when hovering a select option even if the mouse is still hover the target element. We did some acceptance tests to trigger the event with or without a given `relatedTarget` element.

Hopefully this PR will help other users too 😄.


rwjblue (approved):
Thank you!

rwjblue:
Looks like some linting failures here (sorry for taking so long to approve the run) @yannbertrand

yannbertrand:
> Looks like some linting failures here (sorry for taking so long to approve the run) @yannbertrand

Should be fixed now. Thank you for your patience :)

PR #1156, opened 2021-10-27
Include event.detail in click and dblclick events

andrewtimberlake:
This change updates UIEvent.detail (https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail) on `click` and `dblclick` events.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/click.ts
+++ b/addon-test-support/@ember/test-helpers/dom/click.ts
@@ -33,6 +33,7 @@ export function __click__(
   element: Element | Document | Window,
   options: MouseEventInit
 ): void {
+  options = { ...options, detail: 1 };
   let mouseDownEvent = fireEvent(element, 'mousedown', options);
 
   if (!isWindow(element) && !mouseDownEvent?.defaultPrevented) {
--- a/addon-test-support/@ember/test-helpers/dom/double-click.ts
+++ b/addon-test-support/@ember/test-helpers/dom/double-click.ts
@@ -22,6 +22,7 @@ export function __doubleClick__(
   element: Element | Document | Window,
   options: MouseEventInit
 ): void {
+  options = { ...options, detail: 1 };
   let mouseDownEvent = fireEvent(element, 'mousedown', options);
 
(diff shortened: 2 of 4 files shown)


kategengler:
From the documentation linked:

> For click or dblclick events, UIEvent.detail is the current click count.

From some quick experimentation, successive clicks on a button with a click handler yielded an `event.detail` of 1,2,3,4,5,6... 

Knowing that, I would expect that we wouldn't always want `UIEvent.detail` on a click to be 1, as you could have successive `click` events in a tests. 

> For mousedown or mouseup events, UIEvent.detail is 1 plus the current click count.

I haven't tested this, but it sounds like the `detail` option that is passed along to mousedown and mouseup would need to increment.

addon-test-support/@ember/test-helpers/dom/click.ts
```diff
@@ -33,6 +33,7 @@ export function __click__(
   element: Element | Document | Window,
   options: MouseEventInit
 ): void {
+  options = { ...options, detail: 1 };
```
kategengler, line 36: From https://github.com/emberjs/ember-test-helpers/blob/master/API.md#click

I believe this would be a breaking change as users could no longer override `detail` on the event.

andrewtimberlake: That is a good point.

andrewtimberlake:
> > For mousedown or mouseup events, UIEvent.detail is 1 plus the current click count.
> 
> I haven't tested this, but it sounds like the `detail` option that is passed along to mousedown and mouseup would need to increment.

The spec here sounds confusing but the increment is happening. Because mouse down and mouse up trigger before click they start at 0 (click) + 1, then 1 click + 1  etc.

addon-test-support/@ember/test-helpers/dom/click.ts
rwjblue: ya, you'd have to do it the other way around (so the users provided `detail` option "wins"):

```suggestion
  options = { detail: 1, ...options};
```

PR #1168, opened 2021-11-19
Add example docs for `getDebugInfo`

rwjblue:


scalvert (approved):

addon-test-support/@ember/test-helpers/-internal/debug-info.ts
```diff
@@ -57,8 +57,20 @@ export function backburnerDebugInfoAvailable() {
 
 /**
  * Retrieves debug information from backburner's current deferred actions queue (runloop instance).
+ * The debug info returned includes details about anything that is currently unsettled (e.g. timers,
+ * pending test waiters, etc).
+ *
  * If the `getDebugInfo` method isn't available, it returns `null`.
  *
+ * @example
+ *
```
scalvert, line 66:
```suggestion
 *
```

scalvert (approved):

addon-test-support/@ember/test-helpers/-internal/debug-info.ts
```diff
@@ -57,8 +57,20 @@ export function backburnerDebugInfoAvailable() {
...
+ * ```js
+ * import { getDebugInfo } from '@ember/test-helpers';
+ *
+ * // ...snip...
+ * getDebugInfo().toConsole();
```
scalvert, line 71: I don't think this is quite right. Calling `getDebugInfo` returns the inner debug info from backburner's [`getDebugInfo`](https://github.com/BackburnerJS/backburner.js/blob/master/lib/index.ts#L576). It's the [`TestDebugInfo`](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/-internal/debug-info.ts#L82) class that has the [`toConsole`](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/-internal/debug-info.ts#L148) method on it, which is what [`getSettledState`](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/@ember/test-helpers/settled.ts#L236) returns in its result.

scalvert (changes requested):

MelSumner (changes requested):
Also needs to be updated to reflect the current placement of the file.

PR #1169, opened 2021-11-20
Fix CI tests

snewcomer:
Trying to get these tests passing. Fixes for ember-release and beyond

https://github.com/emberjs/ember.js/pull/19728

### DIFF
--- a/tests/unit/setup-context-test.js
+++ b/tests/unit/setup-context-test.js
@@ -849,10 +849,7 @@ module('setupContext', function (hooks) {
 
     hooks.beforeEach(function () {
       const AppConfig = assign({ autoboot: false }, config.APP);
-      // .extend() because initializers are stored in the constructor, and we
-      // don't want to pollute other tests using an application created from the
-      // same constructor.
-      isolatedApp = App.extend({}).create(AppConfig);
+      isolatedApp = App.create(AppConfig);
       let resolver = isolatedApp.Resolver.create({
         namespace: isolatedApp,
         isResolverFromTestHelpers: true,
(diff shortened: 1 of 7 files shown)


tests/unit/setup-context-test.js
```diff
@@ -849,10 +849,7 @@ module('setupContext', function (hooks) {
 
     hooks.beforeEach(function () {
       const AppConfig = assign({ autoboot: false }, config.APP);
-      // .extend() because initializers are stored in the constructor, and we
-      // don't want to pollute other tests using an application created from the
-      // same constructor.
-      isolatedApp = App.extend({}).create(AppConfig);
```
rwjblue, line 855: The thing this comment is describing is absolutely an issue (and why you had to add that `delete` in the after each), we can't share the same parent as `App` here. 

Can you bring back the comment, and assuming the goal was to get rid the manual `.extend()` call we can change to:

```js
isolatedApp = (class extends App {}).create(AppConfig);
```

This does the same thing as the `.extend` did and doesn't require the `delete` you added to the `afterEach`

tests/unit/setup-context-test.js
```diff
@@ -862,6 +859,11 @@ module('setupContext', function (hooks) {
       setApplication(isolatedApp);
     });
 
+    hooks.afterEach(function () {
+      delete isolatedApp.constructor.initializers.foo;
+      isolatedApp.destroy();
```
rwjblue, line 864: we should probably await settled after this

tests/unit/setup-context-test.js
```diff
@@ -862,6 +859,11 @@ module('setupContext', function (hooks) {
       setApplication(isolatedApp);
     });
 
+    hooks.afterEach(function () {
+      delete isolatedApp.constructor.initializers.foo;
```
rwjblue, line 863: let's remove this (in favor of adding the `class extends App {}` thing I mentioned in the other comment)

rwjblue (changes requested):

tests/unit/setup-context-test.js
snewcomer: Thanks and addressed!

rwjblue (approved):

PR #1047, opened 2021-06-14
Ensure jQuery.ajax pending request detection works with Ember 3.27 under Embroider

stefanpenner:
This both tidies up the code, but also ensures embroider compatibility for those users with ember ember-source@3.27.x but still have jQuery.

PR #1173, opened 2021-12-02
Fix the way the pending requests module is used in settled.ts

### LINKED ISSUE #1172: EmberApplicationInstance willDestroy in settled.ts uses the wrong `internalPendingRequests`
upon updating ember-test-helpers, I have test code that resets the app hit the app instance `willDestroy` method, which errors on `internalPendingRequests.clearPendingRequests()` because `internalPendingRequests` is the (nonzero) integer result of calling the `pendingRequests` method.

Here's where the pendingRequests method is chained on the require that prevents accessing the other methods:
https://github.com/emberjs/ember-test-helpers/blob/9c6e884d8572e7c262e4915d88dd61df787deda9/addon-test-support/%40ember/test-helpers/settled.ts#L30-L31

error occurs here:
https://github.com/emberjs/ember-test-helpers/blob/9c6e884d8572e7c262e4915d88dd61df787deda9/addon-test-support/%40ember/test-helpers/settled.ts#L59

likewise the `off` calls are passing `undefined` since `internalPendingRequests` is an integer and not a module

If you look at the deleted code in https://github.com/emberjs/ember-test-helpers/pull/1047 it was not chaining `pendingRequests` on the old `require('ember-testing/lib/test/pending_requests')` which allowed accessing the different methods from that module.

Is a PR welcome here?

raycohen:
Fixes #1172 
Related to #1047

Before these changes, the code is confusing a module with one of the functions on that module.

### DIFF
--- a/addon-test-support/@ember/test-helpers/settled.ts
+++ b/addon-test-support/@ember/test-helpers/settled.ts
@@ -18,23 +18,28 @@ import DebugInfo, { TestDebugInfo } from './-internal/debug-info';
 //
 // This utilizes a local utility method present in Ember since around 2.8.0 to
 // properly consider pending AJAX requests done within legacy acceptance tests.
-const _internalPendingRequests = (() => {
+const _internalPendingRequestsModule = (() => {
   let loader = (Ember as any).__loader;
 
   if (loader.registry['ember-testing/test/pending_requests']) {
     // Ember <= 3.1
-    return loader.require('ember-testing/test/pending_requests')
-      .pendingRequests;
+    return loader.require('ember-testing/test/pending_requests');
   } else if (loader.registry['ember-testing/lib/test/pending_requests']) {
     // Ember >= 3.2
-    return loader.require('ember-testing/lib/test/pending_requests')
-      .pendingRequests;
+    return loader.require('ember-testing/lib/test/pending_requests');
   }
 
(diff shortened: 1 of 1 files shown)


addon-test-support/@ember/test-helpers/settled.ts
```diff
@@ -44,20 +49,16 @@ if (typeof jQuery !== 'undefined' && _internalPendingRequests) {
   // This can be removed once Ember 4.0.0 is released
   EmberApplicationInstance.reopen({
     willDestroy(...args: any[]) {
-      const internalPendingRequests = _internalPendingRequests();
-
-      if (internalPendingRequests !== 0) {
```
raycohen, line 49: the outer if statement checks that `_internalPendingRequestsModule` exists, which I believe makes calling `off` and `clearPendingRequests` safe and desirable regardless of what the PendingRequestsCount is

rwjblue (approved):
Thanks @raycohen!

ginomiglio:
Would it be possible to cut a release that includes this change?

raycohen:
@rwjblue any chance we can get a release with this 🙏

rwjblue:
oh dear, sorry y'all; I thought I had already released it 😭

rwjblue:
Finally released in 2.6.2

PR #1179, opened 2022-01-03
Move `ember-cli-htmlbars` to `devDependencies`

SergeAstapov:
As addon does not ship any hbs templates, there is no need to have `ember-cli-htmlbars` in `dependencies`


SergeAstapov:
This is super minor change, but allows to save few ms at build time.

SergeAstapov:
scratch it! There is `import { hbs, TemplateFactory } from 'ember-cli-htmlbars';` import in `addon-test-support/@ember/test-helpers/setup-rendering-context.ts` which I somehow missed 🤦

PR #1183, opened 2022-01-12
feat: add `'before-fire-event'` hook to helpers with side-effects

shamrt:
A `'before-fire-event'` hook is added to any helper that involves side-effects — i.e. scrolls or changes values in the DOM prior to firing a `change` event. These hooks allow better control of the DOM in cases where code needs to be executed _after_ a helper affects the DOM as well as _before_ an event is fired.

This PR follows from my conversation in #1182, and should serve to resolve an issue I have following @rwjblue's suggestion therein.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/fill-in.ts
+++ b/addon-test-support/@ember/test-helpers/dom/fill-in.ts
@@ -72,6 +72,13 @@ export default function fillIn(target: Target, text: string): Promise<void> {
           '`fillIn` is only usable on form controls or contenteditable elements.'
         );
       }
+
+      return settled();
+    })
+    .then(() => runHooks('fillIn', 'before-fire-event', target, text))
+    .then(() => {
+      let element = getElement(target) as Element | HTMLElement;
+
       fireEvent(element, 'input');
       fireEvent(element, 'change');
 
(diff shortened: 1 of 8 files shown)


addon-test-support/@ember/test-helpers/dom/fill-in.ts
```diff
@@ -72,6 +72,13 @@ export default function fillIn(target: Target, text: string): Promise<void> {
           '`fillIn` is only usable on form controls or contenteditable elements.'
         );
       }
+
+      return settled();
+    })
+    .then(() => runHooks('fillIn', 'before-fire-event', target, text))
```
rwjblue, line 78: I was thinking that you would actually add this to `fireEvent` itself, that way you only haev to change one file (not all of hte dom helpers). 

Is there any reason that wouldn't work?

shamrt: So I'm now deep into this thing, and thought I should check in about my progress and confirm that the direction I'm heading is kosher. 

Specifically, given the hooks system uses Promises, I've been forced to "Promisify" any function where `fireEvent` is called. Given the library forbids the use of `async`/`await`, in places, the changes are fairly substantial and it's not easy to keep things readable. It'd doable, though.

My assumption about `async`/`await` is that we forbid it for backwards compatibility with older apps. Is that correct?

Anyway, I'm about 60–80% of the way done now.

shamrt:
Closing this PR in favour of #1185

Issue #1128, opened 2021-09-15
setupOnError not working


amk221:
[Here](https://github.com/amk221/-ember-setup-on-error/commit/24bb42d2ba3d68bd0381cd639385e43a4bf1102d) are two test cases.

One where a component errors in constructor, and another where it errors in response to a click.

The first can be caught by `setupOnError`, and ignored in the test suite. 👍🏻
The second is not caught by `setupOnError`, and so the test fails. 👎🏻

jfdnc:
Interestingly, this causes your failing test here to respect `setupOnerror` and pass:

```javascript
<button type="button" {{on "click" (action this.check)}}>
  check
</button>
```

_with the example invocation binding via `@action` window object onerror called_
![Screen Recording 2021-10-01 at 11 05 54 PM](https://user-images.githubusercontent.com/15672873/135704333-2317347b-f8c7-45b8-b233-42a1eeb3340e.gif)

_using `(action` helper invokes error in the correct context_
![Screen Recording 2021-10-01 at 11 06 27 PM](https://user-images.githubusercontent.com/15672873/135704343-04fbe050-bef0-4b86-bad7-848060dd5de7.gif)

jfdnc:
@amk221 Looks like this is a known issue, there is some discussion in the latter part of [this thread](https://github.com/emberjs/ember-qunit/issues/592), which it looks like you were part of but there has been more discussion since your last comment there. Seems to be currently unresolved.

buschtoens:
I got fed up with this long-standing issue, so I decided to invest a bit of time and do some detective work. 🕵️😄

This occurs, when an error is thrown outside of the Ember run loop. `setupOnerror` works by hooking into `Ember.onerror`, which in turn [is called by Backburner](https://github.com/emberjs/ember.js/blob/v4.4.0-alpha.3/packages/%40ember/runloop/index.ts#L87-L94) (Ember's backing run loop implementation) when a task in the run loop throws.

The error then propagates to the global [`window.onerror`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) handler or [`window.onunhandledrejection`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunhandledrejection), if it's async.

[...]

Issue #440, opened 2018-10-21
Return the event which is triggered by a DOM event helper


bastimeyer:
Currently, none of the available DOM event helpers (`triggerEvent`, `triggerKeyEvent`, `click`, etc.) return the event which they have triggered on an element and instead return the value of the `isSettled()` call. This means that in cases where for example a component has prevented the browser's default action or has stopped the propagation of a certain event, this can't be tested by using the available DOM helpers of ember-test-helpers.

What is even the reason for returning the value of `isSettled()`? By looking at the code, it seems like the intention for this was to just chain the `settled()` promise in the returned event helper promise.

See this simple use case:
```js
test( "foo", async function( assert ) {
  await render( hbs`{{foo-bar}}` );
  const event = await triggerEvent( ".foo-bar", "eventname" );
  assert.ok( event.defaultPrevented, "Prevents default action" );
});
```

I've been previously using jQuery for having a dead simple API to trigger DOM events in tests, but since Ember is slowly getting rid of jQuery (which is a good thing), the logical choice for devs is to stop using jQuery in tests, too, and using the helper methods of ember-test-helpers instead. But since the test helpers don't return the events which they are triggering, this is a problem in certain test scenarios.

cibernox:
@bastimeyer Right now every helper returns the value of calling `settled()`, which unless I'm wrong it is just a `Promise<void>`. What you are suggesting is that we change it so it returns `Promise<Event>`.

In ok with the idea in principle, as right now afaik we don't have any compromise of what the resolution value of the promises are.

However it can be tricky for events like `click`  or `fillIn` that in reality trigger several events.
Click, for instance, fires `mousedown`, `focus`, `mouseup` and then `click`. 
We could just return click or return an object or array with all the fired events, but we'd have to decide what is the right thing to do first.

PR #1194, opened 2022-02-24
[BREAKING?] fix(settled, waitUntil): reject with error thrown in run loop

### LINKED ISSUE #310: render does not reject when rendering fails
I would have expected this test to pass, but it fails with "expected not to get here" in addition to an out-of-band qunit failure for the actual `Compiler Error: not-a-component is not a helper`.

```js
test('render can throw', async function(assert) {
  try {
    await render(hbs`{{not-a-component x="y"}}`);
  } catch (err) {
    assert.ok(/not-a-component is not a helper/.test(err.message), "expected to see failure");
  }
  throw new Error("expected not to get here");
});

```

### LINKED ISSUE #453: Only jQuery.click() triggers error caught by assert.throws
This [twiddle reproduces](https://ember-twiddle.com/de9115316563715af87fec8ddec9f796?openFiles=tests.integration.components.my-c-test.js%2Ctemplates.components.my-c.hbs) the issue.

### TLDR
There is a component that will throw an error if two attrs arent passed in (`x` and `y`) and the component is clicked on.

There are three tests:
1. Ensure no errors if both attrs passed in and component is clicked ✅
2. Ensure an error is throw when only one attr is passed in and component is clicked (**uses native JS and tests fails**) ❌
3. Ensure an error is throw when only one attr is passed in and component is clicked (uses jquery and test passes) ✅

Here is where things go wrong: 
```
// Qunit throws
try {
   block.call(currentTest.testEnvironment);  <====== This call doesnt trigger an error when it uses native JS. But FUN FACT: if you highlight this line at a breakpoint, it will log an error (???)
} catch (e) {
   actual = e;
}
...
if (actual) {     <==== `actual` needs to be set otherwise it wont catch the error and the test will fail.
```

_Originally posted by @nlfurniss in https://github.com/emberjs/ember-test-helpers/issues/310#issuecomment-412316982_

buschtoens:
This PR is currently still WIP and failing some tests, but I'd like to receive feedback on the implementation and course-correct, where necessary.

---

The goal of this PR is to fix the longstanding issue of test helpers not propagating run loop errors thrown during their execution, but instead causing an out of band "global failure" that fails the test. For example:

```ts
await render(hbs`<UnknownComponent />`);
// or
await click('.element-throws-on-click');
```

Fixes #310. Fixes #453. Relates to #440, #1128, [ember-qunit#592](https://github.com/emberjs/ember-qunit/issues/592)

---

This PR adds an `rejectOnError` option to `waitUntil`. It defaults to `true`, if the test context is set up correctly for use with `setupOnerror`.

If enabled, `waitUntil` will reject when an error propagates to `Ember.onerror`, using `setupOnerror` under the hood. As such, (a/sync) errors thrown in a run loop will cause `waitUntil` to fail.

Almost all test helpers end with a call to `settled`, which just calls `waitUntil`. Therefore they'll all automatically benefit from this change and reject, if an error is thrown during their execution.

### DIFF
--- a/addon-test-support/@ember/test-helpers/wait-until.ts
+++ b/addon-test-support/@ember/test-helpers/wait-until.ts
@@ -1,4 +1,8 @@
 import { futureTick, Promise } from './-utils';
+import setupOnerror, {
+  resetOnerror,
+  _getContextForOnError,
+} from './setup-onerror';
 
 const TIMEOUTS = [0, 1, 2, 5, 7];
 const MAX_TIMEOUT = 10;
@@ -8,6 +12,17 @@ type Falsy = false | 0 | '' | null | undefined;
 export interface Options {
   timeout?: number;
   timeoutMessage?: string;
+
+  /**
+   * Instrument `Ember.onerror` and reject, when it is called. This is useful
+   * for detecting that an operation inside a run loop has failed.
+   *
+   * This uses {@link setupOnerror}, so it will override any error listeners you
+   * might have set up before.
+   *
+   * @default true if the test context has been setup for usage with {@link setupOnerror}
+   */
+  rejectOnError?: boolean;
 }
 
 /**
@@ -21,6 +36,7 @@ export interface Options {
   @param {Object} [options] options used to override defaults
   @param {number} [options.timeout=1000] the maximum amount of time to wait
   @param {string} [options.timeoutMessage='waitUntil timed out'] the message to use in the reject on timeout
+  @param {boolean} [options.rejectOnError] reject when an operation in a run loop has failed; defaults to `true`, if the test context has been setup for usage with {@link setupOnerror}
   @returns {Promise} resolves with the callback value when it returns a truthy value
 
   @example
@@ -33,19 +49,24 @@ export interface Options {
 */
 export default function waitUntil<T>(
   callback: () => T | void | Falsy,
-  options: Options = {}
+  {
+    timeout = 1000,
+    timeoutMessage = 'waitUntil timed out',
+    rejectOnError = Boolean(_getContextForOnError(false)),
+  }: Options = {}
 ): Promise<T> {
-  let timeout = 'timeout' in options ? (options.timeout as number) : 1000;
-  let timeoutMessage =
-    'timeoutMessage' in options
-      ? options.timeoutMessage
-      : 'waitUntil timed out';
-
   // creating this error eagerly so it has the proper invocation stack
   let waitUntilTimedOut = new Error(timeoutMessage);
 
   return new Promise(function (resolve, reject) {
     let time = 0;
+    let error: unknown = undefined;
+
+    if (rejectOnError) {
+      setupOnerror((e) => {
+        error = e;
+      });
+    }
 
     // eslint-disable-next-line require-jsdoc
     function scheduleCheck(timeoutsIndex: number) {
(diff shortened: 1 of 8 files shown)


addon-test-support/@ember/test-helpers/wait-until.ts
```diff
@@ -8,6 +12,17 @@ type Falsy = false | 0 | '' | null | undefined;
...
+  /**
+   * Instrument `Ember.onerror` and reject, when it is called. This is useful
+   * for detecting that an operation inside a run loop has failed.
+   *
+   * This uses {@link setupOnerror}, so it will override any error listeners you
+   * might have set up before.
+   *
+   * @default true if the test context has been setup for usage with {@link setupOnerror}
+   */
+  rejectOnError?: boolean;
```
buschtoens, line 25: I'm not really fond of this, because:

> it will override any error listeners you might have set up before.

> if the test context has been setup for usage with `setupOnerror`

This limitation arises from how `setupOnerror` works internally. Alternatively, we could _always_ instrument `Ember.onerror` in tests, somewhat like:

```ts
const originalOnError = Ember.onerror;
let onErrorOverride?: typeof Ember.onerror;

Ember.onerror = error => {
  // _Always_ let `waitUntil` know about the error.
  notifyInternalOnError(error);

  // If the user called `setupOnerror(...)`, call the override.
  if (onErrorOverride)
    return onErrorOverride(error);
  
  // Otherwise call the original `Ember.onerror`.
  return onErrorOverride?.(error)
};

export function setupOnerror(onError?: typeof Ember.onerror): void {
  let context = getContext();

  if (!context) {
    throw new Error('Must setup test context before calling setupOnerror');
  }

  if (!cachedOnerror.has(context)) {
    throw new Error(
      '_cacheOriginalOnerror must be called before setupOnerror. Normally, this will happen as part of your test harness.'
    );
  }

  if (typeof onError !== 'function') {
    onError = cachedOnerror.get(context);
  }

  onErrorOverride = onError;
}

export function resetOnerror(): void {
  let context = getContext();
  if (context && cachedOnerror.has(context)) {
    onErrorOverride = cachedOnerror.get(context);
  }
}
```

buschtoens: Also, letting `rejectOnError` default to `true` will inevitably break a lot of existing test suites, that currently try to emulate this missing feature by using `setupOnerror` or manually hooking into `Ember.onerror`.

However, I believe that ultimately the behavior implemented in this PR is what a user would expect. So while this breaking change may be a pain, because it will require changing all tests dealing with such erros, I think that it's a _very_ welcome change.

_Maybe_ we could even conjur up a codemod to ease the transition. Or we could add a compatibility mode, that will essentially change the order of `notifyInternalOnError` and `onErrorOverride`, so that "old" tests just continue to work.

```ts
const originalOnError = Ember.onerror;
let onErrorOverride?: typeof Ember.onerror;
let preferOnErrorOverride = true;

Ember.onerror = error => {
  // Compatibility mode for old tests.
  if (preferOnErrorOverride && onErrorOverride)
    return onErrorOverride(error);

  // _Always_ let `waitUntil` know about the error.
  notifyInternalOnError(error);

  // If the user called `setupOnerror(...)`, call the override.
  if (onErrorOverride)
    return onErrorOverride(error);
  
  // Otherwise call the original `Ember.onerror`.
  return onErrorOverride?.(error)
};
```

We should also consider adding a setter trap for `Ember.onerror` to catch users manually overriding it in tests.

ef4:
This seems good. I think it's the best solution that doesn't require changes to ember itself.

But I view this as a workaround for a feature ember itself should offer. There should be public API that lets you implement `render`, and it should include more specific error handling than a single global callback.

NullVoxPopuli:
Should we pursue finishing / deciding what to do with this PR?

Issue #596, opened 2019-03-11
Windows 10 issue with building docs


ygongdev:
For some reason, sometimes running `yarn run docs` would throw these weird errors and fail to construct the doc. It seems like it has to do with `TypeScript`, perhaps from `ember-cli-typescript`. I was able to confirm that this was only happening on my windows 10 machine, but not my Mac.

```
D:\Coding\ember-test-helpers>yarn run docs
yarn run v1.13.0
$ scripts/precompile.sh && documentation build --document-exported "addon-test-support/@ember/test-helpers/index.js" --config documentation.yml -f md -o API.md && ember ts:clean
Done in 12.28s.

D:\Coding\ember-test-helpers>yarn run docs
yarn run v1.13.0
$ scripts/precompile.sh && documentation build --document-exported "addon-test-support/@ember/test-helpers/index.js" --config documentation.yml -f md -o API.md && ember ts:clean
Table of contents defined sorting of blur but no documentation with that namepath was found
Table of contents defined sorting of click but no documentation with that namepath was found
Table of contents defined sorting of doubleClick but no documentation with that namepath was found
Table of contents defined sorting of fillIn but no documentation with that namepath was found
Table of contents defined sorting of focus but no documentation with that namepath was found
Table of contents defined sorting of tap but no documentation with that namepath was found
Table of contents defined sorting of triggerEvent but no documentation with that namepath was found
Table of contents defined sorting of triggerKeyEvent but no documentation with that namepath was found
Table of contents defined sorting of typeIn but no documentation with that namepath was found
Table of contents defined sorting of find but no documentation with that namepath was found
Table of contents defined sorting of findAll but no documentation with that namepath was found
Table of contents defined sorting of getRootElement but no documentation with that namepath was found
Table of contents defined sorting of visit but no documentation with that namepath was found
Table of contents defined sorting of currentRouteName but no documentation with that namepath was found
Table of contents defined sorting of currentURL but no documentation with that namepath was found
Table of contents defined sorting of render but no documentation with that namepath was found
Table of contents defined sorting of clearRender but no documentation with that namepath was found
Table of contents defined sorting of waitFor but no documentation with that namepath was found
Table of contents defined sorting of waitUntil but no documentation with that namepath was found
Table of contents defined sorting of settled but no documentation with that namepath was found
Table of contents defined sorting of isSettled but no documentation with that namepath was found
Table of contents defined sorting of getSettledState but no documentation with that namepath was found
[...]

PR #1196, opened 2022-02-27
feat(build): publish types as part of `@ember/test-helpers`

buschtoens:
**Build and publish the TypeScript `.d.ts` files as part of `@ember/test-helpers` to obsolete `@types/ember__test-helpers`.**

- https://github.com/emberjs/rfcs/pull/800
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59014
- #596 _fixed by_
  https://github.com/documentationjs/documentation/pull/1484


package.json
```diff
@@ -61,7 +61,7 @@
     "@typescript-eslint/parser": "^4.29.0",
     "broccoli-babel-preset-typescript": "^1.0.1",
     "broccoli-merge-trees": "^4.2.0",
-    "documentation": "^13.2.5",
+    "documentation": "npm:@buschtoens/documentation@13.2.6",
```
buschtoens, line 64: We may want to consider taking this opportunity to switch to a more well-maintained tool, like [API Extractor](https://api-extractor.com/). That would also improve the `.d.ts` build process.

chriskrycho:
I just remembered this existed! 😓 @buschtoens if you can rebase and update, we'll get it published.

chriskrycho (changes requested):
Had a couple comments in flight. I'm sure there will be more.

buschtoens:
@chriskrycho Sure! I'll try to get to it in the next hours. Thanks for picking this up again. 🥳

package.json
buschtoens: @chriskrycho

- Should I keep pointing at my fork here?
- Or should we switch to a different tool like [API Extractor](https://api-extractor.com/) first? If so, ...
  - Which tool would prefer?
  - Should this be split out into an extra PR that is merged before this one?
- Or do you have a completely different solution to this in mind?

chriskrycho:
So sorry I let it sit for so long! Just fell off my radar, as these things do!

chriskrycho:
@buschtoens also relevant: #1234, which… may… supersede this? (I had obviously *really* forgotten about this PR. 😢)

package.json
chriskrycho: I think you're exactly right directionally! Let's see if we can get [TypeDoc](https://typedoc.org) wired up to do what we want. TypeDoc is almost certainly what we're going to end up using for Ember itself; it is how we are doing *all* of our TS-powered docs-generation at LinkedIn; and it is what I've used most successfully on other TS projects.

(API Extractor is (a) barely maintained at this point and (b) uninterested in supporting anything other than single-entry-point Node-style packages. TypeDoc is well-maintained, open to contributions, and actively working on supporting things like package `exports` maps. If we find gaps, let's help fix them there!)

chriskrycho:
Going to go ahead and close this as superseded by a number of other commits; thanks again for getting it started @buschtoens!

PR #1211, opened 2022-04-11
add `rerender` helper

cafreeman:


addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -115,9 +121,86 @@ export function render(
...
+            Object.prototype.hasOwnProperty.call(
+              templateOrComponent,
+              '__meta'
+            )) ||
+          (Object.prototype.hasOwnProperty.call(templateOrComponent, 'id') &&
+            Object.prototype.hasOwnProperty.call(templateOrComponent, 'meta'));
+
+        if (!isTemplate) {
+          throw new Error(
+            'Prior to Ember 3.24, you may only pass templates to `render`.'
```
rwjblue, line 138:
```suggestion
            `Using \`render\` with something other than a pre-compiled template is not supported until Ember 3.24 (you are on ${Ember.VERSION}).'
```

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -115,9 +121,86 @@ export function render(
...
+          }
+
+          if (context.get(CALLED_SET_PROPERTIES)) {
+            warn(
+              "Calling `this.setProperties` when rendering a component is an epic troll. Probably don't do it.",
+              {
+                id: 'set-properties-warning',
+              }
+            );
+          }
```
rwjblue, line 165: Let's tweak this together

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -115,9 +121,86 @@ export function render(
...
+            // helper so they can actually be found by the resolver and rendered
+            templateId += 1;
+            let name = `-undertest-${templateId}`;
+            let componentFullName = `component:${name}`;
+            let templateFullName = `template:components/${name}`;
+            context = {
+              ProvidedComponent: name,
+            };
+            ownerToRenderFrom.register(componentFullName, templateOrComponent);
+            templateOrComponent = hbs`{{component this.ProvidedComponent}}`;
```
rwjblue, line 191: Let's hoist this `{{component this.ProvidedComponent}}` template up (no need to create a new template for each test)

package.json
```diff
@@ -36,8 +36,12 @@
     "test": "ember test",
     "test:all": "ember try:each"
   },
+  "peerDependencies": {
+    "ember-source": "*"
+  },
```
rwjblue, line 41: FYI for other's coming along later, this is required in order to work properly with `dependencySatisifies` from `@embroider/macros`.

package.json
```diff
@@ -36,8 +36,12 @@
     "test": "ember test",
     "test:all": "ember try:each"
   },
+  "peerDependencies": {
+    "ember-source": "*"
```
rwjblue, line 40: I think we should probably actually include our "real" minimum version here, what do you think?

If we worry that this might be considered "breaking" to folks (I personally don't, but I could imagine someone using Ember 3.7 or something and things _happen_ to work), we can always make an issue to track updating this to an actual value at our next major release.

package.json
```diff
@@ -36,8 +36,12 @@
...
+  },
   "dependencies": {
     "@ember/test-waiters": "^3.0.0",
+    "@embroider/macros": "^1.5.0",
```
rwjblue, line 44: 1.6.0 is out, we should probably update to that

package.json
```diff
@@ -48,6 +52,9 @@
     "@babel/cli": "^7.14.8",
     "@babel/preset-typescript": "^7.15.0",
     "@ember/optional-features": "^2.0.0",
+    "@glimmer/component": "^1.0.4",
+    "@glimmer/interfaces": "^0.84.1",
+    "@glimmer/reference": "^0.84.1",
```
rwjblue, line 57: _someday_ we'll detangle that circular reference and then not need all these extra packages 😭

tests/integration/rerender-test.js
```diff
@@ -0,0 +1,147 @@
...
+import { module, test } from 'qunit';
+
+module('rerender real-world', function (hooks) {
+  hooks.beforeEach(async function () {
+    await setupContext(this);
+    await setupRenderingContext(this);
+  });
+  hooks.afterEach(async function () {
+    await teardownContext(this);
+  });
```
rwjblue, line 27: 🤔  do we also need to call `await teardownRenderingContext(this)`? I _think_ we do, but maybe I made that obsolete at some point?

rwjblue: nvm, looks like that was removed in v2.0.0 as one of the breakages 🎉

tests/integration/rerender-test.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    setComponentTemplate(
+      hbs`<div data-test-content>The profile for {{@name}} is {{this.value}}</div>`,
+      MyComponent
+    );
+
+    const somePerson = new (class {
+      @tracked name = 'Zoey';
+    })();
```
rwjblue, line 61:
```suggestion
    class Person {
      @tracked name = 'Zoey';
    }
    const somePerson = new Person();
```

tests/integration/rerender-test.js
```diff
@@ -0,0 +1,147 @@
...
+  teardownContext,
+  waitFor,
+} from '@ember/test-helpers';
+import { buildWaiter } from '@ember/test-waiters';
+import GlimmerComponent from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { hbs } from 'ember-cli-htmlbars';
+import { module, test } from 'qunit';
+
+module('rerender real-world', function (hooks) {
```
rwjblue, line 20: I think we need to guard this test with a short circuit return when on Ember < 3.25

```suggestion
module('rerender real-world', function (hooks) {
  // this test requires Ember 3.28 in order to use lexically scoped values in templates
  if (!hasEmberVersion(3, 28)) {
    return;
  }
```

tests/integration/rerender-test.js
```diff
@@ -0,0 +1,147 @@
...
+        },
+      }
+    );
+
+    const component = setComponentTemplate(template, templateOnly());
+
+    const renderPromise = render(component);
+
+    // No actual render
+    assert.equal(this.element.textContent, '', 'Nothing on the screen yet');
```
rwjblue, line 80:
```suggestion
    assert.equal(this.element.textContent, '', 'precond - Nothing on the screen yet');
```

tests/integration/rerender-test.js
```diff
@@ -0,0 +1,147 @@
...
+
+    // This lets us wait for rendering to actually occur without resolving the render promise itself.
+    // This way, we can make assertions about the initial state while also keeping the settled state
+    // suspended
+    await waitFor('[data-test-content]');
+    const element = find('[data-test-content]');
```
rwjblue, line 86: For context, this is basically what you would have had to do **before** `await rerender()` was created.

tests/unit/setup-rendering-context-test.js
```diff
@@ -555,6 +561,295 @@ module('setupRenderingContext', function (hooks) {
...
+          class Foo extends GlimmerComponent {
+            foo = 'foo';
+          }
+
+          const component = setComponentTemplate(template, Foo);
+          await render(component);
+
+          assert.equal(this.element.textContent, 'the value of foo is foo');
+
+          this.set('foo', 'bar');
```
rwjblue, line 657: We'll need to tweak this when we fixup the `this.set` assertion, but the test overall is 👌

tests/unit/setup-rendering-context-test.js
```diff
@@ -555,6 +561,295 @@ module('setupRenderingContext', function (hooks) {
...
+          await rerender();
+          assert.equal(this.element.textContent, 'the value of foo is foo');
+        });
+
+        test('setting properties on the test context when rendering a component triggers a warning', async function (assert) {
```
rwjblue, line 662: Let's tweak this to be:

* an assertion
* move the `set`'s down after `await render` (which I think is what most folks probably do)

addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -414,6 +418,11 @@ export default function setupContext(
         enumerable: true,
         value(key: string, value: any): any {
           let ret = run(function () {
+            if (ComponentRenderMap.has(context)) {
+              assert(
+                'Calling `this.set` when rendering a component does not work since components do not have access to the test context.'
```
rwjblue, line 423:
```suggestion
                'You cannot call `this.set` when you have passed a component to `render()` (the rendered component does not have access to the test context).'
```

addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -427,6 +436,11 @@ export default function setupContext(
         enumerable: true,
         value(hash: { [key: string]: any }): { [key: string]: any } {
           let ret = run(function () {
+            if (ComponentRenderMap.has(context)) {
+              assert(
+                'Calling `this.setProperties` when rendering a component does not work since components do not have access to the test context.'
```
rwjblue, line 441:
```suggestion
                'You cannot call `this.setProperties` when you have passed a component to `render()` (the rendered component does not have access to the test context)'
```

addon-test-support/@ember/test-helpers/-internal/is-component.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import getComponentManager from './get-component-manager';
+
+/**
+ *
+ * @private
+ * @param {Object} maybeComponent The thing you think might be a component
+ * @param {Object} owner Owner, we need this for old versions of getComponentManager
+ * @returns {boolean} True if it's a component, false if not
+ */
```
rwjblue, line 11: Can you update the docs here to mention that we should ultimately be getting an API from glimmer/runtime (and link the RFC)

addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -342,6 +342,10 @@ export function getWarningsDuringCallback(
   return getWarningsDuringCallbackForContext(context, callback);
 }
 
+// This WeakMap is used to track whenever a component is rendered in a test so that we can throw
+// assertions when someone uses `this.{set,setProperties}` while rendering a component.
+export const ComponentRenderMap = new WeakMap();
```
rwjblue, line 347:
```js
export const SetUsage = new WeakMap<Context, string[]>();
```

addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -414,6 +418,11 @@ export default function setupContext(
         enumerable: true,
         value(key: string, value: any): any {
           let ret = run(function () {
+            if (ComponentRenderMap.has(context)) {
+              assert(
+                'You cannot call `this.set` when you have passed a component to `render()` (the rendered component does not have access to the test context).'
+              );
+            }
```
rwjblue, line 425:
```js
            if (ComponentRenderMap.has(context)) {
              assert(
                'You cannot call `this.set` when you have passed a component to `render()` (the rendered component does not have access to the test context).'
              );
            } else {
              let setsUsed = SetUsage.get(context);
              if (setsUsed === undefined) {
                setsUsed = [];
                SetUsage.set(context, setsUsed);
              }

              setsUsed.push([key, new Error()]);
            }
```

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -115,9 +121,69 @@ export function render(
...
+        templateOrComponent = lookupTemplate(
+          ownerToRenderFrom,
+          templateFullName
+        );
+      } else {
+        if (isComponent(templateOrComponent, owner)) {
+          // We use this to track when `render` is used with a component so that we can throw an
+          // assertion if `this.{set,setProperty} is used in the same test
+          ComponentRenderMap.set(context, true);
+
```
rwjblue, line 149:
```suggestion
          ComponentRenderMap.set(context, true);

          let setCallsAlreadySeen = SetUsage.get(context);
          if (setCallsAlreadySeen !== undefined) {
            throw new Error(`You cannot ...., but you called `this.set` for the following properties ${setCallsAlreadySeen.join(', ')}`);
          }
```

rwjblue (changes requested):

addon-test-support/@ember/test-helpers/-internal/render-settled.ts
```diff
@@ -0,0 +1,31 @@
...
+import {
+  macroCondition,
+  importSync,
+  dependencySatisfies,
+} from '@embroider/macros';
+
+let renderSettled: () => Promise<void>;
+
+// TODO need to add the actual version `@ember/renderer` landed once we know it
+if (macroCondition(dependencySatisfies('ember-source', '>=4.9999999.0'))) {
```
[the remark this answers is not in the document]
rwjblue: Ok, now that the Ember change has landed for realz let's update this to 4.5.0-beta.1.

rwjblue (approved):

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -14,12 +14,18 @@ import { hbs, TemplateFactory } from 'ember-cli-htmlbars';
...
 import { runHooks } from './-internal/helper-hooks';
 import hasEmberVersion from './has-ember-version';
+import isComponent from './-internal/is-component';
+import { macroCondition, dependencySatisfies } from '@embroider/macros';
+import { ComponentRenderMap, SetUsage } from './setup-context';
+import type { ComponentInstance } from '@glimmer/interfaces';
 
 const OUTLET_TEMPLATE = hbs`{{outlet}}`;
 const EMPTY_TEMPLATE = hbs``;
+const INVOKE_PROVIDED_COMPONENT = hbs`<this.ProvidedComponent />`;
```
SergeAstapov, line 27: @chriskrycho here is this mystic `this.ProvidedComponent` that fails ember-optimized scenario

PR #1233, opened 2022-07-16
Add more keyCode mappings

CvX:
To paraphrase the `keyFromKeyCode` comment: This is not a comprehensive list, but it is better than before 😉


rwjblue (approved):

rwjblue:
Thank you!

PR #1234, opened 2022-07-21
Introduce public TypeScript support

chriskrycho:
In line with [Ember RFC #0800](https://rfcs.emberjs.com/id/0800-ts-adoption-plan), begin publishing types from this library.

- [x] add documentation and CI configuration
- [x] update to using the Ember v4 types to get many fixes
- [x] start with TS 4.7 as the minimum supported TS version
- [x] add type tests
- [x] eliminate all uses of `any`
- [x] fix all existing type errors

Once this is in a release, we can remove the `@types/ember__test-helpers` package from DefinitelyTyped!


chriskrycho:
And I’m apparently bad at rebasing; *this* part I’ll fix up tomorrow!

addon-test-support/@ember/test-helpers/dom/get-root-element.ts
```diff
@@ -9,14 +9,15 @@ import { isDocument, isElement } from './-target';
...
-  let owner = context && context.owner;
 
-  if (!owner) {
+  if (!context || !isTestContext(context) || !context.owner) {
     throw new Error(
       'Must setup rendering context before attempting to interact with elements.'
     );
   }
 
+  let owner = context.owner;
```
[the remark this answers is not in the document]
chriskrycho: That’s my style preference as well, but in this case I was minimizing changes and the line was already present (it’s just shifted down from above and the diff doesn’t make that clear)!

addon-test-support/@ember/test-helpers/-internal/build-registry.ts
```diff
@@ -31,9 +31,11 @@ function exposeRegistryMethodsWithoutDeprecations(container: any) {
   for (let i = 0, l = methods.length; i < l; i++) {
     let method = methods[i];
 
-    if (method in container) {
+    if (method && method in container) {
       container[method] = function (...args: unknown[]) {
-        return container._registry[method](...args);
+        // SAFETY: `method` is defined because we *just* checked that it is in
+        // the conditional wrapping this.
```
wagenet, line 37: Would this be fixed by changing it to `const method` above?

addon-test-support/@ember/test-helpers/setup-application-context.ts
```diff
@@ -87,17 +88,17 @@ export function setupRouterSettlednessTracking() {
   HAS_SETUP_ROUTER.set(context, true);
 
   let { owner } = context;
-  let router;
+  let router: Router;
   if (CAN_USE_ROUTER_EVENTS) {
-    router = owner.lookup('service:router');
+    router = owner.lookup('service:router') as Router;
```
wagenet, line 93: Is it a pain (or just not really necessary) to do an assertion here?

addon-test-support/@ember/test-helpers/setup-application-context.ts
```diff
@@ -87,17 +88,17 @@ export function setupRouterSettlednessTracking() {
...
 
     // track pending transitions via the public routeWillChange / routeDidChange APIs
     // routeWillChange can fire many times and is only useful to know when we have _started_
     // transitioning, we can then use routeDidChange to signal that the transition has settled
     router.on('routeWillChange', () => (routerTransitionsPending = true));
     router.on('routeDidChange', () => (routerTransitionsPending = false));
   } else {
-    router = owner.lookup('router:main');
+    router = owner.lookup('router:main') as Router;
```
wagenet, line 101: See above.

wagenet (commented):
Looks good overall, just had a couple questions.

addon-test-support/@ember/test-helpers/setup-application-context.ts
chriskrycho: That's a very good call. I'll update!

addon-test-support/@ember/test-helpers/-internal/build-registry.ts
chriskrycho: Good question. I'll check on it and see if I can rewrite in a way that makes it clear to TS—it throws away all control flow-based narrowing for objects like that once in methods because the behavior of the closure means it cannot guarantee. But the rewrite you suggested might work!

chriskrycho: Good catch; it does; I'll update!

addon-test-support/@ember/test-helpers/setup-application-context.ts
chriskrycho: Done.

addon-test-support/@ember/test-helpers/setup-application-context.ts
```diff
@@ -173,7 +183,7 @@ export function currentRouteName(): string {
 
   let router = context.owner.lookup('router:main');
 
-  return get(router, 'currentRouteName');
+  return get(router, 'currentRouteName') as string;
```
wagenet, line 186: Should this also be an assertion?

addon-test-support/@ember/test-helpers/setup-application-context.ts
```diff
@@ -193,9 +203,12 @@ export function currentURL(): string {
   let router = context.owner.lookup('router:main');
 
   if (HAS_CURRENT_URL_ON_ROUTER) {
-    return get(router, 'currentURL');
+    return get(router, 'currentURL') as string;
```
wagenet, line 206: Should this also be an assertion?

wagenet (changes requested):

addon-test-support/@ember/test-helpers/setup-application-context.ts
chriskrycho: Not unreasonable. Really, it's only here till we drop support for older versions, at which point we can just do `router.currentRouteName`. I'll update anyway. Also the assertions about the router broke things, though; which means I asserted the wrong thing apparently!

wagenet (approved):

wagenet:
I approved, but I guess that is pending tests passing.

chriskrycho:
[Ugh](https://github.com/emberjs/ember-test-helpers/runs/7472994996?check_suite_focus=true#step:5:46):

> Global error: Uncaught Error: Could not find module `@ember/routing/router-service` imported from `@ember/test-helpers/setup-application-context` at http://localhost:7357/assets/vendor.js, line 259

Net, I think we’re going to have to drop the import there and use the cast. I forgot that that module *intentionally* does not export a class, explicitly to avoid subclassing. Will update later.

wagenet (approved):

PR #1271, opened 2022-11-30
Upgrade ember v3.27.0...v4.9.2 (in development)

gitKrystan:
This will give us access to the [preview types](https://blog.emberjs.com/announcing-official-typescript-types-public-preview/), which will help with https://github.com/emberjs/ember.js/issues/20254

Read carefully because I guessed on a lot of things. 🤠

PR #1272, opened 2022-11-30
Use ember preview types

gitKrystan:
Depends on https://github.com/emberjs/ember-test-helpers/pull/1271
Should unblock fixing https://github.com/emberjs/ember.js/issues/20254


addon-test-support/@ember/test-helpers/-internal/build-registry.ts
```diff
@@ -58,9 +58,11 @@ const Owner = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixin, {
    * @see {@link https://github.com/emberjs/ember.js/blob/v4.5.0-alpha.5/packages/%40ember/engine/instance.ts#L152-L167}
    */
   unregister(fullName: string) {
+    // @ts-expect-error
```
gitKrystan, line 61: These types aren't public and there is a bunch of `EmberObject` and `as any` in this file, so I figured leaving `@ts-expect-error` was OK...

package.json
```diff
@@ -60,20 +60,7 @@
...
-    "@types/ember__debug": "~4.0.1",
-    "@types/ember__destroyable": "~4.0.0",
-    "@types/ember__engine": "~4.0.4",
-    "@types/ember__object": "~4.0.2",
-    "@types/ember__polyfills": "~4.0.0",
-    "@types/ember__routing": "~4.0.7",
-    "@types/ember__runloop": "~4.0.1",
-    "@types/ember__service": "~4.0.0",
-    "@types/ember__test": "~4.0.0",
+    "@types/jquery": "^3.5.14",
```
gitKrystan, line 63: This is needed because `@types/ember-testing-helpers` was removed.

tsconfig.json
```diff
@@ -9,5 +9,5 @@
       "*": ["./types/*"]
     }
   },
-  "include": ["./addon-test-support/**/*.ts", "tests/**/*.ts"]
+  "include": ["addon-test-support/**/*.ts", "tests/**/*.ts", "types/**/*.ts"]
```
gitKrystan, line 12: This was necessary for the types to see `types/dummy/index.d.ts`

type-tests/tsconfig.json
```diff
@@ -13,5 +13,6 @@
         "types/*"
       ]
     }
-  }
+  },
+  "include": ["api.ts", "../types/**/*"]
```
gitKrystan, line 17: This was necessary for the types to see `types/dummy/index.d.ts`

addon-test-support/@ember/test-helpers/-internal/build-registry.ts
chriskrycho: That's the best move, yep. We can flag it up for figuring out a non-private-API way of tackling in the future!

chriskrycho (approved):
Seems good! Thank you!

PR #1283, opened 2022-12-14
Fix types versions

gitKrystan:
Somehow the issue that resulted in us having to use this wonky "public-types/addon-test-support/@ember/test-helpers/*" path was resolved, so the "typesVersions" config is currently broken on master (and 2.9.0).

An issue with post-build cleanup led me astray here, so I fixed that as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
   "typesVersions": {
     "*": {
       "*": [
-        "public-types/*"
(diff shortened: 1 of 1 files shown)


gitKrystan:
Currently verifying that this is working. 🙃

chriskrycho (approved):

PR #1285, opened 2022-12-14
Export type for Target

gitKrystan:
See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember__test-helpers/index.d.ts#L29


chriskrycho:
This looks good, but needs a merge or rebase for the type tests because of the order other PRs have landed in!

PR #1286, opened 2022-12-14
Upgrade expect-type; fix issues it reveals; add RenderingTestContext

gitKrystan:


addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -271,8 +268,8 @@ export function clearRender(): Promise<void> {
     element).
 
   @public
-  @param {Object} context the context to setup for rendering
-  @returns {Promise<Object>} resolves with the context that was setup
+  @param {TestContext} context the context to setup for rendering
+  @returns {Promise<RenderingTestContext>} resolves with the context that was setup
```
gitKrystan, line 272: AFAICT no OSS code uses the value here. We could opt not to make it public.

chriskrycho: Seems good to export, actually: it moves us one step closer to not having everything depend on `this.<whatever>`: even `this.element` becomes unnecessary if you do this:

```js
import { setupRenderingTest } from 'ember-qunit'; // transitively is this
import { render } from '@ember/test-helpers';

module('cool', function (hooks) {
  let context = setupRenderingTest(hooks);

  test('neato', assert function (assert) {
    await render(<template>Hello!</template>);
    assert(context.element.innerText.includes("Hello!"));
  });
});
```

chriskrycho (approved):

chriskrycho:
This one will *also* need to have a merge or rebase!

PR #1287, opened 2022-12-15
Port some conveniences from @types/ember__test-helpers package

gitKrystan:
I compared the `@types/ember__test-helpers` package to the public API of this package to ensure we were:
1. exporting the same things (aside from some of the missing exports I put in other PRs)
2. not losing useful type info (e.g. the `find` overloads)

In the process I noticed a couple of minor issues w/ the types and fixed those as well.

LMK if you'd prefer I broke this up.

### DIFF
--- a/addon-test-support/@ember/test-helpers/dom/-get-element.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-get-element.ts
@@ -1,7 +1,13 @@
 import getRootElement from './get-root-element';
 import Target, { isDocument, isElement } from './-target';
 
-function getElement(target: string): Element | null;
+function getElement<K extends keyof HTMLElementTagNameMap>(
+  target: K
+): HTMLElementTagNameMap[K] | null;
+function getElement<K extends keyof SVGElementTagNameMap>(
+  target: K
+): SVGElementTagNameMap[K] | null;
+function getElement<E extends Element = Element>(target: string): E | null;
 function getElement(target: Element): Element;
 function getElement(target: Document): Document;
 function getElement(target: Window): Document;
--- a/addon-test-support/@ember/test-helpers/dom/-get-element.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-get-element.ts
@@ -1,13 +1,16 @@
 import getRootElement from './get-root-element';
 import Target, { isDocument, isElement } from './-target';
 
+function getElement<
+  K extends keyof (HTMLElementTagNameMap | SVGElementTagNameMap)
+>(target: K): (HTMLElementTagNameMap | SVGElementTagNameMap)[K] | null;
 function getElement<K extends keyof HTMLElementTagNameMap>(
   target: K
 ): HTMLElementTagNameMap[K] | null;
 function getElement<K extends keyof SVGElementTagNameMap>(
   target: K
 ): SVGElementTagNameMap[K] | null;
-function getElement<E extends Element = Element>(target: string): E | null;
+function getElement(target: string): Element | null;
 function getElement(target: Element): Element;
 function getElement(target: Document): Document;
 function getElement(target: Window): Document;
--- a/addon-test-support/@ember/test-helpers/dom/-get-element.ts
+++ b/addon-test-support/@ember/test-helpers/dom/-get-element.ts
@@ -3,7 +3,7 @@ import Target, { isDocument, isElement } from './-target';
 
 function getElement<
   K extends keyof (HTMLElementTagNameMap | SVGElementTagNameMap)
->(target: K): (HTMLElementTagNameMap | SVGElementTagNameMap)[K] | null;
+>(target: K): (HTMLElementTagNameMap[K] | SVGElementTagNameMap[K]) | null;
 function getElement<K extends keyof HTMLElementTagNameMap>(
   target: K
 ): HTMLElementTagNameMap[K] | null;
--- a/addon-test-support/@ember/test-helpers/dom/find-all.ts
+++ b/addon-test-support/@ember/test-helpers/dom/find-all.ts
@@ -5,6 +5,9 @@ import { toArray } from '../ie-11-polyfills';
 // would simply be defined as a tweaked re-export as `querySelector` is, but it
 // is non-trivial (to say the least!) to preserve overloads like this while also
 // changing the return type (from `NodeListOf` to `Array`).
+export default function findAll<
+  K extends keyof (HTMLElementTagNameMap | SVGElementTagNameMap)
+>(selector: K): Array<(HTMLElementTagNameMap | SVGElementTagNameMap)[K]>;
 export default function findAll<K extends keyof HTMLElementTagNameMap>(
   selector: K
 ): Array<HTMLElementTagNameMap[K]>;
--- a/addon-test-support/@ember/test-helpers/dom/find-all.ts
+++ b/addon-test-support/@ember/test-helpers/dom/find-all.ts
@@ -7,7 +7,7 @@ import { toArray } from '../ie-11-polyfills';
 // changing the return type (from `NodeListOf` to `Array`).
 export default function findAll<
   K extends keyof (HTMLElementTagNameMap | SVGElementTagNameMap)
->(selector: K): Array<(HTMLElementTagNameMap | SVGElementTagNameMap)[K]>;
+>(selector: K): Array<HTMLElementTagNameMap[K] | SVGElementTagNameMap[K]>;
 export default function findAll<K extends keyof HTMLElementTagNameMap>(
   selector: K
 ): Array<HTMLElementTagNameMap[K]>;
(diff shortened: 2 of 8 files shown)


addon-test-support/@ember/test-helpers/dom/-get-element.ts
```diff
@@ -1,7 +1,13 @@
...
 import Target, { isDocument, isElement } from './-target';
 
-function getElement(target: string): Element | null;
+function getElement<K extends keyof HTMLElementTagNameMap>(
+  target: K
+): HTMLElementTagNameMap[K] | null;
+function getElement<K extends keyof SVGElementTagNameMap>(
+  target: K
+): SVGElementTagNameMap[K] | null;
+function getElement<E extends Element = Element>(target: string): E | null;
```
chriskrycho, line 10: I recognize these came from the DT types, but… this is terrible and we should not perpetuate its terribleness.

```suggestion
export default function getElement(selector: string): Element | null;
```

This change prevents people from “casting” it implicitly by writing `getElement<HTMLDivElement>()`. Although that's *convenient*, especially in tests, it's also quite misleading.

addon-test-support/@ember/test-helpers/dom/find.ts
```diff
@@ -1,12 +1,22 @@
...
+// Derived from `querySelector` types.
+export default function find<K extends keyof HTMLElementTagNameMap>(
+  selector: K
+): HTMLElementTagNameMap[K] | null;
+export default function find<K extends keyof SVGElementTagNameMap>(
+  selector: K
+): SVGElementTagNameMap[K] | null;
+export default function find<E extends Element = Element>(
+  selector: string
+): E | null;
```
chriskrycho, line 12: Same here!

```suggestion
export default function find(selector: string): Element | null;
```

addon-test-support/@ember/test-helpers/dom/trigger-event.ts
```diff
@@ -57,7 +57,7 @@ registerHook('triggerEvent', 'start', (target: Target, eventType: string) => {
 export default function triggerEvent(
   target: Target,
   eventType: string,
-  options?: object
+  options?: Record<string, unknown>
```
chriskrycho, line 60: 1. 💙 
2. In the future, we should see if we can make this more specific, but it shouldn't block this change. Making it more accurate will "just" be a bug fix.

type-tests/api.ts
```diff
@@ -45,6 +46,11 @@ import {
   unsetContext,
   teardownContext,
   setupRenderingContext,
+  BaseContext,
+  TestContext,
+  RenderingTestContext,
+  TestMetadata,
+  DebugInfo as InternalDebugInfo,
```
chriskrycho, line 53: No harm done here, but for future reference: I try to keep things like these ordering changes separate from any *meaningful* changes—I just pull them into a separate commit, using `git add --patch` (though I *always* do that via a GUI because I find the text interface for it terrible)—so that it's clear what the meaningful change was. Here, for example, it took some extra reading and comparison to figure out that `RenderingTestContext` was the key change in this set.

chriskrycho (commented):
One tweak requested!

addon-test-support/@ember/test-helpers/dom/trigger-event.ts
gitKrystan: Yeah, I did a deep dive on what the types should be, and `options` is just passed around willy nilly to a bunch of functions that take `any`, so figuring out the actual type will be a Project™️. I figured `Record<string, unknown>` was at least a little better than `object` and more consistent with another `options` type you specified for a similar test helper.

gitKrystan: I could do something like `Event & Record<string, unknown>` bc I'm pretty sure it's mostly used for specifying details on the `Event` to be triggered, and I think that would at least give some hints about the types for things that exist on an `Event`.

type-tests/api.ts
gitKrystan: I'll just un-sort for now.

Re `RenderingTestContext`, should it be made public to consumers of this library? Currently it's not actually exported, so this import doesn't even work. I fix that in https://github.com/emberjs/ember-test-helpers/pull/1286, where the same issue came up for different reasons.

addon-test-support/@ember/test-helpers/dom/-get-element.ts
```diff
@@ -1,6 +1,15 @@
...
 
+function getElement<
+  K extends keyof (HTMLElementTagNameMap | SVGElementTagNameMap)
+>(target: K): (HTMLElementTagNameMap | SVGElementTagNameMap)[K] | null;
+function getElement<K extends keyof HTMLElementTagNameMap>(
+  target: K
+): HTMLElementTagNameMap[K] | null;
+function getElement<K extends keyof SVGElementTagNameMap>(
+  target: K
+): SVGElementTagNameMap[K] | null;
```
gitKrystan, line 12: These types started as copy-pasta from the built-in `querySelector` types, which look like this:

```ts
    /** Returns the first element that is a descendant of node that matches selectors. */
    querySelector<K extends keyof HTMLElementTagNameMap>(selectors: K): HTMLElementTagNameMap[K] | null;
    querySelector<K extends keyof SVGElementTagNameMap>(selectors: K): SVGElementTagNameMap[K] | null;
    querySelector<E extends Element = Element>(selectors: string): E | null;
```

...but as @chriskrycho pointed out [here](https://github.com/emberjs/ember-test-helpers/pull/1287#discussion_r1049755184), those types aren't that great.

Since we're diverging from them a bit by removing the generic from the `target: string` overload, I decided to also fix a type bug from those types, which is this:

`HTMLElementTagNameMap` and `SVGElementTagNameMap` have some overlap in keys, e.g. `a` is a valid tag in both. In the built-in `querySelector` types, the return type for `target: 'a'` would always be an `HTMLAnchorElement`, when the actual value might actually be a `SVGAElement`. So I added a third overload to account for that overlap. You can see the results in the type-tests.

addon-test-support/@ember/test-helpers/dom/trigger-event.ts
gitKrystan: Going to open a separate PR w/ the deep dive since I couldn't help myself. 😂

type-tests/api.ts
chriskrycho: Per my note on #1286, I think exporting it is great. 👍🏼

addon-test-support/@ember/test-helpers/dom/find-all.ts
```diff
@@ -5,6 +5,9 @@ import { toArray } from '../ie-11-polyfills';
 // would simply be defined as a tweaked re-export as `querySelector` is, but it
 // is non-trivial (to say the least!) to preserve overloads like this while also
 // changing the return type (from `NodeListOf` to `Array`).
+export default function findAll<
+  K extends keyof (HTMLElementTagNameMap | SVGElementTagNameMap)
+>(selector: K): Array<(HTMLElementTagNameMap | SVGElementTagNameMap)[K]>;
```
chriskrycho, line 10: While this works, I think it'd be slightly clearer to directly index the types:

```suggestion
>(selector: K): Array<HTMLElementTagNameMap[K] | SVGElementTagNameMap[K]>;
```

addon-test-support/@ember/test-helpers/dom/find.ts
```diff
@@ -1,15 +1,16 @@
 import getElement from './-get-element';
 
 // Derived from `querySelector` types.
+export default function find<
+  K extends keyof (HTMLElementTagNameMap | SVGElementTagNameMap)
+>(selector: K): (HTMLElementTagNameMap | SVGElementTagNameMap)[K] | null;
```
chriskrycho, line 6: Same thing here:

```suggestion
>(selector: K): HTMLElementTagNameMap[K] | SVGElementTagNameMap[K] | null;
```

addon-test-support/@ember/test-helpers/dom/trigger-event.ts
chriskrycho: Hahaha I love it. Feels very https://xkcd.com/356/ 😂 "A Project™" sounds exactly right, though.

chriskrycho (changes requested):
Noted one further pair of stylistic tweaks we should make, but looks good otherwise!

chriskrycho (approved):
I applied those tweaks via GH's review UI, so it's now ready to go. Thank you!

gitKrystan:
Found a tiny bug in this one. Will fix w/o force pushing ;-)

chriskrycho:
All right, I fixed a merge conflict of my own creation, will merge once this is ✅ on CI!

gitKrystan:
Should be all gouda now.

PR #1292, opened 2022-12-16
Make ember-auto-import@2 a dependency

SergeAstapov:
This is breaking change but would allow us to convert the addon to v2 format in a minor release


chriskrycho (approved):
💙

bertdeblock:
I'm unsure if this is correct. AFAIK, it's the consuming app that needs Embroider or `ember-auto-import` v2 to use a v2 addon. That's what's also mentioned in the v2 addon blueprint:
https://github.com/embroider-build/addon-blueprint/blob/main/files/README.md#compatibility.

SergeAstapov:
@bertdeblock idea behind this is that v2 addon requires app to either use Embroider or have eai v2, as you said.

however, any addon having eai v2 as dependency, also requires the app to have eai v2.
If addon didn’t have it before or used eai v1 - this is breaking change now to require eai v2.

idea of this PR is to make a breaking change now.

and once someone has a bandwidth, now we can convert to v2 addon fromat without breaking change.

bertdeblock:
`@ember/test-helpers` doesn't use any v2 addons and doesn't import any normal npm packages as far as I can tell, so technically it doesn't need `ember-auto-import` in `dependencies` in the first place I think?
I guess I don't understand why it's important to be able to do the v2 conversion in a minor release.

SergeAstapov:
> I guess I don't understand why it's important to be able to do the v2 conversion in a minor release.

Only to avoid another major release, no other realreason.

PR #1293, opened 2022-12-16
Drop support for Ember 3.x

rwjblue:
Now that the last Ember LTS of v3 is out of support range, we can drop support here.


chriskrycho (approved):
ESPECIALLY happy to see the various pre-Octane optional features die. Bye, jQuery integration!

PR #1298, opened 2022-12-16
Cleanup fallback codepaths for older Ember versions

rwjblue:
- Remove supporting code for Ember < 4
- Remove guards for supporting older Ember versions


chriskrycho (approved):

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -13,14 +13,24 @@ import { hbs, TemplateFactory } from 'ember-cli-htmlbars';
...
+import { importSync } from '@embroider/macros';
+
+let getInternalComponentManager: (
+  definition: object
+) => InternalComponentManager;
+
+// this would be much better as a regular `import` statement, but there really isn't any way to make typescript happy with thayar
+getInternalComponentManager =
+  //@ts-ignore
+  importSync('@glimmer/manager').getInternalComponentManager;
```
chriskrycho, line 33: (extremely non-blocking) We should see if we can write a type for `importSync` that makes this work. I'm not *sure* we can, but it's worth trying.

chriskrycho (approved):

addon-test-support/@ember/test-helpers/setup-rendering-context.ts
chriskrycho: Update, after poking a very little bit at it: I don’t think the language affordance is there. Alas.

chriskrycho:
This'll need a rebase/merge courtesy of merging #1299. I'll tackle it on Monday.

PR #1301, opened 2022-12-20
Remove the index signature from `TestContext`

dfreeman:
The `TestContext` type has a `[key: string]: unknown` index signature inherited from `BaseContext`. This has led to some confusion, as it muddies the error message folks see when they try to access fields that were present in the DT types but not the native ones, like `element`.

It also doesn't mesh super well with how folks tend to add custom properties to the context type by explicitly declaring them and then setting the `this` type appropriately for their tests. Ultimately we want for that pattern to go away entirely with `<template>`, of course, but in the meantime it would be nice to avoid extra confusion.

The index signature on `BaseContext` doesn't actually appear to be required, and ultimately the way we're using the type is equivalent to just `object`. Accordingly, I'm making `BaseContext` an alias for `object` here and otherwise attempting to touch the code as little as possible. The bits that _did_ have to change were largely places we were implicitly taking advantage of the index signature to skirt typechecking.

### DIFF
--- a/addon-test-support/@ember/test-helpers/setup-context.ts
+++ b/addon-test-support/@ember/test-helpers/setup-context.ts
@@ -59,16 +59,14 @@ registerWarnHandler((message, options, next) => {
   next.apply(null, [message, options]);
 });
 
-export interface BaseContext {
-  [key: string]: unknown;
-}
+export type BaseContext = object;
 
 /**
  * The public API for the test context, which test authors can depend on being
  * available.
  *
  * Note: this is *not* user-constructible; it becomes available by calling
- * `setupContext()` with a `BaseContext`.
+ * `setupContext()` with a base context object.
  */
 export interface TestContext extends BaseContext {
   owner: Owner;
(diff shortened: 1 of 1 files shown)


addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -59,34 +59,33 @@ registerWarnHandler((message, options, next) => {
...
 }
 
 // eslint-disable-next-line require-jsdoc
 export function isTestContext(context: BaseContext): context is TestContext {
+  let maybeContext = context as Record<string, unknown>;
   return (
-    typeof context['pauseTest'] === 'function' &&
-    typeof context['resumeTest'] === 'function'
+    typeof maybeContext['pauseTest'] === 'function' &&
+    typeof maybeContext['resumeTest'] === 'function'
```
chriskrycho, line 88: For a possible follow-on (keeping type changes separate from behavioral changes):

```ts
return (
  'pauseTest' in context && 
  typeof maybeContext['pauseTest'] === 'function' &&
  'resumTest' in context && 
  typeof maybeContext['resumeTest'] === 'function'
);
```

chriskrycho (approved):

addon-test-support/@ember/test-helpers/setup-context.ts
dfreeman: I had that and my editor was happy with it, but it turns out the `in` narrowing for `object` landed in TS 4.9, and definitely didn't want to touch that as part of this 😅

PR #1302, opened 2022-12-21
[backport] Remove the index signature from `TestContext`

chriskrycho:
This incorporates the same basic fix for `isRenderingTestContext` as the original (#1301) does for `isTestContext`, since that mechanic changes between v2 and v3.


eugenioenko:
Because  the type `BaseContext` was switched here from a dictionary to an `object` I'm seeing this type errors in the integration tests

`Property 'element' does not exist on type 'TestContext'` when accessing it with `this.element?.textContent`. And make sense, `element` doesn't exist in type `object`

Issue #1317, opened 2023-01-02
`find` and `findAll` examples are swapped


geneukum:
The `findAll` example shows up in the section of the API docs for the `find` helper and vice versa. We should swap these to make the docs as clear as possible.

PR #1318, opened 2023-01-02
[DOC] Swap `find` and `findAll` examples

geneukum:
The findAll example shows up in the section of the API docs for the find helper and vice versa. We should swap these to make the docs as clear as possible.

While we're here, `Commit: cba01bea` updated `setupContext`, renaming its first param from `context` to `base` in the process. It did not regenerate out the docs as part of the changeset, so building the docs results in this change locally.

Let's fixup that commit here so that we can build the docs without creating any local changes.

This closes https://github.com/emberjs/ember-test-helpers/issues/1317.

### DIFF
--- a/API.md
+++ b/API.md
@@ -834,7 +834,7 @@ Responsible for:
 
 #### Parameters
 
-*   `context` **[Object][72]** the context to setup
+*   `base` **[Object][72]** the context to setup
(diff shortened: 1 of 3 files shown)


chriskrycho (approved):
💙 thank you!

PR #1249, opened 2022-10-06
Convert to V2 addon

NullVoxPopuli:
This is a mega PR that only tries to get things working in order to prove the viability of v2 addon conversion.

Individual PRs will be pulled out once a green build is achieved.

- [ ] convert to pnpm
- [ ] move tests to their own app in a monorepo (pnpm supports to the top-level workspace pretending to be a package)
- [ ] convert the addon to a v2-addon, deleting all the old stuff

PR #1324, opened 2023-01-24
Use node 16 for development instead of node 14

NullVoxPopuli:
Pre-req for converting to a v2 addon.

~~According to @chriskrycho, we're preparing for a major anyway: https://github.com/emberjs/ember-test-helpers/pull/1249#issuecomment-1354062379~~

Requirement stems from: https://github.com/wessberg/rollup-plugin-ts/issues/202

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "webpack": "^5.75.0"
   },
   "engines": {
-    "node": "14.* || 16.* || >= 18"
+    "node": "16.* || >= 18"
   },
   "publishConfig": {
     "access": "public",
@@ -141,7 +141,7 @@
     }
   },
   "volta": {
-    "node": "14.21.1",
+    "node": "16.19.0",
     "yarn": "1.22.19"
   },
   "typesVersions": {


bertdeblock:
To me, it seems a bit strange to drop support for a specific Node version, just because `rollup-plugin-ts` accidentally included a breaking change in a minor release. Also, isn't `rollup-plugin-ts` only needed pre-publish?

NullVoxPopuli:
yup, true true -- so I guess.. motivation is wrong, but the dropping node 14 for development is still correct

kategengler:
Ember-CLI must still support Node 14 per our policy https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md. If this addon drops Node 14 we will not be able to update the version used in the blueprint until ember-cli drops Node 14.

NullVoxPopuli:
@kategengler when this addon is converted to a v2 addon, node doesn't matter at all.

the node version is irrelevant, as this will become browser-only code.

kategengler:
@NullVoxPopuli I understand that, but won't restricting the node version in the package.json as in this PR cause package managers to complain about the node version when used with Node 14?

NullVoxPopuli:
>  but won't restricting the node version in the package.json as in this PR cause package managers to complain about the node version when used with Node 14?

only if we release before the v2 conversion is complete -- I'd just remove the `engines` entry -- it's all much easier to demonstrate in the all-in-one v2 conversion PR.

doing the PRs piecemeal isn't exactly coherent :sweat_smile: 

>  also think there are other options here,

there is no need. We just _don't need to specify `engines` in the package.json_

NullVoxPopuli:
I updated the PR to be more reflective of what's going on / the plan when it comes to v2 conversions

package.json
```diff
@@ -100,7 +100,7 @@
     "webpack": "^5.75.0"
   },
   "engines": {
-    "node": "14.* || 16.* || >= 18"
+    "node": "16.* || >= 18"
```
NullVoxPopuli, line 103: this will be removed when I do the v2 conversion. no need for `engines`

kategengler: I don't see the point of merging a change that _cannot_ be released

NullVoxPopuli: We can't do an incremental migration to a v2 addon _and_ have the main branch always be in a releasable state....

chriskrycho: @NullVoxPopuli your current approach with splitting it up is fine.

@kategengler we do this all the time for "up next" kinds of branches, and it's literally no different from shipping an Ember major. Indeed: we'll do the exact same thing we would on Ember itself here: `master` becomes the branch for the next major, and if we ID things that need back-porting we'll create a `release-2-x` branch and do the usual back port dance.

chriskrycho (approved):

package.json
kategengler: @chriskrycho I agree we do this is in some projects, but most addons release from `master` and release much frequently. think it is fine, so long as anybody that may release knows that `master` is not in a release-able state.

kategengler: There's also a difference here in that dropping Node 14 would be a totally fine breaking major change for this project if we weren't concerned with using it in the CLI blueprint. That `engines` part of package.json will look normal and fine to anybody not thinking about CLI's Node support policy and could easily be accidentally released. 

I'd also like to push back on the idea that we merge things that are in a broken state in `ember.js`, even in a major. The canary branch of `ember.js` _is_ released weekly as an alpha, available as a nightly. We may have breaking changes queued up, but that only affects the version number with which it will be released.

chriskrycho: I’d be fine with targeting a `next` branch to ameliorate that concern, FWIW, but we should differentiate “broken”, which this is not, from “requires communication to avoid breaking folks if we’re not on the same page about release plans”, which you’re absolutely correct that this is.

kategengler: If you released this change CLI would not be able to update the addon, and anybody on Node 14 that took the update would get an error. Seems pretty broken to me? As @NullVoxPopuli said, the intention is this part of the change is never released.

kategengler: @NullVoxPopuli I don't think we need a multi-branch workflow for most addons. I think this is a special case because of `engines` and the CLI node policy. Except for policy-based things such as that, the main branch being releasable is enforced by CI. Personally, I would have left the less restrictive `engines` and just changed the development version of Node to make the move to a v2 addon so there wouldn't be a change that "we need to remember to undo". This is fine to move forward, we just need to remember to undo the engines change.

NullVoxPopuli: My plan was was to remove `engines` entirely, in the 4th PR in this series

1. (this one)
2. the pnpm one, because yarn and peers don't work (I have evidence if you need it, also yarn doesn't even support yarn... so...)
3. test-app split
4. v2 addon and `engines` removal

chriskrycho: @NullVoxPopuli FWIW although I merged this to keep a logjam, I think the better move here (and I should have thought of this!) was: bump the *Volta* version while leaving `engines` alone until that 4th PR. That's what we do in Glint for a variety of reasons. Let's keep that in mind for future reference.

NullVoxPopuli: Will do, for future addons, I'll do this series of PRs:

0. bump volta
1. pnpm 
2. test-app split
3. v2 addon and engines removal

PR #1325, opened 2023-01-24
Use pnpm for predictable dependencies

NullVoxPopuli:
This is prep for v2 addon conversion.
because `yarn@v1` is not supported by yarn, and is not a reliable tool for monorepos (which v2 addons are).

pnpm introduces strict dependency management (as does npm8 and yarn3, so there is no escaping strict dependencies!).

pnpm has growing usage in the ecosystem, and is actively maintained and regularly released.


.github/actions/pnpm/action.yml
```diff
@@ -0,0 +1,14 @@
+name: Setup node and pnpm
```
kategengler, line 1: Should we make a repo and publish this action so it doesn't need to be created in every addon?

addon-test-support/@ember/test-helpers/dom/fire-event.ts
```diff
@@ -20,7 +20,7 @@ const MOUSE_EVENT_CONSTRUCTOR = (() => {
 const DEFAULT_EVENT_OPTIONS = { bubbles: true, cancelable: true };
 
 export const KEYBOARD_EVENT_TYPES = tuple('keydown', 'keypress', 'keyup');
-export type KeyboardEventType = typeof KEYBOARD_EVENT_TYPES[number];
+export type KeyboardEventType = (typeof KEYBOARD_EVENT_TYPES)[number];
```
kategengler, line 23: why these changes?

kategengler (commented):
Is there any documentation that should change? The README?

addon-test-support/@ember/test-helpers/dom/fire-event.ts
NullVoxPopuli: prettier patch-version update adds these parens

.github/actions/pnpm/action.yml
NullVoxPopuli: that'd be helpful, I can make such an action

NullVoxPopuli:
> Is there any documentation that should change? The README?

yes! I forgot about that -- updates have been pushed

README.md
```diff
@@ -18,6 +18,11 @@ Installation
 ------------------------------------------------------------------------------
 ### For ember-qunit v5 and above
 
+#### yarn
```
bertdeblock, line 21:
```suggestion
#### pnpm
```

README.md
```diff
@@ -18,6 +18,11 @@ Installation
...
+```bash
+pnpm install --save-dev @ember/test-helpers
```
bertdeblock, line 23:
```suggestion
pnpm add --save-dev @ember/test-helpers
```

.github/workflows/ci-build.yml
```diff
@@ -16,22 +16,22 @@ jobs:
...
     name: 'Floating Dependencies'
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: volta-cli/action@v4
-      - run: yarn install --no-lockfile
-      - run: yarn test
+      - uses: actions/checkout@v3
+      - uses: NullVoxPopuli/action-setup-pnpm@v1
```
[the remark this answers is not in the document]
NullVoxPopuli: abstracting away `--no-lockfile` is not the point of the action, but to abstract away all the menial stuff needed for a pnpm repo. It's always possible to be more flexible if you don't use an abstraction :sweat_smile:

README.md
NullVoxPopuli: applied, thanks for catching this!

README.md
NullVoxPopuli: applied, thanks for catching this

.github/workflows/ci-build.yml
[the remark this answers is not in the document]
bertdeblock: I'm not sure I'm following. The purpose of the action seems to be to make setting up pnpm easier, but why must it already do the install itself? If you just have users run `pnpm install --no-lockfile` themselves, you don't need the `no-lockfile` option, and reading this CI file becomes clearer because you can read and see what's happening.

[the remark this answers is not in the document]
NullVoxPopuli: > but why must it already do the install itself?

reduced boilerplate

> and reading this CI file becomes clearer because 

everything is clearer when you don't need to read anything unfamiliar :thinking: 


I understand your concerns, I do, but as a maintainer of _many_ addons, I find boilerplate in these config files immensely annoying. 
I also want to make actions that abstract away the setup and such for ember-try and typescript version matrix testing.
IMO, having too much verbosity is about as harmful as too little. Thankfully, when errors occur in these actions, it's clear where, due to the stack trace being totally transparent about what code is where / from what dependency / action / etc.

NullVoxPopuli:
Closing in favor of the approach outlined here: https://github.com/emberjs/ember-test-helpers/pull/1356

PR #1344, opened 2023-03-01
[WIP] Expose registerHook and runHooks as official public APIs

drewlee:
Exposes `registerHook` and `runHooks` as official public APIs.

### DIFF
--- a/addon-test-support/@ember/test-helpers/helper-hooks.ts
+++ b/addon-test-support/@ember/test-helpers/helper-hooks.ts
@@ -1,6 +1,6 @@
-type Hook = (...args: any[]) => void | Promise<void>;
-type HookLabel = 'start' | 'end' | string;
-type HookUnregister = {
+export type Hook = (...args: any[]) => void | Promise<void>;
+export type HookLabel = 'start' | 'end' | string;
+export type HookUnregister = {
   unregister: () => void;
 };
 
@@ -62,7 +62,7 @@ export function registerHook(
 export function runHooks(
   helperName: string,
   label: HookLabel,
-  ...args: any[]
+  ...args: unknown[]
 ): Promise<void> {
   let hooks =
     registeredHooks.get(getHelperKey(helperName, label)) || new Set<Hook>();
--- a/addon-test-support/@ember/test-helpers/index.ts
+++ b/addon-test-support/@ember/test-helpers/index.ts
@@ -49,10 +49,8 @@ export { getDebugInfo } from './-internal/debug-info';
 export { default as registerDebugInfoHelper } from './-internal/debug-info-helpers';
 export type { TestMetadata } from './test-metadata';
 export { default as getTestMetadata } from './test-metadata';
-export {
-  registerHook as _registerHook,
-  runHooks as _runHooks,
-} from './-internal/helper-hooks';
+export { registerHook, runHooks } from './helper-hooks';
+export type { Hook, HookLabel, HookUnregister } from './helper-hooks';
 
 // DOM Helpers
 export { default as click } from './dom/click';
--- a/type-tests/api.ts
+++ b/type-tests/api.ts
@@ -67,6 +67,12 @@ import {
   DeprecationFailure,
   Warning,
   Target,
+  // Helper hooks
+  registerHook,
+  runHooks,
+  HookLabel,
+  Hook,
+  HookUnregister,
 } from '@ember/test-helpers';
 import { ComponentInstance } from '@glimmer/interfaces';
 import { Owner } from '@ember/test-helpers/build-owner';
@@ -264,4 +270,12 @@ expectTypeOf(getDeprecationsDuringCallback).toEqualTypeOf<
 expectTypeOf(getWarnings).toEqualTypeOf<() => Array<Warning>>();
 expectTypeOf(getWarningsDuringCallback).toEqualTypeOf<
   (callback: () => void) => Array<Warning> | Promise<Array<Warning>>
+>();
+
+// Helper hooks
+expectTypeOf(registerHook).toEqualTypeOf<
+  (helperName: string, label: HookLabel, hook: Hook) => HookUnregister
+>();
+expectTypeOf(runHooks).toEqualTypeOf<
+  (helperName: string, label: HookLabel, ...args: unknown[]) => Promise<void>
 >();
--- a/addon-test-support/@ember/test-helpers/dom/blur.ts
+++ b/addon-test-support/@ember/test-helpers/dom/blur.ts
@@ -4,7 +4,7 @@ import settled from '../settled';
 import Target from './-target';
 import { log } from '@ember/test-helpers/dom/-logging';
 import isFocusable from './-is-focusable';
-import { runHooks, registerHook } from '../-internal/helper-hooks';
+import { runHooks, registerHook } from '../helper-hooks';
 
 registerHook('blur', 'start', (target: Target) => {
   log('blur', target);
--- a/addon-test-support/@ember/test-helpers/dom/click.ts
+++ b/addon-test-support/@ember/test-helpers/dom/click.ts
@@ -5,7 +5,7 @@ import settled from '../settled';
 import isFormControl from './-is-form-control';
 import Target, { isWindow } from './-target';
 import { log } from '@ember/test-helpers/dom/-logging';
-import { runHooks, registerHook } from '../-internal/helper-hooks';
+import { runHooks, registerHook } from '../helper-hooks';
 
 const PRIMARY_BUTTON = 1;
 const MAIN_BUTTON_PRESSED = 0;
--- a/addon-test-support/@ember/test-helpers/dom/double-click.ts
+++ b/addon-test-support/@ember/test-helpers/dom/double-click.ts
@@ -6,7 +6,7 @@ import { DEFAULT_CLICK_OPTIONS } from './click';
 import Target, { isWindow } from './-target';
 import { log } from '@ember/test-helpers/dom/-logging';
 import isFormControl from './-is-form-control';
-import { runHooks, registerHook } from '../-internal/helper-hooks';
+import { runHooks, registerHook } from '../helper-hooks';
 
 registerHook('doubleClick', 'start', (target: Target) => {
   log('doubleClick', target);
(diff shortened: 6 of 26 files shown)


thegilby (commented):
Thanks for upgrading these @drewlee!

Would it be helpful to update the [API docs](https://github.com/emberjs/ember-test-helpers/blob/master/API.md) to include these?

drewlee:
@thegilby Yeah, I'm following up with updates to the the API docs.

drewlee:
@chriskrycho @scalvert

gitKrystan (approved):

addon-test-support/@ember/test-helpers/helper-hooks.ts
```diff
@@ -1,6 +1,23 @@
...
+  | 'scrollTo'
+  | 'select'
+  | 'tab'
+  | 'tap'
+  | 'triggerEvent'
+  | 'triggerKeyEvent'
+  | 'typeIn'
+  | 'visit'
+  | string;
+export type HookLabel = 'start' | 'end' | string;
```
chriskrycho, line 19: I recognize this one in particular was here previously, and so misled you, but it turns out that this type definition is basically meaningless: it's just `string`. Here's why: types in TypeScript are best thought of as *sets*, where union types (`A | B`) are the *union of the sets*. The set represented by `| string` at the end of this type *includes* the strings `'start'` and `'end'` (and all other strings):

image


This means that saying `'start' | 'end' | string` is *exactly* identical from a type-checking perspective to `string`.

In this case, that means we can remove this type definition entirely, along with its corresponding export. It also means we'll want to do the same on the newly-introduced `HookHelperName` definition above, which has the same limitation!

```suggestion
```

addon-test-support/@ember/test-helpers/helper-hooks.ts
```diff
@@ -12,7 +29,7 @@ const registeredHooks = new Map<string, Set<Hook>>();
  * @param {string} label A label to help identify the hook.
  * @returns {string} The compound key for the helper.
  */
-function getHelperKey(helperName: string, label: string): string {
+function getHelperKey(helperName: HookHelperName, label: string): string {
```
chriskrycho, line 32:
```suggestion
function getHelperKey(helperName: string, label: string): string {
```

addon-test-support/@ember/test-helpers/helper-hooks.ts
```diff
@@ -22,13 +39,13 @@ function getHelperKey(helperName: string, label: string): string {
...
+ *                       designating either the start or the end of the helper invocation.
  * @param {Function} hook The hook function to run when the test helper is invoked.
- * @returns {HookUnregister} An object containing an unregister function that will unregister
- *                           the specific hook registered to the helper.
+ * @returns {HookUnregister} An object containing an `unregister` function that unregisters
+ *                           the specific hook initially registered to the helper.
  */
 export function registerHook(
-  helperName: string,
+  helperName: HookHelperName,
```
chriskrycho, line 48:
```suggestion
  helperName: string,
```

addon-test-support/@ember/test-helpers/helper-hooks.ts
```diff
@@ -1,6 +1,23 @@
...
+  | 'render'
+  | 'scrollTo'
+  | 'select'
+  | 'tab'
+  | 'tap'
+  | 'triggerEvent'
+  | 'triggerKeyEvent'
+  | 'typeIn'
+  | 'visit'
+  | string;
```
chriskrycho, line 18: There are ways we could make this safe, in the future, by *not* allowing `| string`, and by requiring it to be one of the hooks we supply in the library; but that is *definitely* separate from the work you're doing here, so we should just go ahead and remove this entirely.

```suggestion
```

addon-test-support/@ember/test-helpers/index.ts
```diff
@@ -49,10 +49,13 @@ export { getDebugInfo } from './-internal/debug-info';
...
-  registerHook as _registerHook,
-  runHooks as _runHooks,
-} from './-internal/helper-hooks';
+export { registerHook, runHooks } from './helper-hooks';
+export type {
+  Hook,
+  HookHelperName,
+  HookLabel,
+  HookUnregister,
+} from './helper-hooks';
```
chriskrycho, line 58: We should be able to fold these together as well as dropping the now-removed type exports for `HookHelperName` and `HookLabel`:

```suggestion
export {
  registerHook,
  runHooks,
  type Hook,
  type HookUnregister,
} from './helper-hooks';
```

type-tests/api.ts
```diff
@@ -67,6 +67,13 @@ import {
   DeprecationFailure,
   Warning,
   Target,
+  // Helper hooks
+  registerHook,
+  runHooks,
+  Hook,
+  HookHelperName,
+  HookLabel,
+  HookUnregister,
```
chriskrycho, line 76: Now here we can again drop the unneeded exports; we should also `import type` as appropriate:

```suggestion
  type Hook,
  type HookUnregister,
```

type-tests/api.ts
```diff
@@ -265,3 +272,15 @@ expectTypeOf(getWarnings).toEqualTypeOf<() => Array<Warning>>();
 expectTypeOf(getWarningsDuringCallback).toEqualTypeOf<
   (callback: () => void) => Array<Warning> | Promise<Array<Warning>>
 >();
+
+// Helper hooks
+expectTypeOf(registerHook).toEqualTypeOf<
+  (helperName: HookHelperName, label: HookLabel, hook: Hook) => HookUnregister
```
chriskrycho, line 278:
```suggestion
  (helperName: string, label: string, hook: Hook) => HookUnregister
```

type-tests/api.ts
```diff
@@ -265,3 +272,15 @@ expectTypeOf(getWarnings).toEqualTypeOf<() => Array<Warning>>();
...
+>();
+expectTypeOf(runHooks).toEqualTypeOf<
+  (
+    helperName: HookHelperName,
```
chriskrycho, line 282:
```suggestion
    helperName: string,
```

chriskrycho (changes requested):
This needs one set of changes (which you should be able to apply directly as a batch from the **Files** tab on this PR, if I’ve done things correctly here!) to account for a way the existing types misled you as you were working on this.

Thanks for getting this done!

API.md
```diff
@@ -1166,6 +1169,50 @@ module('awesome-sauce', function(hooks) {
 
 Returns **([Array][70]\<Warning> | [Promise][66]<[Array][70]\<Warning>>)** An array of warnings information
 
+## registerHook
```
chriskrycho, line 1172: It looks like you hand-authored these additions. Unfortunately, that's going to get stomped as soon as anything is published, because [the API docs are generated from source][1]! You'll need to move these details into the TSDoc comments. 😅 

[1]: https://github.com/emberjs/ember-test-helpers/blob/a037f3080a80fd6c645a6b677c35a6cd9065d2d5/package.json#L32

addon-test-support/@ember/test-helpers/helper-hooks.ts
gitKrystan: Are 'start' and 'end' the most common labels that will be used? I assumed this was here more as documentation rather than expecting it to do anything re: type-checking. If we switch to just `string`, we should ensure that the documentation makes the 'start' and 'end' parts obvious.

type-tests/api.ts
gitKrystan: As an aside, we should consider adding this lint rule to the project to catch these: https://typescript-eslint.io/rules/consistent-type-imports/

addon-test-support/@ember/test-helpers/helper-hooks.ts
chriskrycho: They are the *defaults*, and in fact that's already documented at the point of use of the parameter in the function. It's possible someone intended it as "documentation" of sorts, but it tends very much to trip up folks new to TS and even if it was *intended* that way, we shouldn't *leave* it that way so that people don't get the wrong idea.

scalvert: @chriskrycho - you're comment on this ticket presumes that the original author (me) didn't understand how unions work in TS. This isn't the case. We weren't sure how this feature was going to shape up at the time, hence the reason it was private. Adding the `string` operated as a catch all for us, which allowed us the flexibility to do as we wanted without having to update this library constantly. 

So, it's not useless. Hooks evolved to only be used for `start` and `end`, so if anything, we should just remove the string fallback.

chriskrycho: Removing the `string` fallback makes good sense if it's now known that's all it should support! 👍🏼

API.md
drewlee: Doh! Thanks for point this out.

addon-test-support/@ember/test-helpers/helper-hooks.ts
drewlee: In addition to `start` and `end`, we also have `targetFound`. Included it and removed the `string` fallback.

drewlee:
I incorporated the requested changes yesterday and updated the associated TSDoc comments so that things get published out properly to the `API.md` doc (see the screenshot in the PR description). I think this is good to go, would appreciate another look @chriskrycho. Thanks!

chriskrycho (approved):

simonihmig:
Thanks for getting this fixed! @chriskrycho may I ask you for a new release, so that we can get https://github.com/ember-a11y/ember-a11y-testing/issues/498 unblocked? Thank you! 🙏

chriskrycho:
@simonihmig sorry for the delay; let's get #1389 in as well early next week and get a release out!

addon-test-support/@ember/test-helpers/index.ts
```diff
@@ -50,9 +50,12 @@ export { default as registerDebugInfoHelper } from './-internal/debug-info-helpe
...
 export {
-  registerHook as _registerHook,
-  runHooks as _runHooks,
-} from './-internal/helper-hooks';
+  registerHook,
+  runHooks,
+  type Hook,
+  type HookLabel,
+  type HookUnregister,
+} from './helper-hooks';
```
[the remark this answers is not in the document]
MelSumner: @rwjblue I think a new release of both will resolve the issue, yes? Coordinating a major version bump for this package now, and then it can be updated in `ember-a11y-testing`.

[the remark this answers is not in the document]
chriskrycho: @MelSumner we absolutely do not want to do a major of this right now. We have other TS things in flight which we want to land backwards-compatibly before that. This doesn't need a breaking change to release, it just needs a deprecation-and-not-removal here.

PR #1354, opened 2023-03-16
Add types/ to npmignore

elwayman02:
Since this package builds types to `public-types/` for publishing, it should not also be publishing the internal-focused `types/` directory. Adding it to npmignore to remove it from the published package.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -49,6 +49,7 @@
 /.github/
 /documentation.yml
 /server/
+/types/
 
 # typescript
 #


MelSumner:
This seems correct but the file has moved since the folder structure has been updated, so we'll need to update this PR supersede it. Apologies for the delay!

MelSumner:
Closing this in favor of #1396, apologies that it took so long to take care of.

PR #1356, opened 2023-03-21
Convert to single-workspace monorepo

NullVoxPopuli:
Taking a different approach for v2 conversion -- gonna try re-evaluating the validity of using yarn in v2 addon monorepos.

All steps done with https://github.com/NullVoxPopuli/ember-addon-migrator

Step 1: convert to monorepo (this PR)
Step 2: extract test app to separate workspace
Step 3: convert `./addon` to v2 addon
  -> this step 3 will include the removal of all the build-time sillyness and legacy ember-test-helpers things, leaving only `@ember/test-helpers`
:tada:


addon/package.json
```diff
@@ -0,0 +1,158 @@
...
+    "git": {
+      "tagName": "v${version}"
+    },
+    "github": {
+      "release": true,
+      "tokenRef": "GITHUB_AUTH"
+    }
+  },
+  "volta": {
+    "node": "16.19.0",
```
bertdeblock, line 148: Use `extends`? `node` version is now out of sync with the one defined in the root `package.json` file.

bertdeblock (commented):
Looks good AFAICT. Added one small suggestion.

addon/package.json
NullVoxPopuli: can do

phndiaye (approved):

PR #1378, opened 2023-05-01
ensureSafeComponent is not needed on Ember > 3.25

ef4:
This usage of `ensureSafeComponent` is a no-op on all the versions of ember supported by this addon.


MelSumner (approved):
let's ship it. :shipit:

PR #1379, opened 2023-05-01
remove has-ember-version

NullVoxPopuli:
This is a breaking change, but one required for v2 conversion.

We do not need this utility, as we can use embroider/macros. 

(but also `hasEmberVersion` still exists in `@ember/test-helpers` (for now))


NullVoxPopuli:
don't merge this breaking change.

Discovered here: https://github.com/emberjs/ember-test-helpers/pull/1381

that due to all the wierdness in ember-source, and how ember/test-helpers requires using some of ember's virtual packages, `@ember/test-helpers` cannot yet be converted to a v2 addon (easily)

MelSumner (approved):

PR #1381, opened 2023-05-02
V2 conversion

NullVoxPopuli:
I decided to see what is blocking while doing the whole conversion in one go

Pending
- [ ] convert ember-source to v2 (too much weird stuff is happening with the virtual packages rn)
- [x] update ci
- [x] finish conversion
- [ ] resolve build tooling issues
- [ ] ~~upstream patches~~
- [x] convert to pnpm

Things preserved:
 - the way public-types are built
 - addon is compiled via babel (but via rollup)
 - test app can import private APIs 
  
Breaking changes:
- no longer supports `setResolver` as `setApplication` already sets up an event-dispatcher, and for complexity reasons, support for custom event dispatcher has been removed in `createCustomResolver`

Non-Breaking changes:
- consuming node version doesn't matter (this is a browser only bundle)
- build-registry no longer uses private AMD apis, and instead uses embroider/macros
- tests no longer use private AMD apis
- ember-data is an optional peer dependency (for calling setupContainer)

once all core ember addons are v2, we can _much_ more easily do vite/vitest tests in non-monorepo repos (lots of cool things are unlocked once we drop broccoli / v1 addons entirely -- single-repos become an option again!).


NullVoxPopuli:
superseded by https://github.com/emberjs/ember-test-helpers/pull/1471

PR #1382, opened 2023-05-02
Remove code for pre-ember-v4

NullVoxPopuli:
Ember v4 is a requirement for the next release of `@ember/test-helpers`, so we can drop some code branches.


NullVoxPopuli:
Something seems wrong with `@glimmer/manager` :(

MelSumner:
Talked with @NullVoxPopuli and think this could also be good to have in a 3.x release.

addon/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
```diff
@@ -129,55 +128,33 @@ export function render(
       let OutletTemplate = lookupOutletTemplate(owner);
       let ownerToRenderFrom = options?.owner || owner;
 
-      if (macroCondition(dependencySatisfies('ember-source', '<3.24.0'))) {
```
NullVoxPopuli, line 132: this code change is the removal of this if branch, keeping only the else

addon/addon-test-support/@ember/test-helpers/-internal/is-component.ts
```diff
@@ -15,15 +14,21 @@ function isComponent(
   maybeComponent: object,
   owner: object
 ): maybeComponent is ComponentLike {
-  if (macroCondition(dependencySatisfies('ember-source', '>=3.25.0-beta.1'))) {
+  // SAFETY: in more recent versions of @glimmer/manager,
+  //         this throws an error when maybeComponent does not have
+  //         an associated manager.
+  try {
```
NullVoxPopuli, line 20: It appears that `@glimmer/manager` changed its behavior. This change absorbs the change in `@glimmer/manager`

MelSumner (approved):
These changes seem correct to me, and I'll wait until tomorrow before merging in case anyone else wants to take a second look.

PR #1383, opened 2023-05-02
Remove 2 requires

NullVoxPopuli:
require is private API to this package, and we have build-time tools via `@embroider/macros`

### DIFF
--- a/addon/addon-test-support/@ember/test-helpers/-internal/build-registry.ts
+++ b/addon/addon-test-support/@ember/test-helpers/-internal/build-registry.ts
@@ -3,8 +3,12 @@ import ApplicationInstance from '@ember/application/instance';
 import Application from '@ember/application';
 import EmberObject from '@ember/object';
 
-import require, { has } from 'require';
 import Ember from 'ember';
+import {
+  dependencySatisfies,
+  importSync,
+  macroCondition,
+} from '@embroider/macros';
 
 /**
  * Adds methods that are normally only on registry to the container. This is largely to support the legacy APIs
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
It is intended to be merged, apologies

addon/addon-test-support/@ember/test-helpers/-internal/build-registry.ts
```diff
@@ -116,13 +120,15 @@ export default function (resolver: Resolver) {
 
   exposeRegistryMethodsWithoutDeprecations(container);
 
-  if (has('ember-data/setup-container')) {
+  if (macroCondition(dependencySatisfies('ember-data', '>= 2.3'))) {
```
ef4, line 123: This only works if you declare a dependency (probably in this case an optional peer dependency) on ember-data.

Otherwise, a strict package manager like pnpm isn't going to be able to resolve ember-data from here, and the importSync would fail.

addon/tests/helpers/ajax.js
```diff
@@ -16,7 +16,6 @@ export default function ajax(url) {
       });
     });
   } else {
-    let fetch = require('fetch').default;
-    return fetch(url).then((response) => response.text());
+    return waitForPromise(fetch(url).then((response) => response.text()));
```
ef4, line 19: Are we sure this doesn't break fastboot?

I'm all in favor of switching to global fetch everywhere, but doing that correctly requires some work to make sure the global fetch does all the things that `require('fetch')` has been doing.

NullVoxPopuli: We can do the monorepo conversion first, and add a fastboot-using test app to be sure.

NullVoxPopuli:
Closing, as this is heavily conflicted

PR #1384, opened 2023-05-02
Remove extraneous event-dispatcher wiring and a amd/require

NullVoxPopuli:
This isn't needed, it seems?

we don't support folks doing `setResolver` and all apps use `setApplication`, yeah?


simonihmig (approved):
This indeed looks like some _very_ old and long-forgotten left overs! 😅

PR #1388, opened 2023-05-12
[v2 backport] Pass owner instead of registry to ember-data's setupContainer

### LINKED ISSUE #1386: Setup context fails when using ember-data v4 and a custom resolver
Hi

I'm having problems with my tests using a custom resolver [(because of `ember-engines`)](https://ember-engines.com/docs/testing-intro#rendering-unit-tests). The `setupContainer` function of `ember-data` v4+ removes deprecated compatibitlity code that allows using `optionsForType` on the application instance instead of the newer `registerOptionsForType` method:

https://github.com/emberjs/data/blob/7a957ddcf20c334fc6eeaf82cfa4f0dbb7795dfc/packages/-ember-data/addon/setup-container.js#L12-L39

`@ember/test-helpers` passes a faked registry object to `setupContainer` when using a custom resolver which does not have `registerOptionsForType` but only `optionsForType`. Shouldn't it pass the owner directly as `setupContainer` expects an application - not the registry?

I was able to fix the tests by passing `owner` to `setupContainer` instead of `registry || container` here:

https://github.com/emberjs/ember-test-helpers/blob/c25c5995ec60bc934740aaaf5d85299ede5ba0ee/addon/addon-test-support/%40ember/test-helpers/-internal/build-registry.ts#L126

I'd appreciate someone taking a look at this as my understanding of this addon is not enough for fixing this properly.

Please let me know if I can help in any way.

Thanks!

anehx:
The `setupContainer` function of `ember-data` expects an application instance to be passed - in this case that is the faked owner object that we create when using a custom resolver in tests.

This behaviour previously worked because `ember-data` supported the usage of legacy function `optionsForType` which exists on the `registry` object. However, they removed that legacy fallback in v4 and replaced it with `registerOptionsForType` which does not exist on the registry but only on the owner.

Resolves #1386

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -17,9 +17,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: volta-cli/action@v1
-        with:
-          node-version: 10.x
+      - uses: volta-cli/action@v4
       - run: yarn install --frozen-lockfile --ignore-engines
       - run: yarn lint
       - run: yarn test
@@ -31,9 +29,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: volta-cli/action@v1
-        with:
-          node-version: 12.x
+      - uses: volta-cli/action@v4
       - run: yarn install --no-lockfile
       - run: yarn test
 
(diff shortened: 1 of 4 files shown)


anehx:
@chriskrycho @MelSumner The tests for release, beta and canary will always fail here since this v2 of this addon still uses `assign` from `@ember/polyfills` which was removed in v5. However, I don't see the reason for the failing LTS tests.

Can someone support me with this? We currently can't update `ember-data` in various projects because of this issue, so this is important to us. Or is there already a release for v3 planned?

PR #1389, opened 2023-05-12
Support `@types`, preview types, and stable types for Ember.js

gitKrystan:
Requires https://github.com/emberjs/ember-test-helpers/pull/1405

I was getting all sorts of tsc issues on canary builds for an addon so took a stab at resolving them. This is a bit hack-y just to identify the issues.

Currently working to reconcile these types w/ existing preview types, as they are incompatible.


addon/addon-test-support/@ember/test-helpers/build-owner.ts
```diff
@@ -2,14 +2,14 @@ import Application from '@ember/application';
 import type { Resolver } from '@ember/owner';
 
 import legacyBuildRegistry from './-internal/build-registry';
-import ContainerProxyMixin from '@ember/engine/-private/container-proxy-mixin';
-import RegistryProxyMixin from '@ember/engine/-private/registry-proxy-mixin';
+import ContainerProxyMixin from '@ember/-internals/runtime';
+import RegistryProxyMixin from '@ember/-internals/runtime';
```
chriskrycho, line 6: I think these are going to need to be named imports instead!

```suggestion
import {
  ContainerProxyMixin,
  RegistryProxyMixin
} from '@ember/-internals/runtime';
```

addon/addon-test-support/@ember/test-helpers/build-owner.ts
```diff
@@ -2,14 +2,14 @@ import Application from '@ember/application';
...
 import CoreObject from '@ember/object/core';
 
 export interface Owner
   extends CoreObject,
-    ContainerProxyMixin,
-    RegistryProxyMixin {
+    ContainerProxyMixin.ContainerProxyMixin,
+    RegistryProxyMixin.RegistryProxyMixin {
```
chriskrycho, line 12: Then you won't need this. 😄

gitKrystan: LOL that makes sense. The TS error was something about namespaces which totally threw me off.

ef4:
Thanks for working on this!

addon/addon-test-support/@ember/test-helpers/-internal/build-registry.ts
```diff
@@ -57,13 +69,16 @@ const Owner = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixin, {
    * @see {@link https://github.com/emberjs/ember.js/pull/12680}
    * @see {@link https://github.com/emberjs/ember.js/blob/v4.5.0-alpha.5/packages/%40ember/engine/instance.ts#L152-L167}
    */
-  unregister(fullName: string) {
-    // @ts-expect-error
-    this['__container__'].reset(fullName);
+  /* eslint-enable valid-jsdoc */
+  unregister(this: Owner, fullName: FullName) {
+    // SAFETY: this is always present, but only the stable type definitions from
+    // Ember actually preserve it, since it is private API.
```
chriskrycho, line 75: For reference: we cannot `ts-expect-error` this because it *actually works* on the stable types. 😂

addon/addon-test-support/@ember/test-helpers/-internal/build-registry.ts
```diff
@@ -72,46 +87,62 @@ const Owner = EmberObject.extend(RegistryProxyMixin, ContainerProxyMixin, {
...
-  let fallbackRegistry, registry, container;
-  let namespace = EmberObject.create({
-    // @ts-expect-error
-    Resolver: {
-      create() {
-        return resolver;
-      },
+export default function buildRegistry(resolver: Resolver) {
+  let namespace = new Application();
+  // @ts-ignore: this is actually the correcct type, but there was a typo in
```
chriskrycho, line 92: Similarly, many of these work correctly… on *some* versions of Ember. I used `ts-ignore` here and below to make it so that if there *is* an error, it is suppressed :sigh: but if not, it still has type safety.

addon/addon-test-support/@ember/test-helpers/-internal/-owner-mixin-imports.ts
```diff
@@ -0,0 +1,78 @@
...
+// it should be impossible to ever get back more than one; otherwise, you will
+// end up with a union type of the two. That is very unlikely to be what you
+// want!
+type NonNever<Types extends Array<unknown>> = {
+  [Index in keyof Types]: IsNever<Types[Index]>;
+}[number];
+
+// Imports from `@types`
+// @ts-ignore
+import type CPM_DTS from '@ember/engine/-private/container-proxy-mixin.d.ts';
```
gitKrystan, line 43: Can you ELI5 this?

chriskrycho: Roughly, it imports from all the different paths where this type can exist, using `ts-ignore` on all of them to make the ones which are not present on any given Ember types version still pass the type checking, and then go through these type shenanigans so that it only resolves the one that actually does exist on the current version. The net result is that whether you are using the Definitely Typed types, the 4:8 preview types, the 4.12 preview types, or the stable types, one of these will be available and the others will not, and these exports will correctly resolve only that one. If you don’t do the type level shenanigans, you end up mixing `any` in (since that is the type of a `ts-ignore`d import which has no type definition at that module path), which is definitely not what we want.

PR #1394, opened 2023-05-25
Pulled in package.json and changelog from latest published branch

MelSumner:
Trying to figure out how to reconcile latest published with master branch but unsure of what strategy has been used, so creating a draft PR for visibility. 

- PRs seem to exist here
- package.json needed to be reconciled
- changelog needed to be reconciled

Trying to determine if there were any changes that made it into v2.9.1-v.2.9.3 that didn't also already make it into master branch.


MelSumner:
Noticing locally that these tests are pretty flaky. 🤔

MelSumner:
So this was a pretty useful exercise but I think we should cut the next version from the primary branch and not have separate branches for different versions. As such, I'm going to close this PR and focus on the PRs instead.

Issue #1397, opened 2023-06-01
Types missing from version 3


boris-petrov:
There are no TypeScript types in the release.

MelSumner:
the `types/` directory was excluded from the published package because the package builds types to `public-types/` for publishing. Can you check there please?

MelSumner:
ok! That means that our idea was not correct; I'll get a new patch out shortly. Thanks for confirming!

PR #1401, opened 2023-06-01
fixing the release

MelSumner:
in #1397 a user has reported that types are missing and they are getting an error. 

This PR updates the `.npmignore`  file and the package.json file to more correctly reflect what should be getting published re:types. 

This error seems to be related to the differences between `npm pack` and `yarn pack`.

Also not 100% sure this will be the fix but thanks to @gitKrystan @chriskrycho and @NullVoxPopuli for working through this toward a potential solution.

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -149,7 +149,7 @@
   "typesVersions": {
     "*": {
       "*": [
-        "public-types/@ember/test-helpers/*"
+        "public-types/*"
       ]
     }
   }
--- a/addon/package.json
+++ b/addon/package.json
@@ -149,7 +149,7 @@
   "typesVersions": {
     "*": {
       "*": [
-        "./public-types/*"
+        "./dist-types/*"
       ]
     }
   }
--- a/addon/package.json
+++ b/addon/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ember/test-helpers",
-  "version": "3.0.0",
+  "version": "3.0.1",
   "description": "Helpers for testing Ember.js applications",
   "keywords": [
     "ember-addon"
--- a/addon/package.json
+++ b/addon/package.json
@@ -149,7 +149,7 @@
   "typesVersions": {
     "*": {
       "*": [
-        "public-types/*"
+        "./public-types/*"
       ]
     }
   }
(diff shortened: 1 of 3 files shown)


MelSumner:
Ok it does seem like there are some undocumented pre-publish steps; working on a new release now. This reversion isn't necessary so I'm closing the PR.

MelSumner:
so I tried running some of the scripts in the package.json and re-publish, but that didn't seem to fix it either, so looks like we might need to revert the `.npmignore` change after all.

gitKrystan:
Looking into this. I think we need something like this in package.json:

https://github.com/emberjs/ember.js/blob/87c7fc72a3112936ae701a6768072d31caf4730d/package.json#LL173C3-L182C5

NullVoxPopuli:
@MelSumner public-types are emitted via: https://github.com/emberjs/ember-test-helpers/blob/master/addon/types.tsconfig.json#L7

but the `prepack` command should have included that: https://github.com/emberjs/ember-test-helpers/blob/master/addon/package.json#L29 🤔

addon/package.json
```diff
@@ -149,7 +149,7 @@
   "typesVersions": {
     "*": {
       "*": [
-        "public-types/@ember/test-helpers/*"
+        "public-types/*"
```
MelSumner, line 152:
```suggestion
        "./public-types/*"
```

addon/package.json
```diff
@@ -149,7 +149,7 @@
   "typesVersions": {
     "*": {
       "*": [
-        "public-types/@ember/test-helpers/*"
+        "./public-types/*"
```
MelSumner, line 152: @gitKrystan @chriskrycho there was a tiny variation between both of your suggestion, should it be `./public-types/*` or `public-types/*`?

gitKrystan: Either should work.

gitKrystan (approved):

MelSumner:
So even with these changes, I'm still not seeing the `public_types` folder when I run `npm pack` locally. 🤔

MelSumner:
Changing to `dist-types` from `public-types` seems to resolve the issue, in that I now get it locally when I run `npm pack` to check. But...why? makes no real logical sense. Why would it work before but not now? 🤔

PR #1404, opened 2023-06-06
Fix `isComponent` types

gitKrystan:
AFAICT [this change](https://github.com/emberjs/ember-test-helpers/pull/1382/files#diff-d3135c2567a56aba0c3f1dd551a3eca6e8c8287a802a584c2b127d8a032ad51a) from #1382 is incorrect but coincidentally works because `owner` is truthy.

It appears that `getInternalComponentManager`'s signature is `(definition: object, isOptional?: true | undefined) => InternalComponentManager | null` and has been for a while:

https://github.com/glimmerjs/glimmer-vm/blob/14413355ef6b48c1ab416a91fb7ff950cf8511d8/packages/%40glimmer/manager/lib/internal/index.ts#L177-L211

This issue was exposed by https://github.com/emberjs/ember-test-helpers/pull/1389

### DIFF
--- a/addon/addon-test-support/@ember/test-helpers/-internal/get-component-manager.ts
+++ b/addon/addon-test-support/@ember/test-helpers/-internal/get-component-manager.ts
@@ -1,3 +1,0 @@
-import { getInternalComponentManager as getComponentManager } from '@glimmer/manager';
-
-export default getComponentManager;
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
Thanks for fixing this -- this issue really puzzled me in #1382

NullVoxPopuli (approved):

PR #1406, opened 2023-06-06
fix embroider tests

NullVoxPopuli:
the `@embroider/*` packages are no longer released in lockstep, so each version must individually be fetched

### DIFF
--- a/addon/tests/dummy/config/ember-try.js
+++ b/addon/tests/dummy/config/ember-try.js
@@ -1,46 +1,9 @@
 'use strict';
 
 const getChannelURL = require('ember-source-channel-url');
-const latestVersion = require('latest-version');
+const { embroiderSafe } = require('@embroider/test-setup');
 
 module.exports = async function () {
-  const embroiderCore = await latestVersion('@embroider/core');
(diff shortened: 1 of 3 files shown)


addon/tests/dummy/config/ember-try.js
```diff
@@ -4,17 +4,20 @@ const getChannelURL = require('ember-source-channel-url');
 const latestVersion = require('latest-version');
 
 module.exports = async function () {
-  const EMBROIDER_VERSION = await latestVersion('@embroider/core');
+  const embroiderCore = await latestVersion('@embroider/core');
+  const embroiderWebpack = await latestVersion('@embroider/webpack');
+  const embroiderCompat = await latestVersion('@embroider/compat');
+  const embroiderTestSetup = await latestVersion('@embroider/testSetup');
```
gitKrystan, line 10:
```suggestion
  const embroiderTestSetup = await latestVersion('@embroider/test-setup');
```

gitKrystan (approved):

NullVoxPopuli:
thanks, @ef4 I've changed the fix to do the default thing that test-setup recommends

PR #1407, opened 2023-06-06
Check types in test matrix

gitKrystan:
With this change, CI for this package will detect issues caused by "breaking" differences between `@types` packages, ember preview types, and ember stable types.


chriskrycho:
I believe we can close this as superseded by #1389, where @gitKrystan did the same!

PR #1411, opened 2023-06-09
adds new workflow file to provide release branches

MelSumner:
If merged, this PR adds https://github.com/kategengler/put-built-npm-package-contents-on-branch/ which should provide a little extra confidence for folks doing new releases (especially as we add new maintainers and update packages to use the v2/embroider guidance). 

Still need to confirm some of the config settings but after that I will mark it ready for review.

### DIFF
--- a/.github/workflows/push-dist.yml
+++ b/.github/workflows/push-dist.yml
@@ -16,7 +16,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: volta-cli/action@v4
       - run: yarn install --frozen-lockfile --ignore-engines
-      - run: npm run build # Add whatever you need to build your pacakage here
       - uses: kategengler/put-built-npm-package-contents-on-branch@v1.0.0
         with:
           branch: ${{ github.head_ref || github.ref_name }}-dist
--- a/.github/workflows/push-dist.yml
+++ b/.github/workflows/push-dist.yml
@@ -1,0 +1,21 @@
+# .github/workflows/push-dist.yml
+
+name: Push dist
+
+on:
+  push:
+    branches:
+      - main
+      - master
+
+jobs:
+  push-dist:
+    name: Push dist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/push-dist.yml
```diff
@@ -0,0 +1,21 @@
...
+      - main
+      - master
+
+jobs:
+  push-dist:
+    name: Push dist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: npm run build # Add whatever you need to build your pacakage here
```
NullVoxPopuli, line 17: this is pretty close to what I'm using over here: https://github.com/NullVoxPopuli/ember-resources/blob/main/.github/workflows/push-dist.yml

but I think, because this repo is using yarn, we need these extra lines after the checkout:
https://github.com/emberjs/ember-test-helpers/blob/master/.github/workflows/ci-build.yml#L20-L21

NullVoxPopuli (changes requested):

.github/workflows/push-dist.yml
```diff
@@ -0,0 +1,21 @@
...
+    branches:
+      - main
+      - master
+
+jobs:
+  push-dist:
+    name: Push dist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
MelSumner, line 16:
```suggestion
      - uses: actions/checkout@v3
      - uses: volta-cli/action@v4
      - run: yarn install --frozen-lockfile --ignore-engines
```

.github/workflows/push-dist.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: volta-cli/action@v4
+      - run: yarn install --frozen-lockfile --ignore-engines
+      - run: npm run build # Add whatever you need to build your pacakage here
```
NullVoxPopuli, line 19:
```suggestion
```

build is not needed, because it runs as a part of `prepack`

NullVoxPopuli (approved):

PR #1413, opened 2023-06-09
Update addon/.gitignore

MelSumner:
In #1408 I should have caught that we now use `dist-types` instead of `public-types`, so updating the `.gitignore` to reflect that change.


NullVoxPopuli (approved):
lgtm

PR #1417, opened 2023-06-15
bugfix: handle `null` and `undefined` in `setProperties`

chriskrycho:
This implicitly worked for as long as `this.setProperties` existed, because it ultimately forwards to Ember's `setProperties`, which also implicitly works with this. Although this is not public API, and indeed is not desirable, changing it (implicitly and accidentally!) will break many consumers who are relying on it (most likely to end up happening transitively, e.g. via local/custom test helpers).


NullVoxPopuli (approved):
Thanks for the tests, comments, and explanation!

PR #1418, opened 2023-06-15
(backport) bugfix: handle `null` and `undefined` in `setProperties`

chriskrycho:
Backport #1417 to v2.x:

> This implicitly worked for as long as `this.setProperties` existed, because it ultimately forwards to Ember's `setProperties`, which also implicitly works with this. Although this is not public API, and indeed is not desirable, changing it (implicitly and accidentally!) will break many consumers who are relying on it (most likely to end up happening transitively, e.g. via local/custom test helpers).


NullVoxPopuli:
try config needs to be updated to _not_ use release/beta/canary on this branch, probably, but that's a separate set of work.

Thanks for backporting!

PR #1427, opened 2023-07-04
renderSettled was supposed to be public api, yeah?

NullVoxPopuli:


NullVoxPopuli:
jk, it's from 
```
import { renderSettled } from '@ember/renderer';
```

PR #1445, opened 2023-11-18
Fix ember-canary error/deprecation

chancancode:
The canary build started failing after https://github.com/emberjs/ember.js/pull/20570 was merged. We were never supposed to pass a `TemplateFactory` to `OutletState`, but it was previously tolerated. The pattern was fixed in the pass, but regressed when implementing support for `render(<template>...</template>)`.

No new tests because existing ones are already failing. If we wait till https://github.com/emberjs/ember.js/pull/20577 lands then we should see it turn into a deprecation, and this fix will remove the deprecation.

### DIFF
--- a/addon/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
+++ b/addon/addon-test-support/@ember/test-helpers/setup-rendering-context.ts
@@ -88,7 +88,7 @@ export interface RenderOptions {
   Renders the provided template and appends it to the DOM.
 
   @public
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

chancancode:
https://github.com/emberjs/ember-test-helpers/actions/runs/6911462222/job/18809097106 – it did trigger the deprecation (as intended) but we didn't set it to raise on deprecation

PR #1449, opened 2024-06-25
this change makes it so strict package managers installing duplicate …

void-mAlex:
…versions of test-helpers across different addons use a shared global state instead of module state

### DIFF
--- a/addon/addon-test-support/@ember/test-helpers/setup-context.ts
+++ b/addon/addon-test-support/@ember/test-helpers/setup-context.ts
@@ -98,7 +98,21 @@ export function isTestContext(context: BaseContext): context is TestContext {
   );
 }
 
-let __test_context__: BaseContext | undefined;
+/**
+  @private
+  @param {Object} it the global object to test
+  @returns {Boolean} it exists
+*/
+function check(it: any) {
+  // Math is known to exist as a global in every environment.
+  return it && it.Math === Math && it;
+}
+
+const globalObject =
+  check(typeof globalThis == 'object' && globalThis) ||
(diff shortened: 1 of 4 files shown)


ef4:
We discussed the polyfilling for globalThis and decide it's good. This is going to be back ported to majors where the polyfilling is needed. The implementation is inspired by the one in core-js, minus the parts that wouldn't possibly work inside an ES module.

addon/addon-test-support/@ember/test-helpers/setup-context.ts
```diff
@@ -98,7 +98,21 @@ export function isTestContext(context: BaseContext): context is TestContext {
   );
 }
 
-let __test_context__: BaseContext | undefined;
+/**
+  @private
+  @param {Object} it the global object to test
+  @returns {Boolean} it exists
+*/
+function check(it: any) {
```
NullVoxPopuli, line 106: why have this check?

why not:
```js
const globalObject = 
  (typeof globalThis === 'object' && globalThis) ||
  (typeof window === 'object' && window) ||
  (typeof self === 'object' && self) ||
  (typeof global === 'object' && global);
```

void-mAlex: first because it's heavily tested code from core-js implementation
second it tests for actual global instead of accidental variables

NullVoxPopuli:
This has failures right now, but it was previously passing, and the "engines" failures are due to floating deps that I can't control with yarn (I tried, and resolutions are brittle).

I'm going to merge this as is, then switch the repo to pnpm, add release-plan, and get a patch out

PR #1450, opened 2024-06-26
Fix ci due to floating deps

NullVoxPopuli:
~~Unblocks: https://github.com/emberjs/ember-test-helpers/pull/1449~~
(takes a different approach from the above)

Resolves this problem on the main branch: 

```bash
❯ yarn install --no-lockfile
❯ yarn lint:ts
yarn run v1.22.19
$ tsc --noEmit && tsc --noEmit --project type-tests
addon-test-support/@ember/test-helpers/dom/fill-in.ts:36:11 - error TS2345: Argument of type '() => Promise<HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement> | Promise<...>' is not assignable to parameter of type '(value: void) => HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement | PromiseLike<...>'.
  Type 'Promise<HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement> | Promise<...>' is not assignable to type 'HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement | PromiseLike<HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement>'.
    Type 'Promise<HTMLElementContentEditable>' is not assignable to type 'HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement | PromiseLike<HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement>'.
      Type 'Promise<HTMLElementContentEditable>' is not assignable to type 'PromiseLike<HTMLInputElement | HTMLButtonElement | HTMLSelectElement | HTMLTextAreaElement>'.
        Types of property 'then' are incompatible.
[...]

PR #1454, opened 2024-07-19
Test against ember LTS 5.4, and, 5.8

NullVoxPopuli:


NullVoxPopuli:
lint:ts will fail until https://github.com/emberjs/ember-test-helpers/pull/1450 is merged.

The red statuses are partial, as tests pass for most scenarios, but not lint:ts.

PR #1455, opened 2024-07-19
Pin isbinaryfile (a sub-sub dependency) so we can keep node support

NullVoxPopuli:


NullVoxPopuli:
This won't pass, due to not being a part of 
- https://github.com/emberjs/ember-test-helpers/pull/1450

The fix for that will have to be separate

PR #1458, opened 2024-07-19
Don't allow rimraf to be v5.0.7 or higher

NullVoxPopuli:
Rimraf:
5.0.7 - bumped engines.node from 14 to 14.18
5.0.8 - bumped engines.node to 18
5.0.9 - they tried, but still excluded what we use: `14 >=14.20 || 16 >=16.20 || >=18`


NullVoxPopuli:
this went down a rabbit hole of a bunch of things breaking due to yarn reasons, bunch of deps have gone wrong. gonna close in favor of switching to pnpm.

PR #1464, opened 2024-07-22
Use the right attribute - `maxlength` instead of `maxLength`

### LINKED ISSUE #1459: Read "maxlength" attribute instead of "maxLength" in "-guard-for-maxlength.ts"
I had opened [this](https://github.com/emberjs/ember-test-helpers/issues/1202) a while back but didn't end up creating a PR.

**background**: The following piece of code in "-guard-for-maxlength.ts" reads "maxLength" (camelcase") attribute instead of "maxlength" (lowercase)
https://github.com/emberjs/ember-test-helpers/blob/3b83f12656ab9443603da79575ab8a744d56fcb9/addon/addon-test-support/%40ember/test-helpers/dom/-guard-for-maxlength.ts#L22

As per [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) MDN document, the actual attribute is "maxlength" (lowercase) and **NOT** "maxLength" (camel case).

In fact, in the same file, few lines below, the following line of code does this correctly.
https://github.com/emberjs/ember-test-helpers/blob/3b83f12656ab9443603da79575ab8a744d56fcb9/addon/addon-test-support/%40ember/test-helpers/dom/-guard-for-maxlength.ts#L41

I think, this should be fixed for consistency and correctness.
Please let me know if I can work on this and can create a PR. 

Thanks.

shubham88fru:
Fixes #1459 

Issue: `maxLength` attribute is being read instead of the correct `maxlength` attribute.

### DIFF
--- a/addon/addon-test-support/@ember/test-helpers/dom/-guard-for-maxlength.ts
+++ b/addon/addon-test-support/@ember/test-helpers/dom/-guard-for-maxlength.ts
@@ -19,7 +19,7 @@ function isMaxLengthConstrained(
   element: Element
 ): element is HTMLInputElement | HTMLTextAreaElement {
   return (
-    !!Number(element.getAttribute('maxLength')) &&
+    !!Number(element.getAttribute('maxlength')) &&
     (element instanceof HTMLTextAreaElement ||
       (element instanceof HTMLInputElement &&
         constrainedInputTypes.indexOf(element.type) > -1))


NullVoxPopuli:
is there a way to test that `maxLength` is wrong or unsupported somehow?

shubham88fru:
> is there a way to test that `maxLength` is wrong or unsupported somehow?

@NullVoxPopuli per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute), the actual attribute is `maxlength` and not `maxLength`. 
Now the reason that it still worked with `maxLength` was because `el.getAttribute(arg)` internally lowercases whatever arg was passed (check this [point](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#lower_casing) in MDN) effectively converting `maxLength` to `maxlength`. So to your point, `maxLength` is technically wrong but didn't break anything yet because of the lowercasing done by `el.getAttribute()`.

However, for the sake of consistency and correctness, I suggest that we change the attribute to the right value, which is `maxlength`.

Note that in the same file, the same attribute is also read at a different place correctly - 
https://github.com/emberjs/ember-test-helpers/blob/3b83f12656ab9443603da79575ab8a744d56fcb9/addon/addon-test-support/%40ember/test-helpers/dom/-guard-for-maxlength.ts#L41

Does that answer your question? Sorry if I misunderstood what you were asking.

PR #1467, opened 2024-07-24
Fix import type via upgrading @tsconfig/ember

NullVoxPopuli:
As a part of the v2 conversion process, we need to correctly declare our type imports as types (rather than values).

This enables tooling later to completely strip imports if unused. 

-----------

For V2 Conversion
- extract test-app
- fix type-imports <- this PR
- find old code and remove it (optional, unless it's tied to the build)
- native library conversion

depending on how these PRs go will determine if we need to do a major or not.


kategengler:
I'm not sure about this dropping support for TS versions still supported by ember.js itself without doing a major version to indicate changed ember.js support. cc @wagenet @gitKrystan

kategengler:
ember-source only tests against 4.9+ of typescript, if that helps

NullVoxPopuli:
all green! and no longer dropping support for anything!

kategengler (approved):

PR #1468, opened 2024-07-24
Move tests to new test-app

NullVoxPopuli:
V2 addons don't have an integrated test setup atm, so a separate test app is how we test v2 addons.

Test-helpers isn't a v2 addon yet, but this'll greatly reduce the diff of that upcoming change.

This is a non-breaking change as it mostly focuses on using a new app for the tests instead of using the "dummy app"

-----------

For V2 Conversion
- extract test-app <- this PR
- fix type-imports 
- find old code and remove it (optional, unless it's tied to the build)
- native library conversion

depending on how these PRs go will determine if we need to do a major or not.


test-app/tests/test-helper.js
```diff
@@ -35,19 +38,6 @@ QUnit.done(function () {
   }
 });
 
-class TestLoader extends AbstractTestLoader {
```
NullVoxPopuli, line 38: handled by ember-qunit these days

test-app/tests/test-helper.js
```diff
@@ -109,8 +99,8 @@ QUnit.assert.deprecationsInclude = function (expected) {
   });
 };
 
-QUnit.start();
-
-if (typeof Testem !== 'undefined') {
```
NullVoxPopuli, line 114: also handled by ember-qunit

test-app/tests/unit/setup-context-test.js
```diff
@@ -29,12 +29,14 @@ import config from '../../config/environment';
 import Ember from 'ember';
 import { deprecate, warn } from '@ember/debug';
 
+console.log('1', Ember.testing);
```
lifeart, line 32: seems console logs could be omitted here

test-app/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "qunit": "^2.21.1",
+    "qunit-dom": "^3.2.0",
+    "tracked-built-ins": "^3.1.1",
+    "webpack": "^5.78.0"
+  },
+  "engines": {
+    "node": "14.* || 16.* || >= 18"
+  },
+  "ember": {
+    "edition": "octane"
```
lifeart, line 78: is edition still a thing here?

test-app/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "loader.js": "^4.7.0",
+    "pretender": "^3.4.7",
+    "prettier": "^2.8.7",
+    "qunit": "^2.21.1",
+    "qunit-dom": "^3.2.0",
+    "tracked-built-ins": "^3.1.1",
+    "webpack": "^5.78.0"
+  },
+  "engines": {
+    "node": "14.* || 16.* || >= 18"
```
lifeart, line 75: wondering if we could drop 14 and 16..

test-app/package.json
NullVoxPopuli: Yup

test-app/tests/unit/setup-context-test.js
NullVoxPopuli: Oops. Thanks!

test-app/package.json
NullVoxPopuli: We could, but would me separate. During v2 conversion, node would cease to matter anyway

kategengler (approved):
The Readme needs to be updated to reflect this change (right now says `ember test`) as does CONTRIBUTING.md (or removed, there isn't much there that isn't in the Readme)

PR #1471, opened 2024-07-29
V2 Addon Conversion

NullVoxPopuli:
PRs to extract:
- [x] drop support for TS < v5
      https://github.com/emberjs/ember-test-helpers/pull/1474
- [x] don't use `require`
      https://github.com/emberjs/ember-test-helpers/pull/1472
- [x] add a CI check for [docs generation](https://github.com/emberjs/ember-test-helpers/blob/master/addon/package.json#L32) 
      https://github.com/emberjs/ember-test-helpers/pull/1477
- [x] make a strict `.npmrc`
      https://github.com/emberjs/ember-test-helpers/pull/1475
- [x] don't use a shared `pnpm-lock.yaml` (for more robust testing with both `@types/*` and built-in types)
      https://github.com/emberjs/ember-test-helpers/pull/1476

-----------

Steps:
- `cd addon`
- Upgrade local ember-cli
- `npx ember-cli@latest init --blueprint @embroider/addon-blueprint --typescript --addon-only`
- work through diffs
- add file extensions to all imports (required by the default rollup config)
- `pnpm lint:fix`
- manually address / disable remaining lint failures


addon/src/setup-rendering-context.ts
```diff
@@ -1,27 +1,33 @@
...
+import hasEmberVersion from './has-ember-version.ts';
+import isComponent from './-internal/is-component.ts';
+import { ComponentRenderMap, SetUsage } from './setup-context.ts';
+
+// the built in types do not provide types for @ember/template-compilation
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { precompileTemplate } from '@ember/template-compilation';
+
+const OUTLET_TEMPLATE = precompileTemplate(`{{outlet}}`, { strictMode: false });
```
NullVoxPopuli, line 25: I didn't want to import from ember-cli-htmlbars, so I used `@ember/template-compilation` for `precompileTemplate`

addon/src/-internal/build-registry.ts
```diff
@@ -141,20 +145,27 @@ export default function buildRegistry(resolver: Resolver) {
...
 
-  if (has('ember-data/setup-container')) {
-    // ember-data is a proper ember-cli addon since 2.3; if no 'import
-    // 'ember-data'' is present somewhere in the tests, there is also no `DS`
-    // available on the globalContext and hence ember-data wouldn't be setup
-    // correctly for the tests; that's why we import and call setupContainer
-    // here; also see https://github.com/emberjs/data/issues/4071 for context
-    let setupContainer = require('ember-data/setup-container')['default'];
-    setupContainer(owner);
+  if (macroCondition(dependencySatisfies('ember-data', '>= 2.3'))) {
```
NullVoxPopuli, line 154: This change is also meaningful -- we don't want to use requirejs -- we're trying to get rid of it.

Insetad, our abstraction over runtime module loader is provided by embroider-macros.

Additionally, I saw that in this file that setup-container is deprecated without replacement for removal in v6.

So I added another if condition around this code so we can hopefully be a little fault tolerant when ember-data v6 comes around.

NullVoxPopuli: This could also be pulled out to a separate PR, if folks want

addon/src/setup-rendering-context.ts
NullVoxPopuli: This could also be pulled out to a separate PR, if folks want

PR #1473, opened 2024-08-15
Get tests passing w/ Chrome's new selectionchange event behavior

NullVoxPopuli:
Unblocks: https://github.com/emberjs/ember-test-helpers/pull/1470

Last successful build (nightly cron): https://github.com/emberjs/ember-test-helpers/actions/runs/10173407362
Next nightly cron failed: https://github.com/emberjs/ember-test-helpers/actions/runs/10191329093

Findings:

- failures only affect chrome (127.0.6533.99, 127.0.6533.100, (etc?))
  - Chrome Canary (129.0.6658.0) has the same behavior
  - relevant changes https://chromium.googlesource.com/chromium/src/+log/127.0.6533.77..127.0.6533.120?pretty=fuller&n=10000
  - potentially related bug reports
    - https://issues.chromium.org/issues/346898523 
      - but none of them recent: https://issues.chromium.org/issues?q=selectionchange%20created%3E2024-07-30
  - Could this be caused by https://chromestatus.com/feature/5255454895898624 ? 
    - Relates to this: https://w3c.github.io/selection-api/#selectionchange-event
      - which _could_ imply that we _are_ changing the selection range somehow in tests. 
        - I could see this being true for dblclick tests, but not for single click 🤔 
        - We can reproduce lots of `selectionchange` events in this jsbin demo I made: https://jsbin.com/feseyif/edit?html,output
- no failures are present in firefox (129.0.1, 130.0b5)
- no failures are present in safari (7.5 (19618.2.12.11.6))

In the changes in this PR:
- added selection event, where appropriate, conditioned on "isChrome"  
  - of note, if the browser doesn't have focus the order of the keyup and selection change events flip
    - same for the mouseup, click and selectionchange behavior

### DIFF
--- a/test-app/tests/unit/dom/blur-test.js
+++ b/test-app/tests/unit/dom/blur-test.js
@@ -7,7 +7,7 @@ import {
   registerHook,
 } from '@ember/test-helpers';
 import { buildInstrumentedElement, insertElement } from '../../helpers/events';
-import { isEdge } from '../../helpers/browser-detect';
+import { isEdge, isChrome } from '../../helpers/browser-detect';
 import hasEmberVersion from '@ember/test-helpers/has-ember-version';
 import { createDescriptor } from 'dom-element-descriptors';
 
@@ -16,6 +16,10 @@ let blurSteps = ['blur', 'focusout'];
 
 if (isEdge) {
   blurSteps = ['focusout', 'blur'];
+}
+
+if (isChrome) {
+  focusSteps.push('selectionchange');
 }
 
 module('DOM Helper: blur', function (hooks) {
@@ -53,7 +57,11 @@ module('DOM Helper: blur', function (hooks) {
   });
 
   test('it executes registered blur hooks', async function (assert) {
-    assert.expect(13);
+    if (isChrome) {
+      assert.expect(15);
+    } else {
+      assert.expect(13);
+    }
 
     let element = document.createElement('input');
     insertElement(element);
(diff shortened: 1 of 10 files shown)


test-app/tests/unit/dom/blur-test.js
```diff
@@ -18,6 +18,10 @@ if (isEdge) {
   blurSteps = ['focusout', 'blur'];
 }
 
+if (isChrome) {
```
wagenet, line 21: Why did we not do `selectionchange` for Firefox since that supports it also?

test-app/tests/unit/dom/click-test.js
```diff
@@ -202,6 +203,17 @@ module('DOM Helper: click', function (hooks) {
...
     let clickSteps = ['mousedown', 'focus', 'focusin', 'mouseup', 'click'];
 
+    if (isChrome) {
+      clickSteps = [
+        'mousedown',
+        'focus',
+        'focusin',
+        'mouseup',
+        'click',
+        'selectionchange',
```
wagenet, line 213: Should we just push `selectionchange` to the existing list?

wagenet (approved):

test-app/tests/unit/dom/blur-test.js
NullVoxPopuli: The behavior is slightly different.

When I was working on this PR, I had 5 browsers open
- FireFox
- FireFox Developer Edition
- Safari
- Chrome
- Chrome Canary

😅

test-app/tests/unit/dom/click-test.js
NullVoxPopuli: then we'd need to add inverted conditions for safari and firefox

PR #1486, opened 2024-08-23
Remove old code that existed for now unsupported ember-sources (2.4, etc)

NullVoxPopuli:
NOTE:
- lockfile deps may need an update to get this to work -- in particular, you'll need to make sure your lockfile at least contains these versions: https://github.com/embroider-build/embroider/releases/tag/v3.6.1-%40embroider%2Fcompat 

--------------

Not having embroider-optimized enabled has caused some bug reports related to resolving packages (`@ember/-internals`, `@ember/renderer`).

See: https://github.com/emberjs/ember-test-helpers/pull/1487

embroider-optimized testing was disabled in https://github.com/emberjs/ember-test-helpers/commit/6828d0c74a8f400d6ddaf7d089b9f9196db850a3#diff-efa7561fd196a5304c1c80012f1f6153c58faf5697ba3a59465ae7ce06b8e550R146
(in August 2021)

This library's minimum supported ember version is 4.0: https://github.com/emberjs/ember-test-helpers
image


test-app/tests/unit/setup-rendering-context-test.js
```diff
@@ -287,15 +286,21 @@ module('setupRenderingContext', function (hooks) {
     });
 
     test('can invoke helper', async function (assert) {
-      await render(hbs`{{jax "max"}}`);
+      this.setProperties({
+        jax: helper(([name]) => `${name}-jax`),
+      });
+      await render(hbs`{{this.jax "max"}}`);
```
NullVoxPopuli, line 292: these helpers were (rightfully so!) trying to look up the helper defined in app/helpers/jax.

However, that doesn't exist, and since we don't actually support custom resolvers for normal apps, this is fine. These tests are testing that `render` can invoke helpers, and helpers can be defined on `this`

addon/src/settled.ts
```diff
@@ -190,19 +190,24 @@ if (loader.registry['ember-testing/test/waiters']) {
...
+  if (EmberTest.waiters) {
     if (
       EmberTest.waiters.some(([context, callback]) => !callback.call(context))
     ) {
       return true;
     }
   }
 
-  return false;
+  return EmberTest.checkWaiters();
```
NullVoxPopuli, line 210: changing this from `false` to `EmberTest.checkWaiters()` _is_ a meaningful change, but the only mainingful change in this PR (for consumers).

The code for `checkWaiters` is here: https://github.com/emberjs/ember.js/blob/85a4f298f67ff70395cf7f9103682335162e0606/packages/ember-testing/lib/test/waiters.ts#L111

and those callbacks are part of the legacy waiter system.

There are existing comments about removing code for accessing legacy-waiter stuff (because ember.js has since abstracted it so we don't need to be so weird in this library) -- cleaning that up can happen in a followup PR, and should be non-breaking, and retain legacy waiter support.

NullVoxPopuli: So.... I suppose this should be a patch-release,  rather than non-release.

driesdl (approved):
To me it looks fine, but I'm not familiair with the in-depths aspects of it all. (trying to do one PR each day ;-) )

simonihmig:
We just run into an issue where tests were failing when switching to Embroider optimized (works for safe mode!) due to `await settled()` not capturing some pending stuff with test waiters, because those AMD requires that were still around wouldn't work. So AFAICT I think this PR should hopefully fix these cases! 🤞

NullVoxPopuli:
Ye! I'm hopeful -- we need this tho:
https://github.com/embroider-build/embroider/pull/2075


This pr currently contains a patch

NullVoxPopuli:
Updated embroider (via lockfile bumps), and things are green now without the patch!

lifeart (approved):

ef4 (approved):

addon/src/settled.ts
```diff
@@ -2,9 +2,9 @@
 // "provides" this public API, but does not for earlier versions. As a result,
 // this type will be `any`.
 import { _backburner } from '@ember/runloop';
-import Ember from 'ember';
-
-import EmberApplicationInstance from '@ember/application/instance';
+import { Test } from 'ember-testing';
+export const checkWaiters = Test.checkWaiters;
```
simonihmig, line 6: Why is this being exported? This seems to have caused a type regression, see https://github.com/emberjs/ember-test-helpers/issues/1502

NullVoxPopuli: great question -- I had originally thought that the original implementation was exported -- but I see in this diff that it is not. let's see if not exporting it resolves that issue

PR #1508, opened 2024-09-25
Revert "Get tests passing w/ Chrome's new selectionchange event behavior"

NullVoxPopuli:
Reverts emberjs/ember-test-helpers#1473

Was this a temporary whoopsie that i mistakenly got caught up in hue to lack of luck?


NullVoxPopuli:
Good, still fails.

Issue #1452, opened 2024-07-19
Proposal: Allow context to teardown and setup repeatedly in a test


mixonic:
### The goal

I maintain a suite of tests that use a helper `simulatePageRefresh` to destroy the currently running application, then instantiate a new one. The implementation, compatible until ember-qunit 4 (and possibly 5), looks like this:

```js
export async function simulatePageRefresh() {
  let context = getContext();

  await teardownContext(context);
  await setupContext(context);
  await setupApplicationContext(context);
}
```

This helper allows us to assert, in application tests, that a fresh page load (from Ember's perspective) performs data loading correctly (either from mocked APIs or from real APIs).

This functionality is not a robust as a true page reload, such as we would get from writing these tests in Playwright. However, as we already have these tests in Ember-QUnit and Ember-QUnit already embraces booting an application multiple times in the same browsing context, this functionality doesn't seem bizarre.

### Why our implementation no longer works

In ember-qunit, `setupContext` and `teardownContext` define how a testing context is configured and reset. See https://github.com/emberjs/ember-qunit/blob/dafee6bdb9a93ab6a83de7421adad6669df31d95/addon/src/index.js#L51

`setupContext`, at https://github.com/emberjs/ember-test-helpers/blob/9cec68dc6aa9c0a7a449eb89797eb81299fa727f/addon/addon-test-support/%40ember/test-helpers/setup-context.ts#L382 uses the QUnit test context (passed [here](https://github.com/emberjs/ember-qunit/blob/dafee6bdb9a93ab6a83de7421adad6669df31d95/addon/src/index.js#L58)) as a key for destructors and even adds the context into the destroyable tree (associated with the Ember `owner` instance) at https://github.com/emberjs/ember-test-helpers/blob/9cec68dc6aa9c0a7a449eb89797eb81299fa727f/addon/addon-test-support/%40ember/test-helpers/setup-context.ts#L425.

Although `teardownContext` calls `destroy` with the context at https://github.com/emberjs/ember-test-helpers/blob/9cec68dc6aa9c0a7a449eb89797eb81299fa727f/addon/addon-test-support/%40ember/test-helpers/teardown-context.ts#L25, this does not prevent the `context` from being tracked as "destroyed". That state then makes it impossible to call `setupContext` a second time, as the destructible codepaths complain the instance was already destroyed.

### Proposal for discussion

I think there are several ways forward:

* We can restore using `teardownContext` and `setupContext` multiple times in a test as a functionality of this library
[...]

PR #1511, opened 2024-10-22
Make test context reusable during a test

eliasdawson:
In v2.0.0 @ember/test-helpers refactored to use `@ember/destroyable` for teardown which also meant
that the application context was destroyed during teardown, where previously it was only cleaned up
and the application (`context.owner`) was destroyed.

In our application tests we simulate a page refresh by calling `teardownContext` and
`setupContext`. The destroyable change broke this functionality because the context is now marked
as destroyed.

This commit effectively reverts the change to use `@ember/destroyable` and adds tests to ensure that
the context is not destroyed so that it can be reset during a test.

This addresses https://github.com/emberjs/ember-test-helpers/issues/1452 raised by @mixonic 

The @ember/destroyable change was made in #914

### DIFF
--- a/addon/src/teardown-context.ts
+++ b/addon/src/teardown-context.ts
@@ -1,5 +1,7 @@
 import type { TestContext } from './setup-context';
-import settled from './settled.ts';
+import Ember from 'ember';
+import { unsetContext } from './setup-context.ts';
+import settled, { _teardownAJAXHooks } from './settled.ts';
 import { _cleanupOnerror } from './setup-onerror.ts';
 import { destroy } from '@ember/destroyable';
 
@@ -30,7 +32,14 @@ export default function teardownContext(
     .then(() => {
       _cleanupOnerror(context);
 
-      destroy(context);
+      _teardownAJAXHooks();
+
+      // SAFETY: this is intimate API *designed* for us to override.
+      (Ember as any).testing = false;
+
+      unsetContext();
+
+      destroy(context.owner);
     })
     .finally(() => {
       if (waitForSettled) {
--- a/test-app/tests/unit/dom/fill-in-test.js
+++ b/test-app/tests/unit/dom/fill-in-test.js
@@ -16,16 +16,6 @@ if (isFirefox || isChrome) {
   clickSteps.push('selectionchange');
 }
 
-/**
- * Prior to Chrome 129 (canary),
- * Chrome 127.x emits an extra selectionchange event sometimes
- *
- * Delete this once we don't want to test against Chrome 127
- */
-if (isChrome) {
-  clickSteps.push('selectionchange');
-}
-
 module('DOM Helper: fillIn', function (hooks) {
   if (!hasEmberVersion(2, 4)) {
     return;
--- a/addon/src/setup-context.ts
+++ b/addon/src/setup-context.ts
@@ -7,7 +7,7 @@ import type { Resolver } from '@ember/owner';
 import { setOwner } from '@ember/application';
 
 import buildOwner, { type Owner } from './build-owner.ts';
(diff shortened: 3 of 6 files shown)


addon/src/teardown-context.ts
```diff
@@ -1,5 +1,7 @@
 import type { TestContext } from './setup-context';
-import settled from './settled.ts';
+import Ember from 'ember';
```
NullVoxPopuli, line 2: we cannot import from `ember` as it's planned to be deprecated soonish in the v6 series.

See: https://github.com/emberjs/rfcs/pull/1003

the RFC also has a big list of alternate ways to get what you need though.

NullVoxPopuli:
wouldn't we want "refresh" to mean you get a whole new context?

a destroy + new one?

eliasdawson:
> first, hello!!!!
> 
> second,
> 
> wouldn't we want "refresh" to mean you get a whole new context?
> 
> a destroy + new one?
> 
> (as close to actual page refresh as possible?)

Hello, yourself :)

I don't think we do want a new context. We only want a new app (context.owner). The context is the qunit context passed in from [ember-qunit](https://github.com/emberjs/ember-qunit/blob/0fd92b57dfb8726f4d6ce99fd104cf265e1a200f/addon/src/index.js#L51) and has things like `pauseTest` and other test setup which we would want to preserve.

NullVoxPopuli:
that makes sense!

addon/src/teardown-context.ts
eliasdawson: Good to know. This isn't new though, just moved from `setup-context.ts` and there are several other imports currently in the repo.

Also this project is setting `Ember.testing`, I only see an alternative for checking it.

test-app/tests/unit/dom/fill-in-test.js
```diff
@@ -16,16 +16,6 @@ if (isFirefox || isChrome) {
...
 
-/**
- * Prior to Chrome 129 (canary),
- * Chrome 127.x emits an extra selectionchange event sometimes
- *
- * Delete this once we don't want to test against Chrome 127
- */
-if (isChrome) {
-  clickSteps.push('selectionchange');
-}
```
eliasdawson, line 27: This is unrelated to the context reuse issue but fixes test failures currently on master. Since tests are running on Chrome 130 now, I assume this is safe to remove.

addon/src/teardown-context.ts
```diff
@@ -30,7 +32,14 @@ export default function teardownContext(
     .then(() => {
       _cleanupOnerror(context);
 
-      destroy(context);
```
eliasdawson, line 33: I think the only potential impact of this change would be that if consumers are calling `registerDestructor` on the context for their own cleanup, then that would break and they would need to call it on `context.owner` instead.

addon/src/teardown-context.ts
```diff
@@ -30,7 +32,14 @@ export default function teardownContext(
...
+      _teardownAJAXHooks();
+
+      // SAFETY: this is intimate API *designed* for us to override.
+      (Ember as any).testing = false;
```
NullVoxPopuli, line 38: since this wasn't here before, can we remove it :sweat_smile:

NullVoxPopuli: oh apologies, you said it was from here: https://github.com/emberjs/ember-test-helpers/pull/1511/files#diff-93af636d6cd5ada55ea13d76db06d4e9447f451caae3fd2dc83a7cd4783bd385L223

ok. nevermind.

This needs some thinking unrelated to this PR

eliasdawson:
Any updates/concerns on merging this?

NullVoxPopuli (approved):
Approved, and marked as a breaking change (due to the risk of folks registering destructors on context)

PR #1519, opened 2024-11-22
Cleanup legacy code

chancancode:
Remove the branching needed for less than supported versions (<4) of Ember. Turns out that is everything at this point.

I did not very carefully examine what each branch is doing, but just mechanically evaluated each conditional and deleted the logically dead branch.


addon/src/setup-application-context.ts
```diff
@@ -88,29 +57,19 @@ export function setupRouterSettlednessTracking() {
...
-    router.on('routeDidChange', () => (routerTransitionsPending = false));
-  } else {
-    // SAFETY: similarly, this cast cannot be made safer because on the versions
-    // where we fall into this path, this is *also* not an exported class.
-    const mainRouter = owner.lookup('router:main');
-    assert('router:main is not available', !!mainRouter);
-    router = mainRouter as Router;
-    ROUTER.set(context, router);
-  }
+
```
chancancode, line 60: this..

addon/src/setup-application-context.ts
```diff
@@ -88,29 +57,19 @@ export function setupRouterSettlednessTracking() {
...
+  // SAFETY: unfortunately we cannot `assert` here at present because the
+  // class is not exported, only the type, since it is not designed to be
+  // sub-classed. The most we can do at present is assert that it at least
+  // *exists* and assume that if it does exist, it is bound correctly.
+  const router = owner.lookup('service:router') as RouterService;
+  assert('router service is not set up correctly', !!router);
```
chancancode, line 66: ..and this hunk, should be the only non-trivial modification (even then this is still pretty trivial, just following the lead of the linter and consolidated some variables)

NullVoxPopuli:
any chance you'd be willing to rebase? <3

PR #1521, opened 2024-11-29
Add `force` argument to `triggerEvent`

jrmlt:
I believe `triggerEvent` is too restrictive if it forbids all events on a form control with the `disabled` attribute. While the specification only mentions "click events" and MDN refers to "browsing events," all browsers currently allow mouse tracking events, such as `mouseenter` and `mouseleave`, to trigger on disabled controls. Additionally, there are events unrelated to user interaction, like `transitionend` and `animationend`.  

Beyond known events, there's also the case of `CustomEvent`, which is entirely valid in browsers but is currently blocked by `triggerEvent`.  

Maintaining a comprehensive list of forbidden or allowed events might not be practical or sustainable. However, I suggest providing testers with an option to trigger events on disabled elements if necessary.

This PR allows to bypass the availability check on a `disabled` form control using the `force` argument.


NullVoxPopuli (approved):
lgtm! thank you!!

PR #1528, opened 2025-01-27
Remove `application-template-wrapper` feature

bobisjan:
This is a follow-up to https://github.com/emberjs/ember.js/pull/20683

### DIFF
--- a/addon/src/setup-application-context.ts
+++ b/addon/src/setup-application-context.ts
@@ -4,7 +4,6 @@ import {
   isTestContext,
   type TestContext,
 } from './setup-context.ts';
-import global from './global.ts';
 import hasEmberVersion from './has-ember-version.ts';
 import settled from './settled.ts';
 import getTestMetadata from './test-metadata.ts';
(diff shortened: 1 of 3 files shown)


addon/src/setup-application-context.ts
```diff
@@ -168,13 +167,7 @@ export function visit(
...
     })
     .then(() => {
-      if (global.EmberENV._APPLICATION_TEMPLATE_WRAPPER !== false) {
-        context.element = document.querySelector(
-          '#ember-testing > .ember-view',
-        );
-      } else {
-        context.element = document.querySelector('#ember-testing');
-      }
+      context.element = document.querySelector('#ember-testing');
```
bobisjan, line 170: Should be the `getRootElement()` used instead?

ef4:
We should help land this, it's a good fix. Needs some conflicts to be fixed now.

NullVoxPopuli (approved):
no supported ember version needs this code, so this PR can be a bugfix

Issue #1536, opened 2025-03-22
Triggers deprecations for barrel file use on latest ember-source canary


runspired:
The following code uses the barrel file triggering the deprecation:

https://github.com/emberjs/ember-test-helpers/blob/eaf72946e639cdc007c09aec47f4869d48f2d1f1/addon/src/-internal/-owner-mixin-imports.ts#L61-L78

related issue in ember-qunit https://github.com/emberjs/ember-qunit/issues/1187

PR #1537, opened 2025-03-25
Resolve deprecations from ember barrel file

NullVoxPopuli:


runspired:
resolves https://github.com/emberjs/ember-test-helpers/issues/1536

addon/src/-internal/build-registry.ts
```diff
@@ -102,10 +99,9 @@ export default function buildRegistry(resolver: Resolver) {
   const fallbackRegistry = Application.buildRegistry(namespace);
   // TODO: only do this on Ember < 3.13
```
kategengler, line 100: Is this comment relevant? If so, do we even need to be supporting Ember < 3.13 here?

addon/src/-internal/build-registry.ts
```diff
@@ -2,18 +2,15 @@ import type { Resolver } from '@ember/owner';
 import ApplicationInstance from '@ember/application/instance';
 import Application from '@ember/application';
 import EmberObject from '@ember/object';
-
-import Ember from 'ember';
+import { Registry } from '@ember/-internals/container';
```
kategengler, line 5: I know we specified these with the barrel deprecation RFC, but I wonder if there is anything in ember-source testing to make sure we don't break such import paths?

test-app/tests/helpers/qunit-module-for.js
```diff
@@ -1,17 +1,18 @@
...
 import { module } from 'qunit';
 import QUnitTestAdapter from './qunit-test-adapter';
+import { setAdapter } from 'ember-testing/lib/setup_for_testing';
 
 export default function qunitModuleFor(testModule) {
   module(testModule.name, {
     beforeEach(assert) {
-      if (Ember.testing) {
+      if (isTesting()) {
         throw new Error('should not have Ember.testing === true in beforeEach');
```
kategengler, line 10:
```suggestion
        throw new Error('should not have  isTesting() === true in beforeEach');
```

test-app/tests/helpers/qunit-module-for.js
```diff
@@ -1,17 +1,18 @@
...
       }
-      Ember.Test.adapter = QUnitTestAdapter.create();
+      setAdapter(QUnitTestAdapter.create());
       testModule.setContext(this);
       return testModule.setup(assert).finally(() => {
-        if (!Ember.testing) {
+        if (!isTesting()) {
           throw new Error(
             'should have Ember.testing === true after tests have started'
```
kategengler, line 17:
```suggestion
            'should have isTesting() === true after tests have started'
```

test-app/tests/helpers/qunit-module-for.js
```diff
@@ -20,8 +21,8 @@ export default function qunitModuleFor(testModule) {
     },
     afterEach(assert) {
       return testModule.teardown(assert).finally(() => {
-        Ember.Test.adapter = null;
-        if (Ember.testing) {
+        setAdapter(null);
+        if (isTesting()) {
           throw new Error(
             'should not have Ember.testing === true after tests have finished'
```
kategengler, line 27:
```suggestion
            'should not have isTesting() === true after tests have finished'
```

test-app/tests/unit/settled-test.js
```diff
@@ -69,20 +69,13 @@ module('settled', function (hooks) {
...
 
-    // In Ember < 2.8 `registerWaiter` expected to be bound to
-    // `Ember.Test` 😭
-    //
-    // Once we have dropped support for < 2.8 we should swap this to
-    // use:
-    //
-    // import { registerWaiter } from '@ember/test';
-    Ember.Test.registerWaiter(this._legacyWaiter);
+    registerWaiter(this._legacyWaiter);
```
kategengler, line 72: TIL we support the legacy waiters still

kategengler (changes requested):

addon/src/-internal/build-registry.ts
NullVoxPopuli: they're all private API, so I think it's totally fair for ember to break them if it wishes.

@ember/test-helpers should not be using private APIs haha

(we can explore migrate off of them in a separate PR)

test-app/tests/helpers/qunit-module-for.js
NullVoxPopuli: I'm not entirely sure this file is used.

Will investigate in follow up

test-app/tests/unit/settled-test.js
NullVoxPopuli: yea, and probably will continue to do so until they're deprecated

addon/src/-internal/build-registry.ts
NullVoxPopuli: We don't, but I'd argue cleanup for legacy stuff should be in a different PR (which I can submit next! as there are a number of other cleanup opportunities, I think)

addon/src/-internal/build-registry.ts
kategengler: They may be private but if we don't do that follow up PR all the pain will be ours.

test-app/tests/unit/settled-test.js
kategengler: I think the entire legacy test system is still in there...

kategengler (approved):

addon/src/-internal/build-registry.ts
NullVoxPopuli: aye, I'm doing this now 🎉

PR #1553, opened 2025-10-21
`waitForFocus` DOM Helper

mogstad:
Convinient helper that resolves when a target receives focus. Useful for verifying keyboard navigation handling and default focus.

Uses the pull based waitUntil helper to support the element not being in the DOM when invoking the helper.

Alternatives without this helper is asserting `document.activeElement` is the target. It usually work, but there are cases it focus may happne async. The element isn't in view yet.

### DIFF
--- a/API.md
+++ b/API.md
@@ -28,48 +28,49 @@
     *   [clearRender][24]
 *   [Wait Helpers][25]
     *   [waitFor][26]
-    *   [waitUntil][27]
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (commented):
looks good to me -- want to add docs and all that?

mogstad:
Will do!

PR #1560, opened 2025-12-10
Run settled before destroy()

NullVoxPopuli:
Situation:
- have old code
- do things in constructor instead of deriving data "on page load"
- switch to `@embroider/router`
- tests can sometimes fail now depending on execution order due to registerBundle being async (with waiters! (good)), but not all tests that call the aforementioned constructor are async, and thus wouldn't be doing anything with settled behavior anyway

Solution:
- call await settled before destruction

### DIFF
--- a/addon/src/teardown-context.ts
+++ b/addon/src/teardown-context.ts
@@ -29,6 +29,7 @@ export default function teardownContext(
   { waitForSettled = true }: TeardownContextOptions = {},
 ): Promise<void> {
   return Promise.resolve()
+    .then(settled)
     .then(() => {
       _cleanupOnerror(context);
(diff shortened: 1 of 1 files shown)


addon/src/teardown-context.ts
```diff
@@ -29,6 +29,7 @@ export default function teardownContext(
   { waitForSettled = true }: TeardownContextOptions = {},
 ): Promise<void> {
   return Promise.resolve()
+    .then(settled)
```
kategengler, line 32: when does the argument waitForSettled get passed in as false, and when it does, wouldn't the new await of settled be unexpected?

PR #1564, opened 2026-04-24
pnpm dlx create-release-plan-setup@latest

NullVoxPopuli:
oidc


mansona (approved):
This looks like the standard output of `pnpm dlx create-release-plan-setup@latest` without any changes so it's well tested and expected to work 👍 

we should not merge until OIDC is setup on npm **first** to prevent confusion

PR #1565, opened 2026-04-24
Ember 7 compat

NullVoxPopuli:
Get CI working with ember 7

diif looks large, because this repo has multiple lockfiles

summary of changes
- dep updates
- ts matrix change - this doesn't mean we don't support earlier TSes, but it is too cumbersome to test against earlier TSes


mansona:
> ts matrix change - this doesn't mean we don't support earlier TSes, but it is too cumbersome to test against earlier TSes
for safety we should probably make this a breaking no?

NullVoxPopuli:
>  for safety we should probably make this a breaking no?

I don't think so -- nothing about the addon actually changed.

NullVoxPopuli:
we also don't do breaking changes for TS very frequently due to the rolling policy

kategengler (approved):

PR #1571, opened 2026-06-12
No extend

NullVoxPopuli:


kategengler (approved):
I think this conversion is fine and should be safe. 

I'm more concerned about the changes in Ember that made this necessary -- will they affect other consumers?

NullVoxPopuli:
> will they affect other consumers?

from disco in disco-rd -- highly likely not

- test-helpers still allows for resolver overriding correctly
- any property that isn't the resolver should be safe still
- as long as people are using class or half-class syntax with their application, they can still override the `Resolver` property

PR #1573, opened 2026-07-15
Add renderComponent support for emberjs/emberjs#21460

BobrImperator:
Related https://github.com/emberjs/ember.js/pull/21460, https://github.com/mainmatter/ember.js/compare/rfc-1169-route-manager...mainmatter:ember.js:test-new-outlet

**This change is backwards compatible**

Provides a rendering support using a new private `renderRootComponent` method which exists on `ApplicationInstance`.
- `-outlet` / `OutletView` is being removed in the `rfc-1169` implementation
- bumps `ember-source` to `6.8` which has the `@ember/-internals/component` import.

### DIFF
--- a/addon/src/setup-rendering-context.ts
+++ b/addon/src/setup-rendering-context.ts
@@ -146,19 +146,22 @@ function renderViaRenderComponent(
 
   const ownerToRenderFrom = options?.owner || owner;
 
+  // wrapping in `run` enables `setupOnerror` hook
   if (
     ownerToRenderFrom === owner &&
     typeof (owner as any).renderRootComponent === 'function'
   ) {
-    (owner as any).renderRootComponent(component);
+    run(() => (owner as any).renderRootComponent(component));
   } else {
     // @TODO: Evaluate if `renderRootComponent` should allow an alternative owner
     // as per `RenderOptions.owner` comment
-    renderComponent(component, {
-      into: getRootElement() as Element,
-      owner: ownerToRenderFrom,
-      appendIntoTarget: true,
-    });
+    run(() =>
+      renderComponent(component, {
+        into: getRootElement() as Element,
+        owner: ownerToRenderFrom,
+        appendIntoTarget: true,
+      }),
+    );
   }
 }
 
--- a/addon/src/setup-rendering-context.ts
+++ b/addon/src/setup-rendering-context.ts
@@ -19,6 +19,9 @@ import isComponent from './-internal/is-component.ts';
 import { precompileTemplate } from '@ember/template-compilation';
 import { setComponentTemplate } from '@glimmer/manager';
 import { setComponentManager, capabilities } from '@ember/component';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { renderComponent } from '@ember/-internals/glimmer';
 
 const OUTLET_TEMPLATE = precompileTemplate(`{{outlet}}`, { strictMode: false });
 const EMPTY_TEMPLATE = precompileTemplate(``, { strictMode: false });
@@ -50,6 +53,10 @@ export function isRenderingTestContext(
 
 function supportsRenderRootComponent(owner: Owner): boolean {
   return typeof (owner as any).renderRootComponent === 'function';
+}
+
+function supportsRenderComponent(): boolean {
+  return typeof renderComponent === 'function';
 }
 
 /**
@@ -128,6 +135,7 @@ function renderViaRenderComponent(
   owner: Owner,
   context: object,
   templateFactoryOrComponent: object,
+  options?: RenderOptions,
 ): void {
   let component;
   if (isComponent(templateFactoryOrComponent)) {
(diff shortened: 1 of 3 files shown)


addon/src/setup-rendering-context.ts
```diff
@@ -85,6 +91,114 @@ export interface RenderOptions {
...
+function renderViaRenderComponent(
+  owner: Owner,
+  context: object,
+  templateFactoryOrComponent: object,
+): void {
+  let component: object;
+  if (isComponent(templateFactoryOrComponent)) {
+    component = templateFactoryOrComponent;
+  }
+  component = contextComponentFor(templateFactoryOrComponent, context);
```
mansona: shouldn't this be in an else or something? 🤔

BobrImperator: Yes, I had removed a little too much when refactoring and forgot to rebuild :see_no_evil:

addon/src/setup-rendering-context.ts
```diff
@@ -17,6 +17,14 @@ import isComponent from './-internal/is-component.ts';
...
 // @ts-ignore
 import { precompileTemplate } from '@ember/template-compilation';
+import {
+  setComponentManager,
+  setComponentTemplate,
+  capabilities,
+} from '@ember/component';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { renderComponent } from '@ember/-internals/glimmer';
```
NullVoxPopuli: why not import from `@ember/renderer`?

addon/src/setup-rendering-context.ts
```diff
@@ -13,10 +13,13 @@ import getTestMetadata from './test-metadata.ts';
 import { runHooks } from './helper-hooks.ts';
 import isComponent from './-internal/is-component.ts';
 
-// the built in types do not provide types for @ember/template-compilation
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
```
BobrImperator, line 18: That comment no longer held since 6.8 dev-dependency bump

addon/src/setup-rendering-context.ts
```diff
@@ -46,6 +49,14 @@ export function isRenderingTestContext(
   return isTestContext(context) && hasCalledSetupRenderingContext in context;
 }
 
+function supportsRenderRootComponent(owner: Owner): boolean {
+  return typeof (owner as any).renderRootComponent === 'function';
+}
+
+function supportsRenderComponent(): boolean {
+  return typeof renderComponent === 'function';
```
NullVoxPopuli: if we imported it, we support it -- an ember-source-gated macroCondition may be better here

addon/src/setup-rendering-context.ts
```diff
@@ -85,6 +96,132 @@ export interface RenderOptions {
...
+  Render `component` into the testing root element using the `renderComponent`
+*/
+function renderViaRenderComponent(
+  owner: Owner,
+  context: object,
+  templateFactoryOrComponent: object,
+  options?: RenderOptions,
+): void {
+  let component;
+  if (isComponent(templateFactoryOrComponent)) {
```
NullVoxPopuli, line 135: wouldn't we always have a component? no "template factory"?

BobrImperator: Users still can provide an hbs template which we then turn to a component via `contextComponentFor` so we need to keep this code path I believe

addon/src/-internal/render-component.ts
```diff
@@ -0,0 +1,16 @@
...
+  macroCondition,
+  importSync,
+  appEmberSatisfies,
+} from '@embroider/macros';
+
+type RendererModule = typeof import('@ember/renderer');
+
+let renderComponent: RendererModule['renderComponent'] | undefined;
+
+if (macroCondition(appEmberSatisfies('>=6.8.0-alpha.1'))) {
```
NullVoxPopuli, line 11: looks good!

NullVoxPopuli:
had to tickle ci a bit

BobrImperator:
@NullVoxPopuli give it a poke please

It's fixed and confirmed by https://github.com/mainmatter/ember-test-helpers/pull/1

addon/src/setup-rendering-context.ts
```diff
@@ -85,6 +92,131 @@ export interface RenderOptions {
...
+
+  // wrapping in `run` enables `setupOnerror` hook
+  if (
+    ownerToRenderFrom === owner &&
+    typeof (owner as any).renderRootComponent === 'function'
+  ) {
+    run(() => (owner as any).renderRootComponent(component));
+  } else {
+    run(() =>
+      renderComponent!(component, {
```
NullVoxPopuli, line 151:
```suggestion
      renderComponent(component, {
```

if we don't have `renderComponent` here, we should throw an error. I trust TS here more than I do "assumptions" <3

addon/src/setup-rendering-context.ts
```diff
@@ -85,6 +92,131 @@ export interface RenderOptions {
...
+  } else {
+    component = contextComponentFor(templateFactoryOrComponent, context);
+  }
+
+  const ownerToRenderFrom = options?.owner || owner;
+
+  // wrapping in `run` enables `setupOnerror` hook
+  if (
+    ownerToRenderFrom === owner &&
+    typeof (owner as any).renderRootComponent === 'function'
```
NullVoxPopuli, line 146: can we extract this to a `hasRouteManager` function to expose intent / feature-detection a bit more precisely?

renderRootComponent and renderComponent are conflicting in my gut, which I don't like <3

NullVoxPopuli: oh, we already have a function here: https://github.com/emberjs/ember-test-helpers/pull/1573/changes#diff-a7db04a222f038670f0d679608a71e396542f41c744f7449bd8258bfa0accbefR52


can we just use that and add a clarifying comment on top about why it exists and what it covers?

ef4 (approved):

PR #1574, opened 2026-08-09
[SPIKE] settled(): await settledness instead of polling for it

NullVoxPopuli-ai-agent:
Exploration for the RFC 957 test story (context: emberjs/ember.js#21520, the runloop-less scheduler spike). Draft — for signal and discussion, not to merge as-is. **Depends on emberjs/ember-test-waiters#525** for `waitersSettled` (CI here will not typecheck until that lands and releases).

## What changes

`settled()` stops polling `isSettled` on a timer cadence and awaits the settledness promises instead:

```js
for (;;) {
  await Promise.race([
    Promise.all([renderSettled(), waitersSettled()]),
    new Promise((resolve) => setTimeout(resolve, 50)),
  ]);
  await new Promise((resolve) => setTimeout(resolve, 0));
  if (isSettled()) return;
}
```

Rendering is bridged into a test waiter, so on builds that report render edges (`_onRenderSettledChange`) a pending render is counted in `hasPendingWaiters` and is named in waiter debug output instead of being an anonymous `isRenderPending` boolean.

## Why each piece survived review

Three things look deletable and aren't; each was measured after trying the deletion:

- **the 50ms race** — with `settled()` reduced to a bare `await Promise.all([renderSettled(), waitersSettled()])`, this repo's suite goes 553/553 → **5 failures**. Run loop timers, legacy `Ember.Test.registerWaiter` callbacks, the request counter, and `Waiter`s without `settled()` cannot announce completion. 50ms rather than 10 so it loses the race to a frame-paced render tick (at 10ms it decided 30 of 117 iterations; at 50ms, 1 of 92 — the promises decide in practice).
- **the 0ms yield** — quiet must be observed from a macrotask; already-queued task sources (worker messages, zero-delay timers) register waiters right after a microtask-context check says settled. ~1-in-4 flake without it on a worker-based compile suite. `waitUntil` imposed this implicitly.
- **the render bridge** — deleting it takes limber's tutorial suite from **71/71 to 10+ 20s timeouts**. `renderSettled()` is one-shot ("is the work I asked about done?"); `isSettled()` is checked afterwards and needs a *current* render signal, which classic builds get from `hasRunLoop` and runloop-less builds otherwise have nothing for.

The helper functions those pieces lived in are gone — the logic is inline with its reasoning.

## Validation

- This repo's test-app: **553/553** (against a linked build of the companion PR, and against published test-waiters where `waitersSettled` is absent and the fallback drives).
- Runloop-less ember spike, via the equivalent dist patches in NullVoxPopuli/limber#2214: full chrome matrix green (repl + tutorial + ember-repl), 7/7 repeat runs on the suite that exposed the macrotask flake.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/addon/src/settled.ts
+++ b/addon/src/settled.ts
@@ -1,14 +1,9 @@
-import {
-  macroCondition,
-  dependencySatisfies,
-  importSync,
-} from '@embroider/macros';
+import { importSync } from '@embroider/macros';
 import { Test } from 'ember-testing';
 
 import { nextTick } from './-utils.ts';
 import { hasPendingTransitions } from './setup-application-context.ts';
-import { buildWaiter, hasPendingWaiters } from '@ember/test-waiters';
-import * as testWaiters from '@ember/test-waiters';
+import { hasPendingWaiters, waitersSettled } from '@ember/test-waiters';
 import type DebugInfo from './-internal/debug-info.ts';
 import { TestDebugInfo } from './-internal/debug-info.ts';
 import renderSettled from './-internal/render-settled.ts';
@@ -18,38 +13,6 @@ import renderSettled from './-internal/render-settled.ts';
 // namespace -- a missing export degrades to `undefined` here instead of a
 // build-time missing-export error in consuming apps.
 const _backburner: any = (importSync('@ember/runloop') as any)._backburner;
-
-// Ember builds that schedule rendering without the runloop report the
-// EDGES of rendering work (pending / complete) rather than exposing a
-// pollable flag. Bridging those edges into a test waiter folds rendering
-// into the same settledness protocol as every other async source: it
-// needs no clause of its own in `isSettled` below, and a render that
-// never completes is reported by name in test-waiter debug output.
-const renderWaiter = buildWaiter('@ember/test-helpers:render');
-let renderWaiterToken: unknown = null;
-
-const usesRenderWaiter = (() => {
-  if (macroCondition(dependencySatisfies('ember-source', '>=4.5.0-beta.1'))) {
-    const renderer = importSync('@ember/renderer') as any;
-
-    if (typeof renderer._onRenderSettledChange === 'function') {
-      renderer._onRenderSettledChange((pending: boolean) => {
-        if (pending) {
-          renderWaiterToken ??= renderWaiter.beginAsync();
-        } else if (renderWaiterToken !== null) {
-          const token = renderWaiterToken;
-
-          renderWaiterToken = null;
-          renderWaiter.endAsync(token);
-        }
-      });
-
-      return true;
-    }
-  }
-
-  return false;
-})();
 
 let requests: XMLHttpRequest[];
 const checkWaiters = Test.checkWaiters;
@@ -185,11 +148,10 @@ export function getSettledState(): SettledState {
   const hasPendingTestWaiters = hasPendingWaiters();
   const pendingRequestCount = pendingRequests();
   const hasPendingRequests = pendingRequestCount > 0;
-  // On runloop-driven builds, a pending render is observable as backburner's
-  // autorun instance. Scheduler-driven builds report render edges into the
-  // render waiter above, so a pending render is already counted in
-  // `hasPendingTestWaiters` -- reporting it here too would double-count it.
-  const isRenderPending = usesRenderWaiter ? false : !!hasRunLoop;
+  // On runloop-driven builds a pending render is observable as backburner's
+  // autorun instance. Builds that schedule without the runloop have nothing
+  // to observe here -- `settled()` awaits `renderSettled()` directly.
+  const isRenderPending = !!hasRunLoop;
 
   return {
     hasPendingTimers,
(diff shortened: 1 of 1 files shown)


NullVoxPopuli-ai-agent:
Pushed a second iteration that goes further, and I think lands somewhere better: **rendering becomes a test waiter.**

Rather than the framework exposing a pollable `isRenderPending()` that `isSettled()` has to read as its own clause, the renderer reports the *edges* of its work (`_onRenderSettledChange`: `true` when rendering or destruction becomes outstanding, `false` when complete), and this addon bridges those edges into `buildWaiter('@ember/test-helpers:render')`. Consequences:

- `isSettled()` has no render-specific clause at all on such builds — a pending render is already counted in `hasPendingWaiters` (the old clause is hardwired `false` to avoid double-counting).
- A render that never completes is reported **by name** in test-waiter debug output, instead of as an anonymous `isRenderPending: true` boolean.
- Settledness becomes one push-based protocol rather than 'waiters, plus a separately-polled render flag'.

Runloop-driven builds are untouched — they still infer pending renders from backburner's autorun instance.

Validation: this repo's test-app **553/553** on stock ember; against the runloop-less spike build, limber's full app+addon chrome matrix green and 7/7 repeat runs on the suite that originally exposed the macrotask-confirm flake.

addon/src/settled.ts
NullVoxPopuli, line 1: bad bot. this is not what I meant. 

I meant in the test-waiters package, when we go beginAsync, we can have that be internally new Promise, and then endAsync resolves the promise.

test-waiters can have a waitersSettled() function which sees if all known promises are resolved (tho, if they resolve, I would also assume they remove themselves from the structure that waitersSettled() would look at)

NullVoxPopuli-ai-agent: Understood, and you're right — I built the aggregate version (one module-level deferred that resolved when `hasPendingWaiters()` went false), not per-operation promises. Redone as you described, in emberjs/ember-test-waiters#525:

- `beginAsync` creates a promise for that operation
- `endAsync` resolves it **and deletes it** from the pending map, so the map only ever holds work in flight
- `waitersSettled()` composes the promises of what's currently tracked, via a new optional `settled()` on each waiter

Your design is also strictly better than mine for a reason I hadn't considered: my version called `hasPendingWaiters()` on every `endAsync`, which walks every registered waiter and builds `debugInfo` arrays for the pending ones. Per-operation promises do no scanning at all — `endAsync` resolves one deferred.

One case the interface forced a decision on: a `Waiter` implemented directly against the interface has no `settled()` to compose. Rather than reporting a quiet it can't verify, `waitersSettled()` returns a never-resolving promise while such a waiter is pending, so the caller's fallback tick drives and pull-only waiters keep working.

Re-validated: test-waiters `base-tests` 56/56; this repo's test-app 553/553 both against a linked build of the new API and against published test-waiters (degraded to the fallback); limber's full chrome matrix green with 7/7 repeat runs against the runloop-less spike.

addon/src/settled.ts
```diff
@@ -209,6 +251,91 @@ export function isSettled(): boolean {
...
+function waitersQuiet(): Promise<unknown> {
+  return waitersSettled === null ? new Promise(() => {}) : waitersSettled();
+}
+
+/**
+ * Yields to the task queue, so quiet is observed from a macrotask.
+ *
+ * @private
+ */
+function macrotask(): Promise<void> {
```
NullVoxPopuli: delete this function

addon/src/settled.ts
```diff
@@ -209,6 +251,91 @@ export function isSettled(): boolean {
...
+ * of 117), costing an extra pass each time; at 50ms it decided 1 of 92.
+ *
+ * Without waiter notification it is the loop's only clock, so it stays
+ * at the cadence the previous `waitUntil`-based implementation used.
+ *
+ * @private
+ */
+const FALLBACK_MS = waitersSettled === null ? 10 : 50;
+
+function fallbackTick(): Promise<void> {
```
NullVoxPopuli: delete this function

addon/src/settled.ts
```diff
@@ -1,15 +1,55 @@
...
+// Ember builds that schedule rendering without the runloop report the
+// EDGES of rendering work (pending / complete) rather than exposing a
+// pollable flag. Bridging those edges into a test waiter folds rendering
+// into the same settledness protocol as every other async source: it
+// needs no clause of its own in `isSettled` below, and a render that
+// never completes is reported by name in test-waiter debug output.
+const renderWaiter = buildWaiter('@ember/test-helpers:render');
+let renderWaiterToken: unknown = null;
+
+const usesRenderWaiter = (() => {
```
NullVoxPopuli: why all of this condition? we don't care, right? because we can just await renderSettled()

addon/src/settled.ts
NullVoxPopuli-ai-agent: Function deleted and inlined — but the 50ms timer itself has to stay, and I have data rather than an opinion on that now.

I first built exactly the minimal version (`settled()` = `await Promise.all([renderSettled(), waitersSettled()])`, no race, no loop, no yield). This repo's own suite goes from 553/553 to **5 failures**:

```
not ok 9  - rerender real-world: using rerender to test loading states works
not ok 11 - Global error: calling set on destroyed object … internalValue = async value
not ok 12 - settled real-world scenarios: it works when async exists in `init`
not ok 13 - Global error: calling set on destroyed object … internalValue = async value
not ok 14 - settled real-world scenarios: rerender - it basically works
```

The reason is that not everything announces completion: run loop timers (`hasPendingTimers`), legacy `Ember.Test.registerWaiter` callbacks, the AJAX request counter, and any `Waiter` implemented against the interface without `settled()`. `settled()` returns while those are pending, and the destroyed-object assertions are the fallout.

With the timer inlined and the loop kept: 553/553.

addon/src/settled.ts
NullVoxPopuli-ai-agent: Function deleted and inlined too. The 0ms yield it did stays, for the same measured reason as before: quiet observed in microtask context beats already-queued task sources (worker messages, zero-delay timers) that are about to register waiters or dirty state, and `settled()` returns early. That flaked ~1-in-4 runs on limber's worker-based compile suite until the yield was there; `waitUntil` had been imposing the same boundary implicitly by scheduling every check through `setTimeout`.

addon/src/settled.ts
NullVoxPopuli-ai-agent: `settled()` **does** just await `renderSettled()` — that part is unconditional now. But I tried deleting the bridge entirely and it doesn't hold up, so I'd rather show you the A/B than quietly keep it.

Same limber tutorial suite, runloop-less ember build, only difference is this block:

| | result |
| --- | --- |
| bridge present | **71/71 pass** |
| bridge deleted | **10+ chapters fail**, each at the 20s test timeout |

Reproducible, both directions, re-run to confirm.

The reason the awaited promise isn't sufficient: `renderSettled()` answers *"has the work that was outstanding when I asked finished?"* — one-shot. `isSettled()` answers *"is anything outstanding right now?"*, and it is checked **after** the await, past a macrotask yield in which new dirt can land. Without the bridge, `isSettled()` has no render signal at all on runloop-less builds (`hasRunLoop` is false — there's no backburner), so a render that starts during that window is invisible and the loop exits with rendering pending. On classic builds `hasRunLoop` covers exactly this, which is why the gap only shows up on the runloop-less side.

So the bridge isn't a second settledness mechanism — it's how `isSettled()` keeps the render signal it already has on classic builds, expressed through the waiter protocol instead of a new `isRenderPending()` API. That framing is also what lets it be *deleted* the day rendering-pending is folded into waiters upstream in ember itself.

(I haven't traced the tutorial hang to the exact line inside its compile pipeline — what I can say confidently is the A/B above and that the missing sync signal is the only behavioural difference.)

